PR #1, opened 2017-05-19
Scale down feature

Henrod:
### How it's done
- Get from config/local.yaml how many room to remove per scale down
- Using this amount N, it access redis and get N random rooms that have status ready
- Remove their pods and services
- Remove their keys from Redis
- In GetSchedulerScalingInfo function, when a scheduler doesn't exist on DB it doesn't return error. Instead it just don't set any value on scheduler variable. So I am checking if YAML property is empty.

### DIFF
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -274,17 +274,11 @@ func ScaleDown(logger logrus.FieldLogger, mr *models.MixedMetricsReporter, db pg
 	}()
 
 	l.Debug("accessing redis")
-	pipe := redisClient.TxPipeline()
-	sReady := pipe.SPopN(models.GetRoomStatusSetRedisKey(scheduler.Name, models.StatusReady), int64(amount))
-	_, err := pipe.Exec()
-	if err != nil {
-		l.WithError(err).Error("scale down error")
-		return err
-	}
+	sReady := redisClient.SPopN(models.GetRoomStatusSetRedisKey(scheduler.Name, models.StatusReady), int64(amount))
 
 	idleRooms, err := sReady.Result()
 	if err != nil {
-		l.WithError(err).Error("scale down error")
+		l.WithError(err).Error("scale down error, failed to retrieve ready rooms from redis")
 		return err
 	}
 
@@ -349,14 +343,6 @@ func deleteSchedulerHelper(logger logrus.FieldLogger, mr *models.MixedMetricsRep
 		}
 	}
 
-	err = mr.WithSegment(models.SegmentDelete, func() error {
-		return scheduler.Delete(db)
-	})
-	if err != nil {
-		logger.WithError(err).Error("failed to delete scheduler from database while deleting scheduler")
-		return err
-	}
-
 	configYAML, _ := models.NewConfigYAML(scheduler.YAML)
 	// Delete pods and wait for graceful termination before deleting the namespace
 	err = mr.WithSegment(models.SegmentNamespace, func() error {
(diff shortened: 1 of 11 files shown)


controller/controller.go
```diff
@@ -258,6 +261,80 @@ func ScaleUp(logger logrus.FieldLogger, mr *models.MixedMetricsReporter, db pgin
...
+	})
+	timeout := make(chan bool, 1)
+	go func() {
+		time.Sleep(time.Duration(timeoutSec) * time.Second)
+		timeout <- true
+	}()
+
+	l.Debug("accessing redis")
+	pipe := redisClient.TxPipeline()
+	sReady := pipe.SPopN(models.GetRoomStatusSetRedisKey(scheduler.Name, models.StatusReady), int64(amount))
```
cscatolini: I never knew redis had a SPopN method. Cool 😎

cscatolini: If your are calling just one method in redis there's no need for a txpipeline.

controller/controller.go
```diff
@@ -258,6 +261,80 @@ func ScaleUp(logger logrus.FieldLogger, mr *models.MixedMetricsReporter, db pgin
...
+	_, err := pipe.Exec()
+	if err != nil {
+		l.WithError(err).Error("scale down error")
```
cscatolini: use a better message here so we can distinguish this error from the one below: "failed to retrieve ready rooms from redis" maybe?

controller/controller.go
```diff
@@ -272,6 +349,14 @@ func deleteSchedulerHelper(logger logrus.FieldLogger, mr *models.MixedMetricsRep
 		}
 	}
 
+	err = mr.WithSegment(models.SegmentDelete, func() error {
```
cscatolini: Why did you move the scheduler removal from db to the first operation? It was the last beacause if something failed, the watcher would recreate the other parts (namespace, redis info, etc) and if the user still wanted to delete the scheduler he could call the endpoint again.

If you first delete it from the db and then attempt to do the remaining ops when there's a failure we'll have junk info (maybe a namespace, junk info in redis, etc) that we won't be able to detect easily.

Please move it back to the last op and add a comment just above it explaining why this is needed.

watcher/watcher.go
```diff
@@ -196,10 +186,24 @@ func (w *Watcher) AutoScale() {
...
-		"roomsByStatus": roomCountByStatus,
-		"state":         scheduler.State,
+		"ready":       roomCountByStatus.Ready,
+		"creating":    roomCountByStatus.Creating,
+		"occupied":    roomCountByStatus.Occupied,
+		"terminating": roomCountByStatus.Terminating,
+		"state":       scheduler.State,
 	})
 
+	err = controller.CreateNamespaceIfNecessary(
```
cscatolini: Ok this makes sense. We'll just create the namespace if the scheduler still exists in the db.

controller/controller.go
Henrod: That makes sense.
I changed it because when I deleted a scheduler, the namespace was recreated without services and pods. I thought it was because the line on DB was not deleted, but actually just moving CreateNamespaceIfNecessary solved the problem. 
I will move it back.

controller/controller.go
Henrod: Yeah, I need access to github.com/topfreegames/extensions to add this new method to RedisClient interface

cscatolini: Done! You should be able to commit and push to github.com/topfreegames/extensions

controller/controller.go
cscatolini: Well, the watcher should recreate it with services and pods. This might be a bug so please take some time to make sure it works as expected.

Henrod: It does, the problem was calling CreateNamespaceIfNecessary before GetSchedulerScalingInfo

controller/controller.go
```diff
@@ -332,13 +403,17 @@ func deleteSchedulerHelper(logger logrus.FieldLogger, mr *models.MixedMetricsRep
 		}
 	}
 
+	// Delete from DB must be the last operation because
```
cscatolini: great, thanks

PR #15, opened 2017-08-01
Threshold

Henrod:
# Threshold

A new way to scale GRUs in Maestro

Watcher gets a window of time that is the scheduler's _autoscaling.up.trigger.time_. At each cycle (every _watcher.autoScalingPeriod_ seconds) if the usage is above _autoscaling.up.trigger.usage_, a counter is incremented. 

At the end of the window, if the counter over the number of points in this window is above _autoscaling.up.trigger.percentage_, the GRUs are scaled up. 

The same works for scale down.

### DIFF
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -308,7 +308,6 @@ func ScaleUp(logger logrus.FieldLogger, mr *models.MixedMetricsReporter, db pgin
 	err = waitForPods(
 		willTimeoutAt.Sub(time.Now()),
 		clientset,
-		amount,
 		configYAML.Name,
 		pods,
 		l,
@@ -421,7 +420,7 @@ func UpdateSchedulerConfig(
 	redisClient *redis.Client,
 	clientset kubernetes.Interface,
 	configYAML *models.ConfigYAML,
-	timeoutSec, lockTimeoutMS int,
+	timeoutSec, lockTimeoutMS, maxSurge int,
 	lockKey string,
 	clock clockinterfaces.Clock,
 	schedulerOrNil *models.Scheduler,
@@ -431,6 +430,10 @@ func UpdateSchedulerConfig(
 		"source":    "updateSchedulerConfig",
 		"scheduler": schedulerName,
 	})
+
+	if maxSurge <= 0 {
+		return errors.New("invalid parameter: maxsurge must be greater than 0")
+	}
 
 	var scheduler *models.Scheduler
 	var oldConfig models.ConfigYAML
@@ -448,16 +451,34 @@ func UpdateSchedulerConfig(
 		}
 	}
 
+	denominator := 100 / maxSurge
+	sum := func(start, total, denominator int) int {
+		delta := total / denominator
+		if delta == 0 {
+			delta = 1
+		}
+		start = start + delta
+		if start > total {
+			start = total
+		}
+		return start
+	}
+
 	var lock *redisLock.Lock
 	if MustUpdatePods(&oldConfig, configYAML) {
 		// Lock watchers so they don't scale up and down
 		timeoutDur := time.Duration(timeoutSec) * time.Second
 		willTimeoutAt := clock.Now().Add(timeoutDur)
-		lock, err = redisClient.EnterCriticalSection(redisClient.Client, lockKey, time.Duration(lockTimeoutMS)*time.Millisecond, 0, 0)
-		ticker := time.NewTicker(500 * time.Millisecond)
+		ticker := time.NewTicker(2 * time.Second)
 		timeout := time.NewTimer(timeoutDur)
 	waitForLock:
 		for {
+			lock, err = redisClient.EnterCriticalSection(
+				redisClient.Client,
+				lockKey,
+				time.Duration(lockTimeoutMS)*time.Millisecond,
+				0, 0,
+			)
 			select {
 			case <-timeout.C:
 				return errors.New("timeout while wating for redis lock")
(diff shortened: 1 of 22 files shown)


manifests/scheduler-config-1.yaml
```diff
@@ -16,9 +16,11 @@ autoscaling:
       usage: 70
       time: 10
     cooldown: 30
+      percentage: 80
```
felipejfc: indentation seems wrong

manifests/scheduler-config-1.yaml
```diff
@@ -16,9 +16,11 @@ autoscaling:
...
   down:
     delta: 1
     trigger:
       usage: 50
       time: 10
     cooldown: 30
+      percentage: 80
```
felipejfc: here too

docs/api.md
```diff
@@ -473,15 +475,17 @@ All API responses include a `X-Maestro-Version` header with the current Maestro
           "delta": 10,
           "trigger": {
             "usage": 70,
-            "time": 600
+            "time": 600,
```
felipejfc: talvez seja melhor chamar de window agora

models/scheduler.go
```diff
@@ -49,9 +49,13 @@ type EnvVar struct {
 }
 
 // ScalingPolicyTrigger has the configuration for a scaling policy trigger
+// During 'Time' seconds with n measures, 'Usage'% of the machines needs
+// to be occupied on 'Percentage'% of these n points.
+// This will trigger a scale up or scale down.
 type ScalingPolicyTrigger struct {
-	Time  int `yaml:"time" json:"time" valid:"int64"`
-	Usage int `yaml:"usage" json:"usage" valid:"int64"`
+	Time       int `yaml:"time" json:"time" valid:"int64"`
```
felipejfc: Window

watcher/watcher.go
```diff
@@ -50,6 +50,22 @@ type Watcher struct {
...
 	EventForwarders   []eventforwarder.EventForwarder
+	ScaleUpInfo       *ScaleInfo
+	ScaleDownInfo     *ScaleInfo
+}
+
+//ScaleInfo holds information about last time scheduler was verified if it needed to be scaled
+// and how many time it was above or below threshold
+type ScaleInfo struct {
+	LastTimeCheckedScale time.Time
+	PointsAboveThreshold int
```
felipejfc: PointsAboveOrBelowThreshold?

watcher/watcher.go
```diff
@@ -375,54 +393,81 @@ func (w *Watcher) AutoScale() {
...
+		return false, false, changedState
 	}
 
 	if roomCount.Total() < autoScalingInfo.Min {
-		return true, false, false
+		return true, false, changedState
 	}
 
 	if 100*roomCount.Ready < (100-autoScalingInfo.Up.Trigger.Usage)*roomCount.Total() {
+		inSync = false
```
felipejfc: is this right? shouldn't we set inSync to false only when w.passedThreshold("up", autoScalingInfo)==true ?

watcher/watcher.go
```diff
@@ -375,54 +393,81 @@ func (w *Watcher) AutoScale() {
...
 		}
-		inCooldownPeriod = true
+		w.ScaleUpInfo.LastTimeCheckedScale = now
+		w.ScaleUpInfo.PointsAboveThreshold = 0
 	}
 
+	downTriggerDuration := time.Duration(autoScalingInfo.Down.Trigger.Time) * time.Second
 	if 100*roomCount.Ready > (100-autoScalingInfo.Down.Trigger.Usage)*roomCount.Total() &&
 		roomCount.Total()-autoScalingInfo.Down.Delta >= autoScalingInfo.Min {
+		inSync = false
```
felipejfc: should we set inSync to false here?

watcher/watcher.go
```diff
@@ -375,54 +393,81 @@ func (w *Watcher) AutoScale() {
...
-			return false, false, true
+			changedState = true
 		}
+	}
+	if w.ScaleDownInfo.LastTimeCheckedScale.Add(downTriggerDuration).Before(now) {
 		if nowTimestamp-scheduler.LastScaleOpAt > int64(autoScalingInfo.Down.Cooldown) &&
-			nowTimestamp-scheduler.StateLastChangedAt > int64(autoScalingInfo.Down.Trigger.Time) {
-			return false, true, false
+			w.passedThreshold("down", autoScalingInfo) {
+			shouldScaleDown = true
```
felipejfc: I guess that inSync should be set to false here

manifests/scheduler-config-1.yaml
Henrod: yes, ok

watcher/watcher.go
Henrod: It's right. I want to indicate that the scheduler is over dimensioned before it scales down, for example. Setting here shows this log and when a cycle completes, it scales down, getting back to insync state.

watcher/watcher.go
Henrod: ok

models/scheduler.go
Henrod: good, i will change that

controller/controller.go
```diff
@@ -481,122 +502,147 @@ func UpdateSchedulerConfig(
...
+			if err != nil {
+				logger.WithError(err).Error("error when updating scheduler and waiting for new pods to run")
+				for _, podToDelete := range newPods {
+					deletePod(l, mr, clientset, redisClient.Client, configYAML.Name, podToDelete.GetName())
+					room := models.NewRoom(podToDelete.GetName(), schedulerName)
+					err := room.ClearAll(redisClient.Client)
 					if err != nil {
-						logger.WithError(err).Error("error when creating pod")
-						continue
+						//TODO: try again so Redis doesn't hold trash data
```
felipejfc: This TODO seems important

PR #17, opened 2017-09-15
feature/reporters

ghostec:
created a Reporter interface with a DogStatsD implementation.

There's a Reporters singleton instance:
- reporters.GetInstance()
- reporters.MakeReporters() - initialization from config/{}.yaml

The following scenarios are covered by reporters.GetInstance().Report(...):

#### gru.delete - pod.Delete(...)
ping_timeout
occupied_timeout
scale_down
#### gru.new - NewPod(...)


controller/controller.go
```diff
@@ -378,7 +379,7 @@ func ScaleDown(logger logrus.FieldLogger, mr *models.MixedMetricsReporter, db pg
 	var deletionErr error
 
 	for _, roomName := range idleRooms {
-		err := deletePod(logger, mr, clientset, redisClient, scheduler.Name, roomName)
+		err := deletePod(logger, mr, clientset, redisClient, scheduler.Name, scheduler.Game, roomName, "scale_down")
```
Henrod: It is better if the reasons are at models/contants.go and you call the variables

controller/controller.go
```diff
@@ -555,7 +557,7 @@ waitForLock:
 					msg := "error when deleting old rooms. Maestro will scale up, if necessary, with previous room configuration"
 					return maestroErrors.NewKubernetesError(msg, err)
 				default:
-					err := deletePod(logger, mr, clientset, redisClient.Client, schedulerName, pod.GetName())
+					err := deletePod(logger, mr, clientset, redisClient.Client, schedulerName, gameName, pod.GetName(), "update")
```
Henrod: It is better if the reasons are at models/contants.go and you call the variables

controller/controller.go
```diff
@@ -580,7 +582,7 @@ waitForLock:
 				select {
 				case <-timeout.C:
 					for _, podToDelete := range newPods {
-						err := deletePod(logger, mr, clientset, redisClient.Client, schedulerName, podToDelete.GetName())
+						err := deletePod(logger, mr, clientset, redisClient.Client, schedulerName, gameName, podToDelete.GetName(), "update")
```
Henrod: It is better if the reasons are at models/contants.go and you call the variables

Henrod: This reason should not be update. At this point, update timed out and is deleting the newly created pods. Probably an error of ErrImagePull occurred or the pods are pending. A better reason would be "update error"

controller/controller.go
```diff
@@ -640,7 +642,7 @@ waitForLock:
 			if err != nil {
 				logger.WithError(err).Error("error when updating scheduler and waiting for new pods to run")
 				for _, podToDelete := range newPods {
-					deletePod(l, mr, clientset, redisClient.Client, configYAML.Name, podToDelete.GetName())
+					deletePod(l, mr, clientset, redisClient.Client, configYAML.Name, configYAML.Game, podToDelete.GetName(), "update")
```
Henrod: Same here. The pods are being deleted not because of the update, but because some error occurred during update.

reporters/reporter.go
```diff
@@ -0,0 +1,61 @@
...
+package reporters
+
+import (
+	"sync"
+
+	"github.com/Sirupsen/logrus"
+	"github.com/spf13/viper"
+)
+
+type Reporter interface {
```
Henrod: Before every public function, method, interface and struct we comment a brief description about it. For example,

//Reporter is the reporter interface

worker/worker.go
```diff
@@ -258,16 +258,20 @@ func (w *Worker) EnsureRunningWatchers(schedulerNames []string) {
...
 				configYaml, err = models.NewConfigYAML(configYamlStr)
 				if err == nil {
 					occupiedTimeout = configYaml.OccupiedTimeout
+					gameName = configYaml.Game
 				}
 			}
 			if err != nil {
 				l.Warnf("error loading scheduler %s: %s", schedulerName, err.Error())
 				occupiedTimeout = w.Config.GetInt64("occupiedTimeout")
+				gameName = w.Config.GetString("game")
```
Henrod: configYaml has an attribute Game

models/pod.go
```diff
@@ -145,6 +146,14 @@ func NewPod(
 		pod.ResourcesRequestsMemory = requests.Memory
 	}
 	err := pod.configureHostPorts(clientset, redisClient)
+
+	if err == nil {
+		reporters.GetInstance().Report("gru.new", map[string]string{
+			"name":      game,
+			"scheduler": namespace,
```
Henrod: Shouldn't this receive a reason?

worker/worker.go
ghostec: yes, I'm using it in the upper block (err == nil). I thought that this block was a fallback, so I repeated the existing behavior from the line above it, reading from w.Config. Is it wrong?

models/pod.go
ghostec: Nope. It only matters why a pod was deleted, creations are all put together in the same metric.

PR #23, opened 2017-10-07
Enable event forwarders for each scheduler.

cscatolini:
Each scheduler should be able to configure which event forwards it should use. This PR changes the list of EventForwarders to contain more information than only the forwarder function, such as the plugin and the forwarder name. It also adds a Forwarders field to the scheduler configuration struct. Forwarders that are disabled or not present in a scheduler configuration will not be triggered.


eventforwarder/forward.go
```diff
@@ -0,0 +1,100 @@
...
+	schedulerCache *models.SchedulerCache,
+) error {
+	if len(forwarders) > 0 {
+		metadata["roomId"] = room.ID
+		scheduler, err := schedulerCache.LoadScheduler(db, room.SchedulerName, true)
+		if err != nil {
+			return err
+		}
+		metadata["game"] = scheduler.Game
+		config, err := models.NewConfigYAML(scheduler.YAML)
```
Henrod: It would be better to use schedulerCache's method LoadConfigYaml. This save us another Unmarshal.

eventforwarder/forward.go
```diff
@@ -0,0 +1,100 @@
...
+	status string,
+	metadata map[string]interface{},
+	schedulerCache *models.SchedulerCache,
+) error {
+	if len(forwarders) > 0 {
+		scheduler, err := schedulerCache.LoadScheduler(db, room.SchedulerName, true)
+		if err != nil {
+			return err
+		}
+		config, err := models.NewConfigYAML(scheduler.YAML)
```
Henrod: It would be better to use schedulerCache's method LoadConfigYaml. This save us another Unmarshal.

cscatolini: Ok, I'll change it.

eventforwarder/forward.go
cscatolini: I'll change it too.

cscatolini:
@Henrod I made the changes you requested, could you take another look, please?

PR #30, opened 2017-10-20
watcher time ticker for rooms statuses report

ghostec:
Watcher.Start() now has a time ticker for rooms statuses reporting

### DIFF
--- a/watcher/watcher.go
+++ b/watcher/watcher.go
@@ -26,6 +26,7 @@ import (
 	"github.com/topfreegames/maestro/extensions"
 	"github.com/topfreegames/maestro/metadata"
 	"github.com/topfreegames/maestro/models"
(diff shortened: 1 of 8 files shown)


watcher/watcher.go
```diff
@@ -190,6 +186,88 @@ func (w *Watcher) Start() {
...
+
+	type RoomData struct {
+		Status string
+		Gauge  string
+	}
+
+	roomDataSlice := []RoomData{
+		RoomData{
+			models.StatusCreating,
+			fmt.Sprint(int(roomCountByStatus.Creating)),
```
Henrod: roomCountByStatus.Creating already is an int. Do you need to cast it?

ghostec: You're right!

PR #34, opened 2018-06-11
Adding support for "Accept: application/json" to route GET /scheduler…

matheuscscp:
…/{schedulerName}/config

### DIFF
--- a/api/scheduler_config_handler.go
+++ b/api/scheduler_config_handler.go
@@ -1,12 +1,14 @@
 package api
 
 import (
+	"encoding/json"
 	"errors"
 	"net/http"
 
+	yaml "gopkg.in/yaml.v2"
+
 	"github.com/Sirupsen/logrus"
 	"github.com/topfreegames/maestro/models"
-	ghodssYaml "github.com/ghodss/yaml"
 )
 
 // GetSchedulerConfigHandler handler returns the scheduler config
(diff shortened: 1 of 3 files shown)


api/scheduler_config_handler.go
```diff
@@ -6,6 +6,7 @@ import (
 
 	"github.com/Sirupsen/logrus"
 	"github.com/topfreegames/maestro/models"
+	ghodssYaml "github.com/ghodss/yaml"
```
Henrod: This import is not necessary. You can convert to json using the encoding/json package.

api/scheduler_config_handler.go
```diff
@@ -47,6 +48,20 @@ func (g *GetSchedulerConfigHandler) ServeHTTP(w http.ResponseWriter, r *http.Req
 		return
 	}
 
+	// json support
+	acceptHeader := r.Header.Get("Accept")
+	if acceptHeader == "application/json" {
```
Henrod: This is not necessary and can add some headache.

api/scheduler_config_handler_test.go
```diff
@@ -67,13 +68,29 @@ var _ = Describe("SchedulerConfigHandler", func() {
...
+		})
+
 		It("should return config yaml of specified version", func() {
 			version := "v1.0"
 			MockSelectYamlWithVersion(yamlString, version, mockDb, nil)
 
 			url = fmt.Sprintf("http://%s/scheduler/%s/config?version=%s",
 				app.Address, configYaml.Name, version)
 			request, _ = http.NewRequest("GET", url, nil)
+			request.SetBasicAuth(user, pass)
```
Henrod: By removing on BeforeEach and adding this, you are removing the test with Oauth. 
Please, add a new test that uses oauth.

api/scheduler_config_handler_test.go
```diff
@@ -40,17 +42,16 @@ var _ = Describe("SchedulerConfigHandler", func() {
 			Do(func(destToken *login.DestinationToken, query string, modifier string) {
 				destToken.RefreshToken = "refresh-token"
 			}).AnyTimes()
-		mockLogin.EXPECT().
```
Henrod: It is important to test with oauth too.

api/scheduler_config_handler.go
matheuscscp: Even if the input byte stream is a YAML text?

Henrod: yes

PR #35, opened 2018-06-12
Using github.com/ghodss/yaml to convert from YAML to JSON (this pkg h…

matheuscscp:
…andles non 1:1 cases)


andrehp:
Please resolve the conflicts, @Henrod modified the same parts as you

PR #39, opened 2018-07-11
Adding sort to releases route

andrehp:


Henrod:
Please, fix the tests. Also, change the version on Chart.yaml

models/scheduler.go
```diff
@@ -356,7 +356,7 @@ func ListSchedulerReleases(db interfaces.DB, schedulerName string) (
 ) {
 	_, err = db.Query(
 		&versions,
-		"SELECT version, created_at FROM scheduler_versions WHERE name = ?",
+		"SELECT version, created_at FROM scheduler_versions WHERE name = ? ORDER BY version",
```
Henrod: The order by should be `ORDER BY created_at ASC` because version is a string

cscatolini: we can order by strings as well

Henrod: It wouldn't work for this case:
```
v10.0
v100.0
v11.0

cscatolini: you're right :(

andrehp: Agreed, will fix this

andrehp:
Done

PR #41, opened 2018-08-07
Implement RoomAddrGetter

lftakakura:
Implement RoomAddrGetter to enable all Maestro functionalities using Minikube on development environment

### DIFF
--- a/api/app.go
+++ b/api/app.go
@@ -429,6 +429,7 @@ func (a *App) loadConfigurationDefaults() {
 	a.Config.SetDefault("api.limitManager.keyTimeout", 1*time.Minute)
 	a.Config.SetDefault("jaeger.disabled", false)
 	a.Config.SetDefault("jaeger.samplingProbability", 1.0)
+	a.Config.SetDefault(EnvironmentConfig, ProdEnvironment)
 }
 
 func (a *App) configureJaeger() {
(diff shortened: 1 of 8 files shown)


api/app.go
```diff
@@ -542,6 +545,14 @@ func (a *App) configureServer(showProfile bool) {
 	}
 }
 
+func (a *App) configureRoomAddrGetter() {
+	if a.Config.GetString("environment") != "development" {
```
Henrod: Save these strings as variables on constants file

models/room_address.go
```diff
@@ -0,0 +1,110 @@
...
+		return nil, maestroErrors.NewKubernetesError("no ports found", errors.New("no node port found to host room"))
+	}
+	return rAddresses, nil
+}
+
+// RoomAddressesFromNodePort is the struct that defines room addresses in development (using NodePort service)
+type RoomAddressesFromNodePort struct{}
+
+// Get gets room public addresses
+func (r *RoomAddressesFromNodePort) Get(room *Room, kubernetesClient kubernetes.Interface) (*RoomAddresses, error) {
```
Henrod: The two Get methods have a lot in common, maybe they can use a common private function

config/local.yaml
```diff
@@ -1,3 +1,4 @@
+environment: "development"
```
Henrod: default value must be production

lftakakura: ok

PR #47, opened 2018-08-28
Feature/auto scaling with usage

lftakakura:


watcher/watcher.go
```diff
@@ -600,75 +588,208 @@ func (w *Watcher) checkState(
...
+
+		if isDown {
+			usage = 1 - usage
+			point = roomCount.Ready
+			unbalancedState = models.StateOverdimensioned
+		}
+
+		// If usage is above limit, scale up.
+		if isDown == false && 100*roomCount.Occupied >= roomCount.Total()*trigger.Limit {
+			w.Logger.Debug("Usage is above limit. Should scale up")
```
Henrod: This can be extracted to another function. The idea is reduce the size of this function and simplify it

watcher/watcher.go
```diff
@@ -600,75 +588,208 @@ func (w *Watcher) checkState(
...
+		usage = 1 - usage
+		point = roomCount.Ready
+		unbalancedState = models.StateOverdimensioned
+	}
+
+	// If usage is above limit, scale up.
+	if 100*roomCount.Occupied >= roomCount.Total()*autoScalingInfo.Up.Trigger.Limit {
+		scaleInfo.SendUsage(
+			w.SchedulerName,
+			"legacy", point, roomCount.Total(),
```
Henrod: Better to put "legacy" on constants to avoid typos

watcher/utils.go
```diff
@@ -0,0 +1,35 @@
...
+		return 0, fmt.Errorf("Undefined metric: %s", metricTrigger.Metric)
+	}
+}
+
+func roomDynamicDelta(
+	metricTrigger *models.ScalingPolicyMetricsTrigger,
+	roomCount *models.RoomsStatusCount,
+) int { // delta
+	usageThreshold := float32(metricTrigger.Usage) / 100
+	return int(float32(float32(roomCount.Occupied)-float32(usageThreshold*float32(roomCount.Total()))) / usageThreshold)
```
cscatolini: shouldn't we round up instead of just converting to int?

watcher/watcher.go
```diff
@@ -600,75 +588,208 @@ func (w *Watcher) checkState(
 	roomCount *models.RoomsStatusCount,
 	scheduler *models.Scheduler,
 	nowTimestamp int64,
-) (bool, bool, bool, error) { //shouldScaleUp, shouldScaleDown, changedState
+) (int, bool, bool, bool, error) { //delta, shouldScaleUp, shouldScaleDown, changedState
```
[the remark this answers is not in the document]
cscatolini: can't scaleup/down be inferred from the delta sign? if positive we should scaleup and if negative we should scaledown

watcher/watcher.go
```diff
@@ -600,75 +588,208 @@ func (w *Watcher) checkState(
...
+		// Down
+		if len(autoScalingInfo.Down.MetricsTrigger) > 0 {
+			delta, shouldScaleDown, changedState, inSync, err = w.checkMetricsTrigger(autoScalingInfo, autoScalingInfo.Down, w.ScaleDownInfo, roomCount, scheduler, nowTimestamp)
+		} else {
+			delta, shouldScaleDown, changedState, inSync, err = w.checkLegacyTrigger(autoScalingInfo, autoScalingInfo.Down, w.ScaleDownInfo, roomCount, scheduler, nowTimestamp)
+		}
+	} else if inSync == true { // always send current usage
+		if len(autoScalingInfo.Down.MetricsTrigger) > 0 {
+			for _, trigger := range autoScalingInfo.Down.MetricsTrigger {
+				w.ScaleDownInfo.SendUsage(
```
cscatolini: I couldn't find any sendUsage for scale up, it this intended? or am I missing something?

watcher/watcher.go
cscatolini: words of wisdom

dev-room/app.py
```diff
@@ -37,6 +48,7 @@ def ping():
     except Exception as ex:
         print(str(ex))
         pass
+        break
```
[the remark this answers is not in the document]
lftakakura: it shouldn't have ```break```. Forgot to take it off =/

watcher/watcher.go
lftakakura: On scaleUp it will call SendUsageAndReturnStatus(). SendUsage() is called here because checkMetricsTrigger will not be called for down scaling if up scaling is needed.

cscatolini: so we have two separate usage metrics, one for scaling up and another for scaling down?

watcher/utils.go
lftakakura: yes

watcher/watcher.go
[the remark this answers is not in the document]
Henrod: I agree

lftakakura:
The changes are backwards compatible. The only issue is with the redis set that now is the same for both up and down scaling.

watcher/utils.go
```diff
@@ -0,0 +1,42 @@
...
+
+import (
+	"fmt"
+	"math"
+
+	"github.com/topfreegames/maestro/models"
+)
+
+func dynamicDelta(
+	metricTrigger *models.ScalingPolicyMetricsTrigger,
```
Henrod: You could use ScalingPolicyMetricsTrigger as an interface actually, with the methods Metric() and Usage(). Or with the method DynamicDelta(). That way you won't need a switch/case statement.

watcher/watcher.go
```diff
@@ -600,75 +594,206 @@ func (w *Watcher) checkState(
 	roomCount *models.RoomsStatusCount,
 	scheduler *models.Scheduler,
 	nowTimestamp int64,
-) (bool, bool, bool, error) { //shouldScaleUp, shouldScaleDown, changedState
-	var shouldScaleUp, shouldScaleDown, changedState bool
-	inSync := true
+) (*scaling, error) { //delta, shouldScaleUp, shouldScaleDown, changedState
```
Henrod: We don't need this comment anymore ;)

watcher/watcher.go
```diff
@@ -600,75 +594,206 @@ func (w *Watcher) checkState(
...
+	if autoScalingInfo.Max > 0 && roomCount.Total() > autoScalingInfo.Max {
+		scaling.Delta = autoScalingInfo.Max - roomCount.Total()
+		return scaling, nil
 	}
 
-	threshold := autoScalingInfo.Up.Trigger.Threshold
-	usage := float32(autoScalingInfo.Up.Trigger.Usage) / 100
-	capacity := autoScalingInfo.Up.Trigger.Time / w.AutoScalingPeriod
+	// Send current usage
+	if len(autoScalingInfo.Down.MetricsTrigger) > 0 {
```
Henrod: This method is too long. You could separate the ifs into methods whose names are their intention.

watcher/watcher.go
```diff
@@ -600,75 +594,206 @@ func (w *Watcher) checkState(
...
+				scaling.Delta, err = dynamicDelta(trigger, roomCount)
+				return scaling, err
+			}
+		}
+	}
+
+	return scaling, nil
+}
+
+func (w *Watcher) checkLegacyTrigger(
```
Henrod: There is too much duplicated code with checkMetricsTrigger. Couldn't you merge them? For example, The for loop on checkMetricsTrigger could be extracted and receive the same parameters checkLegacyTrigger uses.

watcher/utils.go
lftakakura: Will do

watcher/watcher.go
lftakakura: will do

lftakakura:
Please ignore the Println/Printf lines. It will all be removed before merging

autoscaler/cpu_and_mem_scaler.go
```diff
@@ -0,0 +1,103 @@
...
+// GetCurrentUtilization returns the current usage percentage
+func (sp *ResourceUsagePolicy) GetCurrentUtilization(roomCount *models.RoomsStatusCount) (currrentUtilization float32) {
+	var resourceUsagePerPodList []float32
+	var resourceRequestsPerPodList []float32
+
+	podList, _ := sp.metricsClientset.Metrics().PodMetricses(sp.schedulerName).List(metav1.ListOptions{})
+
+	for _, pod := range podList.Items {
+		for i, container := range pod.Containers {
+			if container.Name == pod.Name { // Get game room container
```
cscatolini: probably this is not the best way to check

api/app.go
```diff
@@ -56,6 +57,7 @@ type App struct {
 	DBClient         *pg.Client
 	RedisClient      *redis.Client
 	KubernetesClient kubernetes.Interface
+	KubernetesMetricsClient metricsClient.Interface
```
Henrod: linter should align this

dev-room/api/increaseResourceUsage.py
```diff
@@ -0,0 +1,32 @@
+import falcon
```
cscatolini: @lftakakura this is the best file in the project 🤣

PR #50, opened 2018-10-01
Read ipv6 label from node and send it via eventforwarder

lftakakura:


models/room.go
```diff
@@ -30,6 +30,9 @@ const (
 	RoomTerminating = "terminating"
 	// RoomTerminated string representation
 	RoomTerminated = "terminated"
+
+	// Ipv6LabelKey is the label key on kubernetes
+	Ipv6LabelKey = "tfg.io/ipv6"
```
Henrod: You should not use tfg as key. This prefix should be gotten from config file

models/room.go
```diff
@@ -42,8 +45,9 @@ type Room struct {
 
 // RoomAddresses struct
 type RoomAddresses struct {
-	Ports []*RoomPort `json:"ports"`
-	Host  string      `json:"host"`
+	Ports     []*RoomPort `json:"ports"`
+	Host      string      `json:"host"`
+	Ipv6Label string      `json:"Ipv6Label"`
```
Henrod: `json:"ipv6Label"`

Henrod:
Good. But the testes are failing

lftakakura:
> Good. But the testes are failing

I know. Should be ok now!

PR #51, opened 2018-10-02
Send ipv6label metrics

lftakakura:


eventforwarder/forward.go
```diff
@@ -345,6 +355,53 @@ func reportRPCStatus(
...
+		logger.
+			WithField("operation", "reportIpv6Status").
+			WithError(err).
+			Error("failed to report Ipv6 read status to StatsD")
+		return
+	}
+
+	game := scheduler.ConfigYAML.Game
+
+	status := map[string]interface{}{
```
Henrod: It is only necessary to inform NodeHost and success or failed

PR #56, opened 2018-12-04
Feature/servers by metric

cscatolini:
This PR adds a route that returns a list of rooms ordered by a metric. For `legacy` and `room` metrics it returns at most `limit` available rooms. For `cpu` and `mem` metrics it returns at most `limit` rooms ordered by increasing cpu or memory usage.

Also, I'm now reporting room usage metrics (cpu and memory) when a ping event is received if the corresponding metrics trigger is enabled.

Last but not least, I've fixed a bug that allowed a scheduler to be updated with metrics autoscaling bu no requests specified.

In a later PR we might change the cpu and memory autoscaler to retrieve the data we're now storing in Redis instead of calling Kubernetes API.

### DIFF
--- a/models/room.go
+++ b/models/room.go
@@ -9,6 +9,7 @@ package models
 
 import (
 	"fmt"
+	"math"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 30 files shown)


models/room.go
```diff
@@ -337,3 +405,55 @@ func GetRoomsOccupiedTimeout(redisClient interfaces.RedisClient, schedulerName s
 	})
 	return result, err
 }
+
+func getReadyRooms(tx redis.Pipeliner, schedulerName string, size int, mr *MixedMetricsReporter) ([]string, error) {
+	var result []string
+	rooms := tx.SScan(
```
Henrod: Why SScan and not SRandMember?

PR #58, opened 2019-02-20
Save room metricses (cpu, mem) on redis sorted set via watchers

lftakakura:
The room metrics are obtained via kubernetes metrics client and saved in a redis sorted set for further listing the least busy rooms (in terms of cpu, memory and possibly other resources).

Until now this logic was incorporated in the ping route, but it was increasing its response time. This PR aims to transfer this logic to the watchers.

The idea is to add one more routine to the watchers (like AutoScale and RemoveDeadRooms) named AddUtilizationMetricsToRedis that runs every tick.


models/scheduler.go
```diff
@@ -371,6 +374,39 @@ func (c *Scheduler) GetResourcesRequests() map[AutoScalingPolicyType]int64 {
...
+
+	if len(rooms) == 0 || len(usages) == 0 {
+		return nil
+	}
+
+	pipe := redisClient.TxPipeline()
+	for i, room := range rooms {
+		pipe.ZAdd(
+			GetRoomMetricsRedisKey(c.Name, string(metric)),
+			redis.Z{Member: room.ID, Score: float64(usages[i])},
```
cscatolini: are we checking somewhere that `len(rooms) == len(usages)`?

models/scheduler.go
```diff
@@ -371,6 +374,39 @@ func (c *Scheduler) GetResourcesRequests() map[AutoScalingPolicyType]int64 {
...
+		)
+	}
+
+	err := mr.WithSegment(SegmentPipeExec, func() error {
+		var err error
+		_, err = pipe.Exec()
+		return err
+	})
+	if err != nil {
```
cscatolini: `return err` is enough, we don't need this if here

watcher/watcher.go
```diff
@@ -240,9 +256,113 @@ func (w *Watcher) reportRoomsStatusesRoutine() {
 func (w *Watcher) watchRooms() error {
 	w.RemoveDeadRooms()
 	w.AutoScale()
+	w.AddUtilizationMetricsToRedis()
```
cscatolini: Nice one! 
My idea was actually to add the metrics inside `w.AutoScale()` but this way is much cleaner.

watcher/watcher.go
```diff
@@ -240,9 +256,113 @@ func (w *Watcher) reportRoomsStatusesRoutine() {
...
+		logger,
+		w.MetricsReporter,
+		w.DB,
+		w.RedisClient.Client,
+		w.SchedulerName,
+	)
+
+	if err != nil {
+		if strings.Contains(err.Error(), "not found") {
+			w.Run = false
```
cscatolini: why should we stop the watcher if a scheduler is not found?

watcher/watcher.go
```diff
@@ -240,9 +256,113 @@ func (w *Watcher) reportRoomsStatusesRoutine() {
...
+		pmetricsList, err = w.KubernetesMetricsClient.Metrics().PodMetricses(w.SchedulerName).List(metav1.ListOptions{})
+		return err
+	})
+	if err != nil {
+		logger.WithError(err).Error("failed to list pods metricses")
+		return
+	}
+
+	// cant get metrics from pods yet. List pods created and set usage=math.MaxInt64 for them
+	if err != nil || pmetricsList == nil || len(pmetricsList.Items) == 0 {
```
cscatolini: we won't reach here if `err != nil`, right?

watcher/watcher.go
```diff
@@ -240,9 +256,113 @@ func (w *Watcher) reportRoomsStatusesRoutine() {
...
+		if err != nil {
+			logger.WithError(err).Error("failed to list pods on namespace")
+			return
+		}
+	}
+
+	for metric := range metricsMap {
+		var usages []float64
+		var rooms []*models.Room
+		if (err != nil || pmetricsList == nil || len(pmetricsList.Items) == 0) && pods != nil {
```
cscatolini: which `err` is this?

watcher/watcher.go
```diff
@@ -240,9 +256,113 @@ func (w *Watcher) reportRoomsStatusesRoutine() {
...
+		var err error
+		pmetricsList, err = w.KubernetesMetricsClient.Metrics().PodMetricses(w.SchedulerName).List(metav1.ListOptions{})
+		return err
+	})
+	if err != nil {
+		logger.WithError(err).Error("failed to list pods metricses")
+		return
+	}
+
+	// cant get metrics from pods yet. List pods created and set usage=math.MaxInt64 for them
```
cscatolini: Will this be applied to all rooms in a scheduler? I mean, if we have 9 rooms with metrics and 1 without them, will we set MaxInt64 only for the one without metrics?

lftakakura: No. If we have 9 rooms with metrics and 1 without, it will only save the 9 rooms with metrics.
It will set MaxInt64 only if we have 0 rooms with metrics. It can happen if all rooms have recently started.

Maybe it is better to return an empty list when there is no room with metrics.

Do you think it is necessary to return all rooms and fill the ones without metrics with math.MaxInt64?

watcher/watcher.go
lftakakura: correct. Will remove this

watcher/watcher.go
lftakakura: Actually I've just copied the behaviour from `AutoScale()`. It will force the watcher to restart if the scheduler exists

models/scheduler.go
lftakakura: No. It can't be different but I'm gonna make it send only one struct with usage and room in it

cscatolini: Yep, I saw that this shouldn't happen but I'd rather have an array of structs instead of two arrays.

watcher/watcher.go
cscatolini: Ok, this makes sense. I made a confusion between the watcher and the worker.

watcher/watcher.go
cscatolini: It depends on how long it takes to start collecting the metrics. My concern is that we'll have ready rooms that won't be retrievable because we still don't have their metrics. If we use `math.MaxInt64` we put them with less priority but we still can use them if necessary.

watcher/watcher.go
```diff
@@ -240,9 +267,98 @@ func (w *Watcher) reportRoomsStatusesRoutine() {
...
+	}
+
+	// If it does not use metricsTriggers we dont need to save metrics
+	if len(metricsMap) == 0 {
+		return
+	}
+
+	// Load pods and set their usage to MaxInt64 for all resources
+	var pods *v1.PodList
+	err = w.MetricsReporter.WithSegment(models.SegmentMetrics, func() error {
```
cscatolini: actually, here it should be `models.SegmentPod` instead of `models.SegmentMetrics`

watcher/watcher.go
```diff
@@ -240,9 +267,98 @@ func (w *Watcher) reportRoomsStatusesRoutine() {
...
+		var err error
+		pmetricsList, err = w.KubernetesMetricsClient.Metrics().PodMetricses(w.SchedulerName).List(metav1.ListOptions{})
+		return err
+	})
+	if err != nil {
+		logger.WithError(err).Error("failed to list pods metricses")
+	}
+
+	for metric := range metricsMap {
```
cscatolini: This loop is much cleaner! Great job 👍

PR #59, opened 2019-02-22
Feature/exclude terminating rooms when listing by metrics

lftakakura:


watcher/watcher.go
```diff
@@ -35,9 +37,24 @@ import (
...
 )
 
+func createRoomUsages(pods *v1.PodList, availableRooms map[string]bool) ([]*models.RoomUsage, map[string]int) {
+	var roomUsages []*models.RoomUsage
+	roomUsagesIdxMap := map[string]int{}
+	for i, pod := range pods.Items {
+		if availableRooms[pod.Name] == true {
+			roomUsages = append(roomUsages, &models.RoomUsage{Name: pod.Name, Usage: float64(math.MaxInt64)})
+			roomUsagesIdxMap[pod.Name] = i
+			fmt.Println("createRoomUsages", roomUsagesIdxMap)
```
cscatolini: oops

watcher/watcher.go
```diff
@@ -35,9 +37,24 @@ import (
...
 	"k8s.io/client-go/kubernetes"
+	"k8s.io/metrics/pkg/apis/metrics/v1beta1"
 	metricsClient "k8s.io/metrics/pkg/client/clientset_generated/clientset"
 )
 
+func createRoomUsages(pods *v1.PodList, availableRooms map[string]bool) ([]*models.RoomUsage, map[string]int) {
+	var roomUsages []*models.RoomUsage
+	roomUsagesIdxMap := map[string]int{}
+	for i, pod := range pods.Items {
+		if availableRooms[pod.Name] == true {
```
cscatolini: all elements in the map are true, it would be better to return an array instead of a map

watcher/watcher_test.go
```diff
@@ -3678,7 +3684,9 @@ var _ = Describe("Watcher", func() {
 					configYaml.AutoScaling.Min,
 					configYaml.AutoScaling.Max,
 				)
+				fmt.Println("BEFORE")
```
cscatolini: oops

watcher/watcher_test.go
```diff
@@ -3678,7 +3684,9 @@ var _ = Describe("Watcher", func() {
...
 				simulateMetricsAutoscaling(simSpec, configYaml, yamlActive, models.StateInSync)
+				fmt.Println("AFTER")
```
cscatolini: oops

cscatolini:
@lftakakura please rebase with master (I merged the other PR)

api/room_handler.go
```diff
@@ -375,14 +375,35 @@ func (h *RoomListByMetricHandler) ServeHTTP(w http.ResponseWriter, r *http.Reque
...
+	readyAndOccupiedRooms, err := models.GetAllRegisteredRoomsByStatus(
+		h.App.RedisClient.Client,
+		params.SchedulerName,
+		[]string{models.StatusReady, models.StatusOccupied})
+
+	if err != nil {
+		logger.WithError(err).Error("failed to list available pods on namespace")
+	}
+
+	// remove terminating and creating rooms
```
cscatolini: The problem here is that, since you are filtering out unavailable rooms you'll probably end up returning less rooms that expected (`limit` param). Also, `GetAllRegisteredRoomsByStatus` can be quite expensive if a scheduler has many rooms, specially since this route can be called quite frequently. Could we pass a list of rooms and validate which are ready or occupied? Or maybe filter out creating/terminating ones? We'll probably have less rooms being returned in the second case.

The best approach IMO would be to receive an array of rooms and check using redis `mult` op of they fit the criteria (ready or occupied). If we have less rooms than necessary try it again.

Finally, I'd move this to `models.GetRoomsByMetric` since this won't be useful if we're using this if metric is not cpu or mem.

models/room.go
```diff
@@ -394,26 +380,81 @@ func getReadyRooms(tx redis.Pipeliner, schedulerName string, size int, mr *Mixed
 	return res, err
 }
 
+func filterReadyAndOccupiedRooms(tx redis.Pipeliner, mr *MixedMetricsReporter, availableRooms *[]string, rooms []string, schedulerName string) error {
```
cscatolini: Why receive `availableRooms`? I would rather build the slice and return it in this function.

models/room_status.go
```diff
@@ -132,3 +132,32 @@ func GetAllRegisteredRooms(
 
 	return registeredRooms, nil
 }
+
+// GetAllRegisteredRoomsByStatus returns an map of registered pods on redis with determined status
```
cscatolini: this method is no longer necessary, right?

PR #64, opened 2019-06-04
Room Addresses caching & go-extensions-k8s-client-go update

ghostec:
* local cache with configurable expiration/cleanup times.
* updating go-extensions-k8s-client-go to tag `kubernetes-1.11.7` that uses reflection to set restClient in all components

### DIFF
--- a/models/room_address.go
+++ b/models/room_address.go
@@ -8,12 +8,15 @@
 package models
 
 import (
+	"encoding/json"
 	"errors"
 	"fmt"
 	"time"
 
 	"github.com/btcsuite/btcutil/base58"
-	"github.com/pmylund/go-cache"
+	"github.com/go-redis/redis"
+	"github.com/sirupsen/logrus"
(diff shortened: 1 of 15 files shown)


models/room_address.go
```diff
@@ -8,48 +8,199 @@
...
+	}
+	if err := json.Unmarshal(b, addrs); err != nil {
+		l.WithError(err).Error("failed unmarshal room addresses from redis")
+		return nil
+	}
+	return addrs
+}
+
+func (r *RoomAddressesFromHostPort) buildCacheKey(room *Room) string {
+	return fmt.Sprintf("%s-%s", room.SchedulerName, room.ID)
```
cscatolini: I'm pretty sure this could conflict with other keys. Maybe something a bit more specific like
`fmt.Sprintf("addr-%s-%s", room.SchedulerName, room.ID) `

ghostec: makes complete sense, will do

PR #66, opened 2019-08-07
Feature/rolling update

lftakakura:
Deploy pods before deleting them in a rolling-update respecting maxSurge

### DIFF
--- a/controller/utils.go
+++ b/controller/utils.go
@@ -48,6 +48,8 @@ func replacePodsAndWait(
 
 	// create a chunk of pods (chunkSize = maxSurge)
 	for range podsToDelete {
+		logger.Debug("creating pods")
+
 		newPod, err := roomManager.Create(logger, mr, redisClient,
 			db, clientset, configYAML, scheduler)
 
@@ -58,11 +60,10 @@ func replacePodsAndWait(
 		}
 	}
 
-	now := clock.Now()
-	timeout := willTimeoutAt.Sub(now)
-	deletedPods, timedout, canceled = waitCreatingAndDeleteOldPods(
-		logger, clientset, redisClient, timeout, configYAML,
-		createdPods, podsToDelete, roomManager, operationManager, mr)
+	timeout := willTimeoutAt.Sub(clock.Now())
+	timedout, canceled = waitCreatingPods(
+		logger, clientset, timeout, configYAML.Name,
+		createdPods, operationManager, mr)
 	if timedout || canceled {
 		return createdPods, deletedPods, timedout, canceled
 	}
@@ -78,6 +79,14 @@ func replacePodsAndWait(
 		if err != nil {
 			logger.WithError(err).Debugf("error deleting pod %s", pod.GetName())
 		}
+	}
+
+	timeout = willTimeoutAt.Sub(clock.Now())
+	timedout, canceled = waitTerminatingPods(
+		logger, clientset, timeout, configYAML.Name,
+		deletedPods, operationManager, mr)
+	if timedout || canceled {
+		return createdPods, deletedPods, timedout, canceled
 	}
 
 	return createdPods, deletedPods, false, false
@@ -140,13 +149,12 @@ func rollback(
 			}
 
 			waitTimeout := willTimeoutAt.Sub(time.Now())
-			err = waitTerminatingPods(
+			timedout, _ := waitTerminatingPods(
 				logger, clientset, waitTimeout, configYAML.Name,
-				createdPodChunks[i],
-				mr,
+				createdPodChunks[i], nil, mr,
 			)
-			if err != nil {
-				return err
+			if timedout {
+				return errors.New("timeout waiting for rooms to be removed")
 			}
 		}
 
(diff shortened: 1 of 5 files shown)


andrehp:
Would it add too much complexity to allow maxSurge and maxUnavailable on the same config like deployments do?

cscatolini:
> Would it add too much complexity to allow maxSurge and maxUnavailable on the same config like deployments do?

I'd rather not change any behaviour and/or scheduler config at this moment @andrehp. Do you think this is a must have?

andrehp:
> > Would it add too much complexity to allow maxSurge and maxUnavailable on the same config like deployments do?
> 
> I'd rather not change any behaviour and/or scheduler config at this moment @andrehp. Do you think this is a must have?

I wouldn't say it's a must have, but it might be nice to speed up deploys, though I do agree it might not be the best idea to change this behavior right now.

> I agree @cscatolini. Maestro already has the query string maxsurge which suffices many cases

I did a quick search on the docs about maxsurge but didn't find anything. But I think maestro's maxsurge is different from a k8s deployment's, isn't it?

controller/utils.go
```diff
@@ -166,81 +178,65 @@ func rollback(
...
 	ticker := time.NewTicker(1 * time.Second)
 	defer ticker.Stop()
 
-	i := 0
 	for {
 		exit := true
 		select {
 		case <-ticker.C:
-			if operationManager.WasCanceled() {
+			if operationManager != nil && operationManager.WasCanceled() {
```
cscatolini: in which scenario operationManager would be nil?

controller/utils.go
```diff
@@ -276,51 +271,83 @@ func waitTerminatingPods(
 		exit := true
 		select {
 		case <-ticker.C:
-			for _, pod := range deletedPods {
+			if operationManager.WasCanceled() {
```
cscatolini: shouldn't you have the same check here?
`if operationManager != nil && operationManager.WasCanceled() {`

controller/utils.go
```diff
@@ -276,51 +271,83 @@ func waitTerminatingPods(
...
+						logger.
+							WithError(err).
+							WithField("pod", pod.GetName()).
+							Errorf("error recreating pod")
+					}
+					break
+				}
+
+				if len(createdPod.Status.Phase) == 0 {
+					//HACK! Trying to detect if we are running unit tests
```
cscatolini: can't this have any unintended side effects? Maybe add a logger here?

cscatolini:
> > > Would it add too much complexity to allow maxSurge and maxUnavailable on the same config like deployments do?
> > 
> > 
> > I'd rather not change any behaviour and/or scheduler config at this moment @andrehp. Do you think this is a must have?
> 
> I wouldn't say it's a must have, but it might be nice to speed up deploys, though I do agree it might not be the best idea to change this behavior right now.

You can talk to @lftakakura. We don't want major changes but we believe that creating 25% new pods before deleting previous versions may be a big burden in our infrastructure. He is thinking about some way to speed the process and will update this PR soon.

> 
> > I agree @cscatolini. Maestro already has the query string maxsurge which suffices many cases
> 
> I did a quick search on the docs about maxsurge but didn't find anything. But I think maestro's maxsurge is different from a k8s deployment's, isn't it?

Yes. It is a global maestro's config instead of per scheduler. The default we're using in prod is 25%.

controller/utils.go
lftakakura: operationManger is nil when rolling back (rollback can't be canceled)

controller/utils.go
lftakakura: Yes it should

controller/utils.go
```diff
@@ -166,81 +178,65 @@ func rollback(
...
-					logger.WithField("pod", pod.GetName()).Debugf("pod still exists")
-					exit = false
-					break
-				}
 
-				newPod, err := roomManager.Create(logger, mr, redisClient,
-					db, clientset, configYAML, scheduler)
-				if err != nil {
+				if err == nil || !strings.Contains(err.Error(), "not found") {
+					logger.WithField("pod", pod.GetName()).Debugf("pod still exists, deleting again")
```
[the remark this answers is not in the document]
lftakakura: Updated

controller/utils.go
lftakakura: Updated

controller/utils.go
```diff
@@ -39,41 +39,59 @@ func replacePodsAndWait(
...
 	scheduler *models.Scheduler,
 	operationManager *models.OperationManager,
 ) (createdPods []v1.Pod, deletedPods []v1.Pod, timedout, canceled bool) {
 	createdPods = []v1.Pod{}
 	deletedPods = []v1.Pod{}
 
-	for _, pod := range podsToDelete {
-		logger.Debugf("deleting pod %s", pod.GetName())
+	// create a chunk of pods (chunkSize = maxSurge)
+	for _, pod := range podsChunk {
```
ghostec: This way I think maxSurge is locked to 1 no matter what. What needs to be done is invoking the block inside this for in routines and waiting on all of them (using channel or sync.WaitGroup) to finish before continuing the loop

lftakakura: What you're suggesting is for each pod in the chunk to be created in parallel? If so, we could have the problem of creating too many pods at once

PR #67, opened 2019-08-12
Feature/allow scaling with scheduler update

lftakakura:
Still need to fix some unit tests

### DIFF
--- a/controller/constants.go
+++ b/controller/constants.go
@@ -1,0 +1,18 @@
+package controller
+
+import "fmt"
+
+// SchedulerVersions rolling update statuses
+const (
+	timedoutStatus   = "timed out"
+	erroredStatus    = "error"
+	canceledStatus   = "canceld"
+	inProgressStatus = "in progress"
+	deployedStatus   = "deployed"
+)
+
+var (
+	rollbackStatus = func(version string) string {
+		return fmt.Sprintf("rolled back to %s", version)
+	}
+)
--- a/controller/constants.go
+++ b/controller/constants.go
@@ -6,7 +6,7 @@ import "fmt"
 const (
 	timedoutStatus   = "timed out"
 	erroredStatus    = "error"
-	canceledStatus   = "canceld"
+	canceledStatus   = "canceled"
 	inProgressStatus = "in progress"
 	deployedStatus   = "deployed"
 )
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -14,7 +14,6 @@ import (
 	"strings"
 	"time"
 
-	redisLock "github.com/bsm/redis-lock"
 	goredis "github.com/go-redis/redis"
 	uuid "github.com/satori/go.uuid"
 	clockinterfaces "github.com/topfreegames/extensions/clock/interfaces"
@@ -22,7 +21,6 @@ import (
 	"github.com/topfreegames/extensions/redis"
 	redisinterfaces "github.com/topfreegames/extensions/redis/interfaces"
 	maestroErrors "github.com/topfreegames/maestro/errors"
-	"github.com/topfreegames/maestro/reporters"
 	reportersConstants "github.com/topfreegames/maestro/reporters/constants"
 	yaml "gopkg.in/yaml.v2"
 	v1 "k8s.io/api/core/v1"
@@ -356,6 +354,7 @@ func ScaleUp(
 		"scheduler": scheduler.Name,
 		"amount":    amount,
 	})
+
 	configYAML, _ := models.NewConfigYAML(scheduler.YAML)
 
 	existPendingPods, err := pendingPods(clientset, scheduler.Name, mr)
@@ -438,6 +437,8 @@ func ScaleDown(
 		"scheduler": scheduler.Name,
 		"amount":    amount,
 	})
+
+	// check if scheduler update is in place
 
 	willTimeoutAt := time.Now().Add(time.Duration(timeoutSec) * time.Second)
 
@@ -570,12 +571,9 @@ func UpdateSchedulerConfig(
 	operationManager *models.OperationManager,
 ) error {
 	configYAML.EnsureDefaultValues()
-
 	timeoutSec := config.GetInt("updateTimeoutSeconds")
-	lockTimeoutMS := config.GetInt("watcher.lockTimeoutMs")
-	lockKey := models.GetSchedulerLockKey(config.GetString("watcher.lockKey"), configYAML.Name)
-	maxVersions := config.GetInt("schedulers.versions.toKeep")
-
+	timeoutDur := time.Duration(timeoutSec) * time.Second
+	willTimeoutAt := clock.Now().Add(timeoutDur)
 	schedulerName := configYAML.Name
 	l := logger.WithFields(logrus.Fields{
 		"source":    "updateSchedulerConfig",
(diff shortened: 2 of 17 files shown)


controller/controller.go
```diff
@@ -570,12 +571,9 @@ func UpdateSchedulerConfig(
 	operationManager *models.OperationManager,
 ) error {
 	configYAML.EnsureDefaultValues()
-
 	timeoutSec := config.GetInt("updateTimeoutSeconds")
-	lockTimeoutMS := config.GetInt("watcher.lockTimeoutMs")
-	lockKey := GetLockKey(config.GetString("watcher.lockKey"), configYAML.Name)
-	maxVersions := config.GetInt("schedulers.versions.toKeep")
-
+	timeoutDur := time.Duration(timeoutSec) * time.Second
```
ghostec: UpdateSchedulerConfig does so much work that is incredibly hard to follow :/ However, I'm not sure if it's worth looking into that at this point.

controller/utils.go
```diff
@@ -845,3 +889,366 @@ func validateMetricsTrigger(configYAML *models.ConfigYAML, logger logrus.FieldLo
...
+	downScalingLockKey := GetDownScalingLockKey(config.GetString("watcher.lockKey"), schedulerName)
+	timeout, _ := time.ParseDuration("5ms")
+	lock, err := redisClient.EnterCriticalSection(
+		redisClient.Trace(ctx),
+		downScalingLockKey,
+		timeout,
+		0, 0,
+	)
+	if err != nil || lock != nil {
+
```
ghostec: nit: unnecessary blank lines

controller/utils.go
```diff
@@ -845,3 +889,366 @@ func validateMetricsTrigger(configYAML *models.ConfigYAML, logger logrus.FieldLo
 	}
 	return nil
 }
+
+func schedulerAndConfigFromName(
```
ghostec: this and all functions below it are really new functions?

controller/controller.go
```diff
@@ -671,202 +648,144 @@ waitForLock:
...
-		var kubePods *v1.PodList
-		err := mr.WithSegment(models.SegmentPod, func() error {
-			var err error
-			kubePods, err = clientset.CoreV1().Pods(schedulerName).List(metav1.ListOptions{
-				LabelSelector: labels.Set{}.AsSelector().String(),
-				FieldSelector: fields.Everything().String(),
-			})
-			return err
-		})
+	scheduler.RollingUpdateStatus = "in progress"
```
ghostec: RollingUpdateStatus should not be magic strings. Maybe make it an enum (in PG), and use them in code from consts.

controller/utils.go
lftakakura: They already existed in controller. I just separated the code in functions and put it in utils

ghostec: got it! 💯

controller/controller.go
lftakakura: I decided to leave it as free text. It doesn't has predefined statuses and it is more like a description (should be similar to deis). It has texts like `rolled back to version vXY.Z`.
Also, it should be only human readable.

andrehp: Having some consts would be better than having the strings directly in the code. This way it would at least be easier to avoid typos and have all possible statuses in a single place.

lftakakura: I agree.

models/service.go
```diff
@@ -59,7 +60,7 @@ func NewService(
 	if len(configYaml.Ports) <= 0 && len(configYaml.Containers) > 0 {
 		for i, container := range configYaml.Containers {
 			for j, port := range container.Ports {
-				port.Name = fmt.Sprintf("%s-%d%d", port.Name, i, j)
+				port.Name = fmt.Sprintf("%s-%d%d", strings.Split(port.Name, "-")[0], i, j)
```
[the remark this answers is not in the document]
lftakakura: It had a bug.
tldr - as configYAML was passed as reference, it was concatenating over the concatenated string multiple times for each pod created creating things like tcp-00-00-00-00-00 for port names.
This only affected local development

controller/controller.go
```diff
@@ -671,202 +664,155 @@ waitForLock:
...
-				rollErr := rollback(
-					l, roomManager, mr, db, redisClient.Client, clientset,
-					&oldConfig, maxSurge, 2*timeoutDur, createdPods, deletedPods,
-					scheduler, currentVersion)
-				if rollErr != nil {
-					l.WithError(rollErr).Debug("error during update roll back")
-					err = rollErr
-				}
-				return errors.New(errMsg)
+			if errored != nil {
```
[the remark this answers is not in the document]
lftakakura: It's not correct. Next commit will solve this

controller/constants.go
```diff
@@ -0,0 +1,18 @@
...
+
+import "fmt"
+
+// SchedulerVersions rolling update statuses
+const (
+	timedoutStatus   = "timed out"
+	erroredStatus    = "error"
+	canceledStatus   = "canceld"
+	inProgressStatus = "in progress"
+	deployedStatus   = "deployed"
```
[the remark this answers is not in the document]
lftakakura: We could but it would be redundant as the table row already has the version that was deployed in a version column

controller/controller.go
```diff
@@ -584,84 +582,79 @@ func UpdateSchedulerConfig(
...
+		return err
 	}
 
-	if configYAML.AutoScaling.Max > 0 && configYAML.AutoScaling.Min > configYAML.AutoScaling.Max {
-		return errors.New("invalid parameter: autoscaling max must be greater than min")
-	}
+	// Lock watchers so they don't scale up or down and the scheduler is not
+	// overwritten with older version on database
+	// during the databse write phase
+	globalLockKey := models.GetSchedulerLockKey(config.GetString("watcher.lockKey"), configYAML.Name)
```
ghostec: Maybe using better names might help on confusion. `GetSchedulerLockKey` could mean any type of key. `GetSchedulerScalingLockKey` is a bit more obvious.

controller/controller.go
```diff
@@ -584,84 +582,79 @@ func UpdateSchedulerConfig(
...
-		select {
-		case <-timeout.C:
-			return errors.New("timeout while wating for redis lock")
-		case <-ticker.C:
-			if operationManager.WasCanceled() {
-				l.Warn("operation was canceled")
-				return nil
-			}
+	// Lock updates on scheduler during all the process
+	configLockKey := models.GetConfigLockKey(config.GetString("watcher.lockKey"), configYAML.Name)
```
ghostec: The same about naming goes here. `models` is not exclusive to schedulers. What do you think about `GetSchedulerConfigLockKey` ?.

controller/controller.go
```diff
@@ -671,202 +664,152 @@ waitForLock:
...
+	releaseLock(
+		logger,
+		redisClient,
+		globalLock,
+		schedulerName,
+	)
+
+	if shouldRecreatePods {
+		// Lock down scaling so it doesn't interferer with rolling update surges
+		downScalingLockKey := models.GetDownScalingLockKey(config.GetString("watcher.lockKey"), configYAML.Name)
```
ghostec: `GetSchedulerDownScalingLockKey` :eyes:

cscatolini:
LGTM

If all tests are fixed I think we can merge this.

@Henrod would you like to take a look before we merge?

ghostec:
LGTM

PR #68, opened 2019-08-12
list and delete schedulers locks

ghostec:
* GET `/scheduler/{schedulerName}/locks`: {schedulerName} locks list with "Key" (redis), "TTLInSec" and "IsLocked".
* DELETE `/scheduler/{schedulerName}/locks/{lockName}`: delete {lockName} of {schedulerName}, behind Access/Auth middlewares.

### DIFF
--- a/api/scheduler_locks_handler.go
+++ b/api/scheduler_locks_handler.go
@@ -1,0 +1,48 @@
+package api
+
+import (
+	"encoding/json"
+	"net/http"
+
+	"github.com/sirupsen/logrus"
+	"github.com/topfreegames/extensions/middleware"
+	"github.com/topfreegames/maestro/models"
+)
+
+// SchedulerLocksListHandler handler returns the scheduler config
+type SchedulerLocksListHandler struct {
+	App *App
+}
+
+// NewSchedulerLocksListHandler returns an instance of SchedulerConfigHandler
+func NewSchedulerLocksListHandler(a *App) *SchedulerLocksListHandler {
+	h := &SchedulerLocksListHandler{App: a}
+	return h
+}
+
+func (h *SchedulerLocksListHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	params := schedulerParamsFromContext(r.Context())
+	l := middleware.GetLogger(r.Context())
+	logger := l.WithFields(logrus.Fields{
+		"source":    "schedulerLocksList",
+		"operation": "get locks",
+		"scheduler": params.SchedulerName,
+	})
+	logger.Debug("Getting scheduler locks")
+	locksPrefixes := []string{h.App.Config.GetString("watcher.lockKey")}
+	locks, err := models.ListSchedulerLocks(
+		h.App.RedisClient.Trace(r.Context()), params.SchedulerName, locksPrefixes,
+	)
+	if err != nil {
+		logger.WithError(err).Error("error listing scheduler locks")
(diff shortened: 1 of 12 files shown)


andrehp:
What's the implication of deleting a lock?

ghostec:
@andrehp That you'll be able to invoke actions that would otherwise timeout due to locking. If you delete a lock that you shouldn't just to override an ongoing operation, race conditions in `schedulers updates` and `rooms scaling` will probably happen.

andrehp:
And how dangerous are the implications of possible race conditions?

ghostec:
@andrehp My interpretation of this function https://github.com/topfreegames/maestro/blob/master/controller/controller.go#L557
is that a second update will not delete pods of an initial update that are created after https://github.com/topfreegames/maestro/blob/master/controller/controller.go#L691 this point.

api/scheduler_locks_handler.go
```diff
@@ -0,0 +1,94 @@
...
+	l := middleware.GetLogger(r.Context())
+	logger := l.WithFields(logrus.Fields{
+		"source":    "schedulerLocksList",
+		"operation": "get locks",
+		"scheduler": params.SchedulerName,
+	})
+	logger.Debug("Getting scheduler locks")
+	locksPrefixes := []string{h.App.Config.GetString("watcher.lockKey")}
+	locks, err := models.ListSchedulerLocks(
+		h.App.RedisClient.Trace(r.Context()), params.SchedulerName, locksPrefixes,
```
andrehp: Can't you pass the context to the function instead of wrapping the redis client?

api/scheduler_locks_handler.go
```diff
@@ -0,0 +1,94 @@
...
+	isLockValid := false
+	validLocksKeys := models.ListSchedulerLocksKeys(params.SchedulerName, locksPrefixes)
+	for _, possible := range validLocksKeys {
+		if possible == params.LockKey {
+			isLockValid = true
+			break
+		}
+	}
+	if !isLockValid {
+		logger.Info("trying to delete invaldi lock key")
```
andrehp: s/invaldi/invalid

api/scheduler_locks_handler.go
```diff
@@ -0,0 +1,94 @@
...
+	params := schedulerLockParamsFromContext(r.Context())
+	l := middleware.GetLogger(r.Context())
+	logger := l.WithFields(logrus.Fields{
+		"source":    "schedulerLocksDelete",
+		"operation": "delete lock",
+		"scheduler": params.SchedulerName,
+		"lock":      params.LockKey,
+	})
+	logger.Debug("Getting scheduler locks")
+	locksPrefixes := []string{h.App.Config.GetString("watcher.lockKey")}
```
andrehp: Why is this an array?

andrehp:
Good job, just some comments

api/scheduler_locks_handler.go
ghostec: why would that be better?

api/scheduler_locks_handler.go
ghostec: We'll have more locks (the other open PR)

api/scheduler_locks_handler.go
andrehp: I just think it's ugly for the handler to know of redis, but passing the context wouldn't fix this anyway, given that the handler is already the one managing the client :(

So yeah, I don't think it would matter much

ghostec: Agreed on everything

PR #69, opened 2019-08-14
Feature/add more locks

lftakakura:


cscatolini:
@lftakakura I believe this has more commits than intended, right? Could you rebase onto master and pick only the relevant commits before we can take a look and review?

cscatolini:
LGTM but I'd let @ghostec take a look before merging because I'm not sure I understand why he chose to use an array of prefixes and this PR is changing this behaviour.

ghostec:
@cscatolini @lftakakura the only lock we had before was {prefix}-{schedulerName}, so I thought what would change from one lock to the other would be the prefix. Since we're changing through functions the end result is the same, so it's ok to only expect one prefix now :)

ghostec:
LGTM

PR #70, opened 2019-08-15
Rollback on ErrImageNeverPull, ErrImagePullBackOff and CrashLoopBackOff

lftakakura:


metadata/version.go
```diff
@@ -8,7 +8,7 @@
 package metadata
 
 //Version of Maestro
-var Version = "8.1.1"
+var Version = "8.2.1"
```
andrehp: Why not 8.2.0?

models/constants.go
```diff
@@ -102,3 +102,10 @@ const Global = "global"
 
 // PodNotFitsHostPorts is a message when the pod's host port is no available in any node of the pool
 const PodNotFitsHostPorts = "PodFitsHostPorts"
+
+// InvalidPodWaitingStates are all the states that are not accepted in a waiting pod.
+var InvalidPodWaitingStates = []string{
+	"ErrImageNeverPull",
+	"ErrImagePullBackOff",
+	"CrashLoopBackOff",
```
andrehp: It might be useful to watch for `ErrInvalidImageName` as well

lftakakura: ok! thanks

metadata/version.go
lftakakura: 8.2.0 was released already. Forgot to change the metadata =P

PR #71, opened 2019-08-15
Fix OpManager progress tracking

lftakakura:
With the recent changes, the progress status broke

### DIFF
--- a/models/scheduler.go
+++ b/models/scheduler.go
@@ -248,7 +248,13 @@ func (c *Scheduler) NextMajorVersion() {
 	c.Version = fmt.Sprintf("v%d.0", major+1)
 }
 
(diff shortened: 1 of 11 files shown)


models/scheduler.go
```diff
@@ -248,7 +248,13 @@ func (c *Scheduler) NextMajorVersion() {
 	c.Version = fmt.Sprintf("v%d.0", major+1)
 }
 
-// NextMinorVersion increments the major version
+// PreviousMajorVersion decrements the major version
```
andrehp: Why would you need to decrement the major version?

lftakakura: I don't need anymore, gonna take it off. I was using it in `getOperationRollingProgress`

PR #73, opened 2019-08-20
Refactor/ensure correct rooms

lftakakura:
- Now watcher's EnsureCorrectRooms replaces pods using rolling strategy.
- When an updateConfig needs a rollback, the API only rolls back the scheduler on database. EnsureCorrectRooms should then replace pods asynchronously.

### DIFF
--- a/controller/utils.go
+++ b/controller/utils.go
@@ -36,6 +36,74 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+// SegmentAndReplacePods acts when a scheduler rolling update is needed.
+// It segment the list of current pods in chunks of size maxSurge and replace them with new ones
+func SegmentAndReplacePods(
+	ctx context.Context,
+	logger logrus.FieldLogger,
+	roomManager models.RoomManager,
+	mr *models.MixedMetricsReporter,
+	clientset kubernetes.Interface,
+	db pginterfaces.DB,
+	redisClient redisinterfaces.RedisClient,
+	willTimeoutAt time.Time,
+	configYAML *models.ConfigYAML,
+	pods []v1.Pod,
+	scheduler *models.Scheduler,
+	operationManager *models.OperationManager,
+	maxSurge int,
+	clock clockinterfaces.Clock,
+) (timeoutErr, cancelErr, err error) {
+	schedulerName := scheduler.Name
+	l := logger.WithFields(logrus.Fields{
+		"source":    "SegmentAndReplacePods",
+		"scheduler": schedulerName,
+	})
+
+	// segment pods in chunks
+	podChunks := segmentPods(pods, maxSurge)
+
+	for i, chunk := range podChunks {
+		l.Debugf("updating chunk %d: %v", i, names(chunk))
+
+		// replace chunk
+		timedout, canceled, errored := replacePodsAndWait(
+			l,
+			roomManager,
+			mr,
+			clientset,
+			db,
+			redisClient,
+			willTimeoutAt,
+			configYAML,
+			chunk,
+			scheduler,
+			operationManager,
+			clock,
+		)
+
+		if timedout {
+			timeoutErr = errors.New("timedout waiting rooms to be replaced, rolled back")
+			l.WithError(err).Error(timeoutErr.Error())
+			break
+		}
+
+		if canceled {
+			cancelErr = errors.New("operation was canceled, rolled back")
+			l.WithError(err).Error(cancelErr.Error())
+			break
+		}
+
+		if errored != nil {
+			err = errored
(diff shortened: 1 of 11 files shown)


controller/utils.go
```diff
@@ -36,19 +36,87 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-func replacePodsAndWait(
+// SegmentAndReplacePods acts when a scheduler rolling update is needed.
+// It segment the list of current pods in chunks of size maxSurge and replace them with new ones
+func SegmentAndReplacePods(
```
ghostec: :100: awesome that you're splitting the bigger functions :)

controller/utils.go
```diff
@@ -157,55 +233,46 @@ func createNewRemoveOldPod(
...
-	l := logger.WithFields(logrus.Fields{
-		"source":    "updateSchedulerConfig.rollback",
-		"scheduler": schedulerName,
-	})
-
-	l.Debug("starting rollback")
-
-	globalLock, _, _ := acquireLock(
+) (err error) {
+	globalLock, _, _ := AcquireLock(
```
ghostec: globalLock could be nil, I think you need to check err. Don't you also need to check canceled?

watcher/watcher.go
```diff
@@ -1068,30 +1068,73 @@ func (w *Watcher) EnsureCorrectRooms() error {
...
+		w.SchedulerName,
+	)
+	defer controller.ReleaseLock(
+		logger,
+		w.RedisClient,
+		downScalingLock,
+		w.SchedulerName,
+	)
+
+	if err != nil {
```
ghostec: doesn't it make more sense to check err before deferring ReleaseLock?

controller/utils.go
```diff
@@ -36,19 +36,87 @@ import (
...
+
+		if canceled {
+			cancelErr = errors.New("operation was canceled, rolled back")
+			l.WithError(err).Error(cancelErr.Error())
+			break
+		}
+
+		if errored != nil {
+			err = errored
+			l.WithError(err).Error(err.Error())
```
Henrod: Here you are duplicating the message information. Prefer `l.WithError(err).Error("some tag string")` to be easier to find in the project.

controller/utils.go
```diff
@@ -126,7 +194,11 @@ func createNewRemoveOldPod(
 	}
 
 	// wait for new pod to be created
-	timeout := willTimeoutAt.Sub(clock.Now())
+	now := time.Now()
```
Henrod: Why would you pass a nil clock? Instead, pass a clock that returns `time.Now()`

watcher/watcher.go
```diff
@@ -1041,12 +1041,12 @@ func (w *Watcher) EnsureCorrectRooms() error {
 		return err
 	}
 
-	podsToDelete := []string{}
-	concat := func(podNames []string, err error) error {
+	podsToDelete := []v1.Pod{}
+	concat := func(pods []v1.Pod, err error) error {
```
Henrod: It's better to use pointer to struct than the struct itself

controller/utils.go
```diff
@@ -36,19 +36,87 @@ import (
...
+			configYAML,
+			chunk,
+			scheduler,
+			operationManager,
+			clock,
+		)
+
+		if timedout {
+			timeoutErr = errors.New("timedout waiting rooms to be replaced, rolled back")
+			l.WithError(err).Error(timeoutErr.Error())
```
cscatolini: Here and in line 93 `err` will be nil. I'd log:
```
l.WithError(timeoutErr).Error("failed to replace chunk of pods")
l.WithError(cancelErr).Error("failed to replace chunk of pods")
l.WithError(err).Error("failed to replace chunk of pods")
```

watcher/watcher.go
cscatolini: I think ReleaseLock is a noop if lock is nil, so it doesn't matter. But correct me if I am wrong @lftakakura

watcher/watcher.go
ghostec: 🤔 bold claim

watcher/watcher.go
lftakakura: @cscatolini it's right. It makes no difference as ReleaseLock can be called when a lock is not acquired

controller/utils.go
lftakakura: I actually don't need to global lock here anymore. dbRollback runs in UpdateSchedulerConfig, which already locks config (can't do more than 1 update at the same time).

EnsureCorrectRooms is not using global lock as well (it uses downscaling lock)

cscatolini:
@ghostec and @Henrod do you guys still have remarks with this PR or are we good to go?

PR #74, opened 2019-08-21
[WIP] hotfix+internal/plugin load k8s deps upgrade

ghostec:
By chasing a bug that showed up recently I've upgraded several dependencies. The solution to our crashes is installing "libc6-compat", and all the other changes are related to supporting kubernetes 1.13.

When maestro-api pods started, they most of the time crashed right after loading the plugin for grpc forwarding.

It's still in WIP state since I need to clean up some stuff and upgrade our `protos` repo as well.

I'll edit this description with some more details.


cscatolini:
@ghostec is this still a WIP? Aren't this changes necessary?

PR #75, opened 2019-08-22
Feature/rolling in remove dead rooms

lftakakura:
RemoveDeadRooms will not delete all rooms that have not performed ping for a while or have expired occupied timeout. Instead it will segment pods that are in these conditions and do rolling strategy to remove them (creating new ones before removing old).

### DIFF
--- a/controller/utils.go
+++ b/controller/utils.go
@@ -277,7 +277,7 @@ func waitTerminatingPods(
 	mr *models.MixedMetricsReporter,
 ) (timedout, wasCanceled bool) {
 	logger := l.WithFields(logrus.Fields{
-		"operation": "controller.waitTerminatingPods",
+		"source":    "controller.waitTerminatingPods",
 		"scheduler": namespace,
 	})
 
@@ -358,7 +358,7 @@ func waitCreatingPods(
 	mr *models.MixedMetricsReporter,
 ) (timedout, wasCanceled bool, err error) {
 	logger := l.WithFields(logrus.Fields{
-		"operation": "controller.waitCreatingPods",
+		"source":    "controller.waitCreatingPods",
 		"scheduler": namespace,
 	})
 
@@ -1026,7 +1026,7 @@ func AcquireLock(
 			}
 
 			if lock == nil {
-				l.Warnf("unable to get watcher %s lock, maybe some other process has it...", schedulerName)
+				l.Warnf("unable to get watcher %s lock %s, maybe some other process has it", schedulerName, lockKey)
 				break
 			}
 
--- a/watcher/watcher.go
+++ b/watcher/watcher.go
@@ -239,9 +239,9 @@ func (w *Watcher) Start() {
 	for w.Run == true {
 		select {
 		case <-ticker.C:
-			w.WithRedisLock(l, w.watchRooms)
+			w.watchRooms()
 		case <-tickerEnsure.C:
-			w.EnsureCorrectRooms()
+			w.WithDownscalingLock(w.EnsureCorrectRooms)
 		case <-tickerStateCount.C:
 			w.PodStatesCount()
 		case sig := <-sigchan:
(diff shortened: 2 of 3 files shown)


controller/utils.go
```diff
@@ -126,6 +127,7 @@ func replacePodsAndWait(
 	wg.Add(len(podsChunk))
 	for _, pod := range podsChunk {
 		go func(pod v1.Pod) {
+			defer GinkgoRecover()
```
cscatolini: It doesn't seem right to defer test related stuff in here. Why is this necessary?

watcher/watcher.go
```diff
@@ -268,14 +274,17 @@ func (w *Watcher) reportRoomsStatusesRoutine() {
 }
 
 func (w *Watcher) watchRooms() error {
-	w.RemoveDeadRooms()
-	w.AutoScale()
-	w.AddUtilizationMetricsToRedis()
+	l := w.Logger.WithFields(logrus.Fields{
+		"operation": "watcher.watchRooms",
+	})
+	w.WithRedisLock(l, w.AutoScale)
```
cscatolini: Why do we need full redis lock here?

watcher/watcher.go
```diff
@@ -268,14 +274,17 @@ func (w *Watcher) reportRoomsStatusesRoutine() {
...
+	w.WithDownscalingLock(l, w.RemoveDeadRooms)
+	w.WithRedisLock(l, w.AddUtilizationMetricsToRedis)
```
cscatolini: and here?

controller/utils.go
lftakakura: I need this to debug tests. It shouldn't have slipped through. My bad =(

watcher/watcher.go
lftakakura: I just kept this and AddUtilizationMetricsToRedis as it was. I'm going to think about using downScaling

controller/utils.go
lftakakura: Removed

ghostec:
Can you add a description of how your changes affect Maestro? :eyes:

watcher/watcher.go
cscatolini: ah okay! sorry, with the diff I didn't notice this was already AddUtilizationMetricsToRedis and AutoScale behaviour to lock everything

lftakakura:
> LGTM
> 
> It's up to you if you want to improve AddUtilizationMetricsToRedis and AutoScale locks in this PR or in another one. I'd change it in a separate PR so we can deploy this and validate the behaviour is as expected.

Ok. I'm already working on it. Gonna open a new PR

lftakakura:
> Can you add a description of how your changes affect Maestro? 👀

Sure

watcher/watcher.go
```diff
@@ -472,171 +509,221 @@ func (w *Watcher) WithRedisLock(l *logrus.Entry, f func() error) {
...
+	}
+
+	if len(roomsNoPingSince) > 0 || len(roomsOnOccupiedTimeout) > 0 {
+		podsToReplace, err := w.listPodsToReplace(logger, append(roomsNoPingSince, roomsOnOccupiedTimeout...))
+
+		// load scheduler from database
 		scheduler := models.NewScheduler(w.SchedulerName, "", "")
 		err = scheduler.Load(w.DB)
 		if err != nil {
 			logger.WithError(err).Error("error accessing db while removing dead rooms")
```
ghostec: Right after logging the error, all RemoveDeadRooms does is `logger.Info("finish check of dead rooms")` then it returns nil. Shouldn't you just return this error, remove the else and assume you'll only get there if err == nil?

watcher/watcher_test.go
```diff
@@ -4613,8 +4589,8 @@ var _ = Describe("Watcher", func() {
 				Expect(err).ToNot(HaveOccurred())
 			}
 
-			testing.MockRedisLock(mockRedisClient, downScalingLockKey, config.GetInt("watcher.lockTimeoutMs"), true, nil)
-			testing.MockReturnRedisLock(mockRedisClient, downScalingLockKey, nil)
+			// testing.MockRedisLock(mockRedisClient, downScalingLockKey, config.GetInt("watcher.lockTimeoutMs"), true, nil)
```
ghostec: remove?

ghostec:
LGTM

PR #76, opened 2019-08-23
Use only downscalingLock in watcher operations

lftakakura:
- watcher operations using only downscaling lock
- ScaleUp does not use lock
- ScaleDown use downscalingLock
- Pod metrics saving on redis (for metrics triggers) does not use lock


Henrod:
Why isn't this merged?

PR #77, opened 2019-08-29
Forward forwarder metadata along with gameserver metadata

lftakakura:
forwarding metadata configured in forwarders.<plugin>.<forwarder_name>. Gameserver metadata still takes precedence

### DIFF
--- a/plugins/grpc/forwarder.go
+++ b/plugins/grpc/forwarder.go
@@ -32,6 +32,7 @@ type GRPCForwarder struct {
 	client        pb.GRPCForwarderClient
 	logger        log.FieldLogger
 	serverAddress string
+	metadata      map[string]interface{}
 }
 
 // ForwarderFunc is the type of functions in GRPCForwarder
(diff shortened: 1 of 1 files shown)


plugins/grpc/forwarder.go
```diff
@@ -345,20 +345,32 @@ func (g *GRPCForwarder) SchedulerEvent(ctx context.Context, infos, fwdMetadata m
 
 //Forward send room or player status to specified server
 func (g *GRPCForwarder) Forward(ctx context.Context, event string, infos, fwdMetadata map[string]interface{}) (status int32, message string, err error) {
+
+	// Add forwarder metadata (from maestro config) to request
+	// Client metadata should take priority
+	metadata := g.config.GetStringMap("metadata")
```
ghostec: can't we store this value as a field in GRPCForwarder instead of retrieving from viper.Config every time?

plugins/grpc/forwarder.go
```diff
@@ -345,20 +345,32 @@ func (g *GRPCForwarder) SchedulerEvent(ctx context.Context, infos, fwdMetadata m
...
+		metadata = fwdMetadata
+	}
+
 	l := g.logger.WithFields(log.Fields{
 		"op":          "Forward",
 		"source":      "plugin/grpc",
 		"event":       event,
 		"infos":       fmt.Sprintf("%v", infos),
-		"fwdMetadata": fmt.Sprintf("%v", fwdMetadata),
+		"fwdMetadata": fmt.Sprintf("%v", metadata),
```
ghostec: logging this for every event isn't excessive? why is this info instead of debug?

plugins/grpc/forwarder.go
lftakakura: up

plugins/grpc/forwarder.go
lftakakura: up

PR #78, opened 2019-08-29
Remove rooms that are stuck in terminating state in RemoveDeadRooms

lftakakura:
This prevents leakage of roomStatus terminating keys in redis. This leakage affects the status of maestro schedulers (it sees more rooms terminating than it really has).

### DIFF
--- a/models/constants.go
+++ b/models/constants.go
@@ -46,6 +46,9 @@ const SegmentHMSet = "Redis/HMSet"
 //SegmentZRangeBy represents a segment
 const SegmentZRangeBy = "Redis/ZRangeBy"
 
+//SegmentSMembers represents a segment
+const SegmentSMembers = "Redis/SMembers"
+
 //SegmentSIsMember represents a segment
 const SegmentSIsMember = "Redis/SIsMember"
 
--- a/watcher/watcher.go
+++ b/watcher/watcher.go
@@ -606,27 +606,46 @@ func (w *Watcher) forwardRemovalRoomEvent(logger *logrus.Entry, rooms []string) 
 	return nil
 }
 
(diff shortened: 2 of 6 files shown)


watcher/watcher.go
```diff
@@ -606,27 +606,46 @@ func (w *Watcher) forwardRemovalRoomEvent(logger *logrus.Entry, rooms []string)
...
+	for _, pod := range pods {
+		liveKubePods[pod.GetName()] = true
+	}
+	for _, room := range rooms {
+		if _, ok := liveKubePods[room]; !ok {
+			zombieRooms = append(zombieRooms, models.NewRoom(room, w.SchedulerName))
+		}
+	}
+
+	models.ClearAllMultipleRooms(w.RedisClient.Client, w.MetricsReporter, zombieRooms)
```
ghostec: why do you ignore the possible error here?

lftakakura: I shouldn't. My bad

models/constants.go
```diff
@@ -46,6 +46,9 @@ const SegmentHMSet = "Redis/HMSet"
 //SegmentZRangeBy represents a segment
 const SegmentZRangeBy = "Redis/ZRangeBy"
 
+//SegmentSMembers represents a segment
```
andrehp: Why do you need this constant?

lftakakura: For tracing. But I wasn't using though. I updated the PR, and now I am:
```
err := mr.WithSegment(SegmentSMembers, func() error {
		var err error
		redisKeys, err = redisClient.SMembers(
			GetRoomStatusSetRedisKey(schedulerName, StatusTerminating),
		).Result()
		return err
	})
```

ghostec:
LGTM

PR #79, opened 2019-09-10
Feature/rolling update on watcher only

lftakakura:
- Now during scheduler update, pod creation and deletion is done via EnsureCorrectRooms (watcher).


api/scheduler_operation_handler.go
```diff
@@ -81,25 +81,32 @@ func getOperationStatus(
...
 
-	if status == nil || len(status) == 0 {
-		return empty, "scheduler with operation key not found",
-			NewOperationNotFoundError()
+	k := kubernetes.TryWithContext(app.KubernetesClient, ctx)
+	totalPods, err := k.CoreV1().Pods(schedulerName).List(
+		metav1.ListOptions{},
+	)
+	if err != nil {
+		return empty, "error getting getting pods from kubernetes", err
```
andrehp: getting getting

api/scheduler_operation_handler.go
```diff
@@ -81,25 +81,32 @@ func getOperationStatus(
...
 	}
 
-	if _, ok := status["status"]; ok {
-		return status, "", nil
+	status, err := operationManager.GetOperationStatus(*scheduler, totalPods.Items)
+	if err != nil {
+		return status, "error getting getting progress", err
```
andrehp: getting getting

controller/controller.go
```diff
@@ -240,7 +243,32 @@ func DeleteScheduler(
 		)
 	}
 	namespace := models.NewNamespace(scheduler.Name)
-	return deleteSchedulerHelper(logger, mr, db, redisClient, clientset, scheduler, namespace, timeoutSec)
+
+	// lock so autoscaler doesn't recreates rooms deleted
```
andrehp: s/recreates/recreate

controller/controller.go
```diff
@@ -240,7 +243,32 @@ func DeleteScheduler(
...
+	terminationLockKey := models.GetSchedulerTerminationLockKey(config.GetString("watcher.lockKey"), schedulerName)
+	configLock, _, err := AcquireLock(
+		context.Background(),
```
andrehp: Use ctx here?

controller/controller.go
```diff
@@ -732,6 +748,8 @@ func UpdateSchedulerConfig(
 				oldVersion,
 			)
 
+			// TODO: Send metrics of this error!!!
+			// this is the most critical part. If DBRollback fails, the watcher will keep spawning w
```
andrehp: spawning what?

watcher/watcher.go
```diff
@@ -1218,18 +1281,29 @@ func (w *Watcher) podsOfIncorrectVersion(
...
 
-		schedulerMajorVersion, _, err := w.splitedVersion(scheduler.Version)
+	for _, pod := range pods.Items {
+		podMajorVersion, _, err := w.splitedVersion(pod.Labels["version"])
 		if err != nil {
 			return nil, err
 		}
 
-		if podMajorVersion != schedulerMajorVersion {
+		// if a rollback happened consider 2 versions as correct:
```
cscatolini: This can be confusing. Is this really necessary? Couldn't we use only the actual version(after DB rollback)?

lftakakura: I agree it is confusing, but if I only consider the actual version it will replace all pods (even the ones that are in the correct version) if a rollback occur. If I consider the version that was rolled back to, it will only restart pods that need to

watcher/watcher.go
```diff
@@ -606,27 +578,46 @@ func (w *Watcher) forwardRemovalRoomEvent(logger *logrus.Entry, rooms []string)
...
 		if podNameMap[pod.GetName()] {
 			podsToReplace = append(podsToReplace, pod)
 		}
 	}
 
-	return podsToReplace, nil
+	return podsToReplace
+}
+
+// zombie rooms are the ones that are in terminating state but the pods doesn't exist
```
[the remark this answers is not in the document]
lftakakura: Yes, every pod that does not exist but has a state in redis should be considered zombie. I did it for terminating only cause it was the one causing problems. Will consider the others as well in the next commit

models/operation_manager.go
```diff
@@ -278,9 +285,60 @@ func (o *OperationManager) IsStopped() bool {
...
+	status, err := o.Get(o.operationKey)
+	if err != nil {
+		return nil, err
+	}
+
+	if status == nil || len(status) == 0 {
+		return nil, nil
+	}
+
+	if _, ok := status["status"]; ok {
```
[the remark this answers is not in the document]
lftakakura: Status set means the operation has already finished. Progress is set in `func (o *OperationManager) Finish(status int, description string, opErr error) error` method in this case

controller/controller.go
lftakakura: it was _spawning wrong gameserver_ but I will remove the comments as it does not makes sense anymore (I am sending metrics regarding this error and if it errors the watcher will perform a DBRollback)

models/operation_manager.go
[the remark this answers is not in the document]
cscatolini: okay, maybe add a small comment with this info?

watcher/watcher.go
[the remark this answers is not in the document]
cscatolini: okay! will this commit be in the same PR?

watcher/watcher.go
cscatolini: I can see the tradeoff here. In terms of code simplicity I'd go with replacing all pods, but I can see that this could cause unnecessary load and instability for the game. So I am okay with the approach you proposed.

watcher/watcher.go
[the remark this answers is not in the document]
lftakakura: yes. Will commit soon

PR #81, opened 2019-09-16
Copy metadata to avoid race condition in forwarding

lftakakura:


andrehp:
In what scenario did this race condition happen?

plugins/grpc/forwarder.go
```diff
@@ -349,7 +349,13 @@ func (g *GRPCForwarder) Forward(ctx context.Context, event string, infos, fwdMet
 
 	// Add forwarder metadata (from maestro config) to request
 	// Client metadata should take priority
-	metadata := g.metadata
+	metadata := map[string]interface{}{}
+	if g.metadata != nil {
+		for k, v := range g.metadata {
```
ghostec: do you see any problem in doing a shallow copy?

lftakakura:
> In what scenario did this race condition happen?

Before `metadata` was pointing to `g.metadata`. As `g.metadata` is global for the app, the service was crashing frequently as it was trying to read and write during forwardings

cscatolini:
> > In what scenario did this race condition happen?
> 
> `metadata` was pointing to `g.metadata`. As `g.metadata` is global for the app, the service was crashing frequently as it was trying to read and write during forwardings

It makes no sense for metadata to be an attribute of the struct 🤔 

This change LGTM.

PR #82, opened 2019-09-17
Remove event forward on scheduler operations (create and update).

cscatolini:
This PR removes events forwarding on scheduler operations. The reasoning behind this feature was to centralize all operations on the scheduler, including for external integrations.

However, in practice, this made the forwarder integrations more error prone because engineers do not know where they need to update the information.

This feature forwarded events for scheduler create and update operations. The data sent was the one present in the scheduler forwarders.metadata field.


andrehp:
I think it's a good idea to describe what this feature did to serve as a reference in the future. Which operations triggered the forwarding, for example

cscatolini:
> I think it's a good idea to describe what this feature did to serve as a reference in the future. Which operations triggered the forwarding, for example

I've added a few info. It was only triggered in scheduler create/update ops. There's not much else to add.

PR #83, opened 2019-09-18
Fix/ensure correct rooms in watch rooms

lftakakura:


controller/utils.go
```diff
@@ -1054,6 +1020,47 @@ func AcquireLock(
...
+
+	lock, err := redisClient.EnterCriticalSection(
+		redisClient.Trace(ctx),
+		lockKey,
+		time.Duration(lockTimeoutMS)*time.Millisecond,
+		0, 0,
+	)
+
+	if err != nil {
+		l.WithError(err).Error("error getting watcher lock")
```
cscatolini: Not necessarily watcher lock, right?

controller/utils.go
```diff
@@ -1054,6 +1020,47 @@ func AcquireLock(
...
+		return nil, err
+	}
+
+	if lock == nil {
+		l.Warnf("unable to get watcher %s lock %s, maybe some other process has it", schedulerName, lockKey)
```
cscatolini: same comment here

controller/utils.go
```diff
@@ -1054,6 +1020,47 @@ func AcquireLock(
...
+		return nil, fmt.Errorf("unable to get watcher %s lock %s, maybe some other process has it", schedulerName, lockKey)
```
cscatolini: ditto

controller/utils.go
```diff
@@ -1070,7 +1077,11 @@ func ReleaseLock(
 		err := redisClient.LeaveCriticalSection(lock)
 		if err != nil {
 			l.WithError(err).Error("error releasing lock. Either wait or remove it manually from redis")
+		} else {
+			l.Debug("lock released")
 		}
+	} else {
+		l.Warn("lock is nil. No lock to release")
```
cscatolini: should this be warn or debug?

watcher/watcher.go
```diff
@@ -447,12 +443,11 @@ func (w *Watcher) ReportRoomsStatuses() error {
 // WithLock is a helper function that runs a block of code
 // that needs to hold a type of lock to redis
 func (w *Watcher) WithLock(l *logrus.Entry, lockKey string, f func() error) (lockErr, err error) {
-	lock, _, err := controller.AcquireLock(
+	lock, err := controller.AcquireLockOnce(
```
cscatolini: What was the previous behaviour? What is this commit fixing by changing AcquireLock to AcquireLockOnce?

lftakakura: AcquireLock use ticker to retry acquiring lock. AcquireLockOnce does not retry.

As watchRooms already runs periodically with ticker, I think it is not necessary to have another ticker to acquire lock.

Also, workers were getting stuck trying to get lock until lockTimeout

cscatolini: Okay, makes sense!

PR #84, opened 2019-09-18
Send all forwarder metadata when forwarding an event.

cscatolini:
This PR changes the grpc forwarder mergeInfos method to send all forwarder metadata instead of whitelisting `roomInfo` key.

For example, if your forwarder is configured as follows:

```
"forwarders": {
    "grpc": {
      "matchmaking": {
        "enabled": true,
        "metadata": {
          "ping": true,
          "metadata": {
            "my": "data"
          },
          "roomType": "test"
        }
      }
    }
 ```

This will be necessarily forwarded in all status changes and ping requests:
```
{
   "ping": true,
    "metadata": {
      "my": "data"
    },
    "roomType": "test"
}
```

### DIFF
--- a/plugins/grpc/forwarder.go
+++ b/plugins/grpc/forwarder.go
@@ -327,30 +327,35 @@ func (g *GRPCForwarder) configure() error {
 }
 
 func (g *GRPCForwarder) mergeInfos(infos, fwdMetadata map[string]interface{}) map[string]interface{} {
-	if fwdMetadata != nil {
-		if roomType, ok := fwdMetadata["roomType"]; ok {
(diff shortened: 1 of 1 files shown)


andrehp:
LGTM, but I wasn't able to understand what was the expected previous behavior. Was there any kind of documentation?

cscatolini:
> LGTM, but I wasn't able to understand what was the expected previous behavior. Was there any kind of documentation?

No unfortunately (I blame past me!). Apparently we were only forwarding roomType because it is necessary on the other end (matchmaker service) to properly route the request. Since we have no tests for the plugin this change may break things, I asked @lftakakura to validate this in his test cluster before we merge.

lftakakura:
LGTM

PR #86, opened 2019-09-24
Refactor/rolling update progress

lftakakura:
** New strategy for tracking rolling update progress **

- When Watcher's EnsureCorrecRooms starts rolling updating, it set opManager to rolling update status and saves the amount of rooms to be replaced in a redis string along with a set with the roomIDs to be replaced
- Every time a room is deleted, the watcher remove a roomID from the redis set (`scheduler:%schedulerName:invalidRooms`)
- Instead of polling kube api to check if the rolling update has finished, the API polls redis and gives the progress percentage as `1 - scheduler:%schedulerName:invalidRooms / scheduler:%schedulerName:invalidRooms:count` where `scheduler:%schedulerName:invalidRooms:count` is the redis key holding a count of rooms to be replaced

### DIFF
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -682,24 +682,15 @@ func UpdateSchedulerConfig(
 	}
 
 	if shouldRecreatePods {
-		var kubePods *v1.PodList
 		var status map[string]string
 
-		operationManager.SetDescription(models.OpManagerRollingUpdate)
-
 		// wait for watcher.EnsureCorrectRooms to rolling update the pods
-		for {
-			// get list of actual pods
-			kubePods, err = ListCurrentPods(mr, clientset, schedulerName)
+		for err == nil {
+			status, err = operationManager.GetOperationStatus(mr, *scheduler)
 			if err != nil {
-				scheduler.RollingUpdateStatus = erroredStatus(err.Error())
-				break
-			}
-
-			status, err = operationManager.GetOperationStatus(*scheduler, kubePods.Items)
-			if err != nil {
-				scheduler.RollingUpdateStatus = erroredStatus(err.Error())
-				break
+				logger.WithError(err).Warn("errored trying to get rolling update progress")
+				err = nil
+				continue
 			}
 
 			// operation canceled
@@ -731,8 +722,10 @@ func UpdateSchedulerConfig(
 			time.Sleep(time.Second * 1)
 		}
 
-		if err != nil {
+		// delete invalidRooms key as EnsureCorrectRooms finished
+		models.RemoveInvalidRoomsKey(redisClient.Client, mr, schedulerName)
 
+		if err != nil {
 			l.WithError(err).Error("error during UpdateSchedulerConfig. Rolling back database")
 			dbRollbackErr := DBRollback(
 				ctx,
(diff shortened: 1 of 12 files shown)


models/room.go
```diff
@@ -603,3 +603,92 @@ func GetRoomsByMetric(redisClient interfaces.RedisClient, schedulerName string,
...
+// GetInvalidRoomsCountKey gets the key for the string that keeps count of invalid rooms
+func GetInvalidRoomsCountKey(schedulerName string) string {
+	return fmt.Sprintf("scheduler:%s:invalidRooms:count", schedulerName)
+}
+
+// SetInvalidRooms save a room in invalid redis set
+// A room is considered invalid if its version is not the scheduler current version
+func SetInvalidRooms(redisClient interfaces.RedisClient, mr *MixedMetricsReporter, schedulerName string, roomIDs []string) error {
+	pipe := redisClient.TxPipeline()
+	pipe.Del(GetInvalidRoomsKey(schedulerName))
```
cscatolini: Okay, so you're making sure the key is renovated every time you set the invalid rooms?

models/room.go
```diff
@@ -603,3 +603,92 @@ func GetRoomsByMetric(redisClient interfaces.RedisClient, schedulerName string,
...
+	pipe.Del(GetInvalidRoomsCountKey(schedulerName))
+	pipe.SAdd(GetInvalidRoomsKey(schedulerName), roomIDs)
+	pipe.Set(GetInvalidRoomsCountKey(schedulerName), len(roomIDs), 2*time.Hour)
+	err := mr.WithSegment(SegmentPipeExec, func() error {
```
cscatolini: you can just `return mr.WithSegment...`

models/room.go
```diff
@@ -603,3 +603,92 @@ func GetRoomsByMetric(redisClient interfaces.RedisClient, schedulerName string,
...
+	})
+	return err
+}
+
+// GetCurrentInvalidRoomsCount returns the count of current invalid rooms
+func GetCurrentInvalidRoomsCount(redisClient interfaces.RedisClient, mr *MixedMetricsReporter, schedulerName string) (int, error) {
+	count := 0
+	err := mr.WithSegment(SegmentSMembers, func() error {
+		var err error
+		keys, err := redisClient.SMembers(GetInvalidRoomsKey(schedulerName)).Result()
```
Henrod: It's better to use a lua script that returns the count than loading all members in memory and counting

controller/controller.go
```diff
@@ -1019,18 +1010,10 @@ func ScaleScheduler(
...
-		err := mr.WithSegment(models.SegmentPod, func() error {
-			var err error
-			pods, err = clientset.CoreV1().Pods(schedulerName).List(metav1.ListOptions{
-				LabelSelector: labels.Set{}.AsSelector().String(),
-				FieldSelector: fields.Everything().String(),
-			})
-			return err
-		})
+		// get list of actual pods
+		pods, err := ListCurrentPods(mr, clientset, schedulerName)
```
cscatolini: Why do you still need to list pods here?

models/room.go
lftakakura: Yes

models/room.go
cscatolini: actually, using SCARD (https://redis.io/commands/scard) would be the best choice

controller/controller.go
lftakakura: I do. This is in `ScaleScheduler`. It is triggered by a manual scale request from the API

lftakakura: It is inside the function that scales manually through API.

The API allows you to scale up/down and to a specific number of replicas. This pod listing is in the third case, as we need to know how many pods are running so we calculate the amount of pods to scale up or down to get the desired number

lftakakura: I reckon this feature is not much used

cscatolini: Okay. It makes sense to keep it like this.

PR #91, opened 2020-06-19
Fix maestro tests

victor-carvalho:
This branch fixes all tests from branch https://github.com/topfreegames/maestro/tree/refactor/rolling-update-progress


api/room_handler_test.go
```diff
@@ -315,8 +328,9 @@ forwarders:
 					"metadata": map[string]interface{}{
 						"ipv6Label": "",
 						"region":    "us",
+						"ports":     "[]",
 					},
-				}, gomock.Any())
+				}, gomock.Any()).AnyTimes()
```
lftakakura, line 333: It is not ideal to use AnyTimes. If possible assert that HGet is called the exact number of times it should

victor-carvalho: Yeah, it's not needed here. It's only called one time

controller/controller_test.go
```diff
@@ -1361,23 +1367,25 @@ cmd:
 	})
 
 	Describe("ScaleUp", func() {
+		BeforeEach(func() {
+			config.Set("watcher.maxScaleUpAmount", 100)
```
gabrielcorado, line 1371: Don't we need to unset this value after?

victor-carvalho: Sure, I will change this. I thought config was being reset every test but it was because I was running only one test separately.

PR #93, opened 2020-07-09
 Always set status to RollingUpdate if operation exists

victor-carvalho:
When trying to do operation that requires to recreate the pods (change image, env vars, etc ...) and there are no pods the operation gets stuck until it timeouts. This happens because on `GetOperationStatus` it checks if the operation is in `RollingUpdate` status and if not the progress always returns zero and `EnsureCorrectRooms` do not set the status if the number of invalid pods is zero.

This PR makes `EnsureCorrectRooms` always set the status to `RollingUpdate` if there is a current operation even if there are zero pods to update.


gabrielcorado:
Can this change mess with some Operation that has finished?

leohahn:
Is it hard to add a regression test for this issue?

PR #94, opened 2020-07-17
Worker should stop when operation is canceled

victor-carvalho:
When an operation is canceled the worker should know and stop the operation and the api should issue a rollback. Right now this is not happening because `OperationManager.WasCanceled` checks a cached state and never hits redis and the worker never starts the loop to check for operation updates.

I changed to `OperationManager` to remove the loop and make it so `WasCanceled` always check on redis. This works for most use cases inside maestro. To prevent having multiple goroutines hitting redis at the same time I moved the loop to inside the worker and separated from `OperationManager`.

I also added tests to `SegmentAndReplacePods` that were missing and added a test case to the watcher for when an operation is canceled.

### DIFF
--- a/controller/utils.go
+++ b/controller/utils.go
@@ -14,6 +14,7 @@ import (
 	"math"
 	"math/rand"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/sirupsen/logrus"
@@ -39,7 +40,6 @@ import (
 // SegmentAndReplacePods acts when a scheduler rolling update is needed.
 // It segment the list of current pods in chunks of size maxSurge and replace them with new ones
 func SegmentAndReplacePods(
-	ctx context.Context,
 	logger logrus.FieldLogger,
 	roomManager models.RoomManager,
 	mr *models.MixedMetricsReporter,
@@ -86,13 +86,13 @@ func SegmentAndReplacePods(
 
 		if timedout {
 			timeoutErr = errors.New("timedout waiting rooms to be replaced, rolled back")
-			l.WithError(timeoutErr).Error("error replacing chunk of pods")
+			l.WithError(timeoutErr).Error("operation timed out while replacing chunk of pods")
 			break
 		}
 
 		if canceled {
 			cancelErr = errors.New("operation was canceled, rolled back")
-			l.WithError(cancelErr).Error("error replacing chunk of pods")
+			l.WithError(cancelErr).Error("operation canceled while error replacing chunk of pods")
 			break
 		}
 
@@ -122,12 +122,11 @@ func replacePodsAndWait(
 	goroutinePoolSize int,
 ) (timedout, canceled bool, err error) {
 	logger.Debug("starting to replace pods with new ones")
-	stop := make(chan struct{}, 1)
-	defer close(stop) // Dont leak goroutines
-	finishedReplace := make(chan struct{})
 
-	timedoutChan := make(chan bool)
-	canceledChan := make(chan bool)
+	ctx, cancel := context.WithCancel(context.Background())
+
+	finishedReplace := make(chan struct{})
+	canceledChan := make(chan struct{})
 	errChan := make(chan error)
 
 	pods := make(chan *models.Pod, len(podsChunk))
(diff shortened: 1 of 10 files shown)


controller/utils.go
```diff
@@ -122,104 +122,119 @@ func replacePodsAndWait(
...
+	finishedReplace := make(chan struct{})
+	canceledChan := make(chan struct{})
 	errChan := make(chan error)
 
 	pods := make(chan *models.Pod, len(podsChunk))
 	for _, pod := range podsChunk {
 		pods <- pod
 	}
 
+	inRollingUpdate := operationManager != nil
```
leohahn, line 137: Are there cases where this function will be called without a rolling update?

controller/utils.go
```diff
@@ -122,104 +122,119 @@ func replacePodsAndWait(
...
+				finishedReplace,
+				errChan,
+			)
+		}()
+	}
+	
+	if inRollingUpdate {
+		go func() {
+			for {
+				canceled, err := operationManager.WasCanceled()
```
leohahn, line 166: If this a busy loop? Adding a sleep on the goroutine is then maybe a good idea.

leohahn: When does this goroutine stop?

controller/utils.go
```diff
@@ -122,104 +122,119 @@ func replacePodsAndWait(
...
 	mr *models.MixedMetricsReporter,
 	clientset kubernetes.Interface,
 	db pginterfaces.DB,
 	redisClient redisinterfaces.RedisClient,
-	willTimeoutAt time.Time,
 	configYAML *models.ConfigYAML,
 	scheduler *models.Scheduler,
-	operationManager *models.OperationManager,
-	clock clockinterfaces.Clock,
 	pods <-chan *models.Pod,
```
leohahn, line 210: On the function above pods is an array, however in this function is a channel, is this valid in go?

controller/utils.go
```diff
@@ -1144,9 +1129,12 @@ func AcquireLock(
 			l.Warn("timeout while wating for redis lock")
 			return nil, false, errors.New("timeout while wating for redis lock")
 		case <-ticker.C:
-			if operationManager != nil && operationManager.WasCanceled() {
-				l.Warn("operation was canceled")
-				return nil, true, nil
+			if operationManager != nil {
```
leohahn, line 1132: shouldn't you use inRollingUpdate here as well?

models/operation_manager.go
```diff
@@ -108,55 +98,20 @@ func (o *OperationManager) Start(
...
 		"operationKey": o.operationKey,
 	})
 
-	if !o.continueLoop {
-		return true, nil
-	}
-
 	l.Debug("checking if operation was canceled")
 	r, err := o.redisClient.HGetAll(o.operationKey).Result()
 	l.Debug("successfully accessed redis")
```
leohahn, line 117: This log seems incorrect, maybe we need to log this after checking the `err` variable.

controller/utils.go
victor-carvalho: Yes, when it's a rollback. There's no operation on a rollback.

controller/utils.go
victor-carvalho: Oh yes, I was going to put a ticker there but I forgot. I also have to use the context here too.

controller/utils.go
victor-carvalho: AcquireLock is used in other places as well and I didn't want to change every caller of this function.

models/operation_manager.go
victor-carvalho: Now that you say it does. I didn't change much in this function so I never looked that much into it but will change it.

victor-carvalho:
> Since we're moving the timeout logic to `SegmentAndReplacePods`, why not also do all the cancellation stuff there?
> 
> I think the sub-functions could only just receive a single context, and when it cancels it just stop doing its work, they don't really need to know the reason. And `SegmentAndReplacePods` could return the reason why it stopped. What do you think?

I think it makes sense. Will make the change.

controller/utils.go
leohahn: oh ok, didn't see that.

controller/utils.go
```diff
@@ -61,165 +60,182 @@ func SegmentAndReplacePods(
...
+	}
+
+	ticker := time.NewTicker(10 * time.Second)
+	defer ticker.Stop()
+	for {
+		select {
+		case <-ticker.C:
+			canceled, err := operationManager.WasCanceled()
+			if err != nil {
+				continue
```
gabrielcorado, line 127: In this case where `WasCanceled` returns an error (it can only be a Redis problem). We have two choices here:
* This problem could be resolved at the next tick but if not we'll keep this running until the timeout is reached (not sure if any other operation would have some problem using redis also, but I think we can't assume that);
* Cancel the operation since we actually don't know if Redis has the key or not.

My concern here is that in both cases we can have some "keys leak". What do you guys think?

Not sure if we what to solve this in this PR but, I would recommend adding a `warn` log here so we are aware that this Redis failures are happening here.

leohahn: A log helps, we can also add a maximum amount of times, and then it fails with an error maybe?

leohahn: Failing will not help here I guess as well.

leohahn: I think that having a failure helps, it would be a way of us taking control and fixing things manually maybe, without waiting for the timeout which is really long. Maybe adding a metric here so that we can have an alarm. A log is probably enough however.

controller/utils.go
```diff
@@ -61,165 +60,182 @@ func SegmentAndReplacePods(
...
 }
 
+func watchOperation(ctx context.Context, cancel func(), operationManager *models.OperationManager) {
+	canceled, err := operationManager.WasCanceled()
+	if canceled && err  == nil {
+		cancel()
+		return
+	}
+
+	ticker := time.NewTicker(10 * time.Second)
```
leohahn, line 120: Isn't 10 seconds too long? For the user the cancel operation will have a 10 second delay._

leohahn: Maybe putting this value in a config also helps.

controller/utils.go
victor-carvalho: After talking with corado on slack I think just logging would be enough. If this error happened once only in this place, it would be ok on the next try and if it a generalized redis problem the operation would fail creating or deleting pods and would cancel anyway.

controller/utils.go
victor-carvalho: I just used the same value that was used before on the loop inside OperationManager but I can read this from a config.

PR #95, opened 2020-07-21
Add support for will.iam in maestro

victor-carvalho:
This PR adds support for using [Will.IAM](https://github.com/topfreegames/Will.IAM) for authorization in maestro handlers.


api/william_middleware.go
```diff
@@ -0,0 +1,70 @@
...
+
+	if !hasPermission {
+		m.App.HandleError(w, http.StatusForbidden, "forbidden", nil)
+		return
+	}
+	m.next.ServeHTTP(w, r)
+}
+
+//SetNext handler
+func (m *WilliamMiddleware) SetNext(next http.Handler) {
```
leohahn, line 68: Is there a reason why you don't receive the next parameter on the constructor?

api/william_resolver.go
```diff
@@ -0,0 +1,75 @@
...
+
+type PermissionResolver interface {
+	ResolvePermission(app *App, req *http.Request) (permission, resource string, err error)
+}
+
+type actionResolver struct {
+	action string
+}
+
+func (r actionResolver) ResolvePermission(app *App, req *http.Request) (permission, resource string, err error) {
```
leohahn, line 24: Shouldn't the method receiver be a pointer?

api/william_resolver.go
```diff
@@ -0,0 +1,75 @@
...
+	return r.action, "", nil
+}
+
+type schedulerPathResolver struct {
+	action  string
+	varName string
+}
+
+func (r schedulerPathResolver) ResolvePermission(app *App, req *http.Request) (permission, resource string, err error) {
```
leohahn, line 33: The same comment about the pointer receiver here.

api/william_middleware.go
victor-carvalho: That's the way middlewares work on maestro, you call `Chain` on them and maestro calls `SetNext` to form a linked list of middlewares so you need to create them without a next.

api/william_resolver.go
victor-carvalho: Yeah, will change that. Just forgot to put it there 🤦

api/auth/constants.go
```diff
@@ -0,0 +1,10 @@
+package auth
+
+type AuthenticationResult int
```
gabrielcorado, line 3: Since we have only one success and the other are errors, isn't the `error` returned by the functions enough? In that case `error = nil` would mean a `AuthenticationOk`.

api/auth/william.go
```diff
@@ -0,0 +1,96 @@
...
+	"github.com/sirupsen/logrus"
+	"github.com/topfreegames/extensions/pg/interfaces"
+	"github.com/topfreegames/maestro/errors"
+	"github.com/topfreegames/maestro/models"
+	"github.com/topfreegames/maestro/william"
+	"net/http"
+	"strings"
+)
+
+type PermissionResolver interface {
```
gabrielcorado, line 15: Shouldn't this interface and the "default" implementation live at `api/auth/auth.go` or something since it is no specific to `william`?

api/app.go
```diff
@@ -426,6 +436,7 @@ func (a *App) configureApp(
 	a.configureCache()
 	a.configureSentry()
 	a.configureLogin()
+	a.configureWilliam()
```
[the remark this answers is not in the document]
victor-carvalho: Right now it only instantiates the object when it's configured but I can make a check.

api/access_middleware.go
```diff
@@ -7,149 +7,104 @@
...
 
 //ServeHTTP methods
 func (m *AccessMiddleware) ServeHTTP(w http.ResponseWriter, r *http.Request) {
-	if !m.enabled {
-		m.next.ServeHTTP(w, r)
-		return
-	}
+	ctx := r.Context()
+	logger := middleware.GetLogger(ctx)
+	if auth.BasicAuthEnabled(m.App.Config) {
```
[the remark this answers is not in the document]
victor-carvalho: I think it makes sense. My reasoning here was just trying to avoid any behaviour changes but I agree that the three should be consistent.

api/auth/basic_test.go
```diff
@@ -0,0 +1,62 @@
...
+var _ = Describe("Basic Auth", func() {
+	Describe("CheckBasicAuth", func() {
+		It("should return AuthenticationOk and email when user and pass are valid and x-forwarded-user-email is sent", func() {
+			request, err := http.NewRequest("GET", "/scheduler", nil)
+			Expect(err).ToNot(HaveOccurred())
+
+			request.Header.Add("x-forwarded-user-email", "user@email.com")
+			request.SetBasicAuth(config.GetString("basicauth.username"), config.GetString("basicauth.password"))
+
+			result, email := CheckBasicAuth(config, request)
```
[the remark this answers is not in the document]
victor-carvalho: Yeah, most of the project use this approach. But I can make changes to set the keys on a `BeforeEach` block.

api/auth/constants.go
victor-carvalho: The only reason for this is to differentiate between an error and an invalid token. So I can return InternalServerError or Unauthorized. Thinking more I could probably use two types of error for this too., what do you think?

api/auth/william.go
victor-carvalho: The concept of different permissions are only relevant to william.

api/access_middleware.go
```diff
@@ -7,149 +7,104 @@
...
+			return
+		}
 
-	if status == http.StatusBadRequest {
-		logger.WithError(err).Error("error validating access token")
-		err := errors.NewAccessError("Unauthorized access token", fmt.Errorf(msg))
-		m.App.HandleError(w, http.StatusUnauthorized, "Unauthorized access token", err)
-		return
-	}
+		logger.Debug("authenticated with william")
```
leohahn, line 84: Isnt this log a bit misleading? You're not checking with William if the token is valid or not.

victor-carvalho: Yeah, you are right.

PR #96, opened 2020-07-23
Fix redis cleanup bug.

leohahn:
Redis would not be cleaned up if a pod failed to be created on
kubernetes.

Also, fix some issues with linting and formatting of the code.

### DIFF
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -395,7 +395,6 @@ func ScaleUp(
 		return err
 	}
 	if existPendingPods {
-		// TODO(lhahn): Important.
 		return errors.New("there are pending pods, check if there are enough CPU and memory to allocate new rooms")
 	}
 
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -395,6 +395,7 @@ func ScaleUp(
 		return err
 	}
 	if existPendingPods {
+		// TODO(lhahn): Important.
 		return errors.New("there are pending pods, check if there are enough CPU and memory to allocate new rooms")
 	}
 
--- a/controller/controller_test.go
+++ b/controller/controller_test.go
@@ -1241,7 +1241,7 @@ cmd:
 
 			for _, roomName := range expectedRooms {
 				mt.MockPodNotFound(mockRedisClient, configYaml.Name, roomName)
-				pod, err := models.NewPod(roomName, nil, configYaml, clientset, mockRedisClient, mr)
+				pod, err := models.NewPod(roomName, nil, configYaml, mockRedisClient, mr)
 				Expect(err).NotTo(HaveOccurred())
 				_, err = pod.Create(clientset)
 				Expect(err).NotTo(HaveOccurred())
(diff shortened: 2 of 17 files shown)


controller/controller.go
```diff
@@ -395,6 +395,7 @@ func ScaleUp(
 		return err
 	}
 	if existPendingPods {
+		// TODO(lhahn): Important.
```
leeohaddad, line 398: Will you leave this TODO here?

If yes: does it make sense to describe more specifically what should be done here?

controller/controller_test.go
```diff
@@ -1565,7 +1564,7 @@ cmd:
 			Expect(err).NotTo(HaveOccurred())
 			scheduler := models.NewScheduler(configYaml1.Name, configYaml1.Game, yamlWithLimit)
 
-			mt.MockScaleUp(mockPipeline, mockRedisClient, configYaml1.Name, configYaml1.AutoScaling.Max - currentRooms)
```
leeohaddad, line 1568: Was this change suggested by linter?
I usually like to keep empty spaces in calculations like that to make it easier to visualize.

models/game_room.go
```diff
@@ -180,9 +181,12 @@ func createPod(
 			containers[i].Env = append(containers[i].Env, namesEnvVars...)
 		}
 
-		pod, err = NewPodWithContainers(name, containers, configYAML, clientset, redisClient, mr)
+		pod, err = NewPodWithContainers(name, containers, configYAML, redisClient, mr)
 	}
+
 	if err != nil {
+		// NOTE: We will get an error here if we cannot access redis. This implies that redis is probably not accessible,
+		// se we were not successful in adding the room's keys as well. That is why we don't call room.ClearAll here.
```
leeohaddad, line 189: I believe there is something missing or mistyped in the beginning of the second line of this note

models/pod.go
```diff
@@ -296,16 +292,17 @@ func getContainerWithName(name string, pod *Pod) v1.Container {
...
+// configureHostPorts will fail if it cannot access redis.
 func (p *Pod) configureHostPorts(
 	configYaml *ConfigYAML,
-	clientset kubernetes.Interface,
 	redisClient redisinterfaces.RedisClient,
 	mr *MixedMetricsReporter,
 ) error {
 	pod, err := GetPodFromRedis(redisClient, mr, p.Name, p.Namespace)
 	if err != nil {
+		// TODO, FIXME(lhahn): this error is misleading.
```
leeohaddad, line 303: Is there a specific reason for not fixing it now and just leave the TODO instead?

leeohaddad:
I noticed you actually added tests for the cleanup. Do you know why is coveralls complaining about decreased coverage?

controller/controller.go
leohahn: I'll remove them.

controller/controller_test.go
leohahn: this is gofmt.

PR #98, opened 2020-07-31
Cancel operation after an error

victor-carvalho:
After the refactoring to fix the cancellation of an operation there was a bug in SegmentAndReplacePods where the operation could be stuck for a long time after finished (with an error for example) waiting for the goroutines to complete.

This PRs solves a lot of problems of goroutine leaks in this flow and also the new bug by using a child context.

### DIFF
--- a/controller/controller_test.go
+++ b/controller/controller_test.go
@@ -6583,7 +6583,63 @@ containers:
 			Expect(err).ToNot(HaveOccurred())
 		})
 
-		It("should return error when it fails to create pod", func() {
+		It("should return error when it fails to create pod with goroutinePoolSize == 1", func() {
+			pods := []*models.Pod{
+				&models.Pod{Name: "room-1"},
+				&models.Pod{Name: "room-2"},
+			}
+
+			scheduler := models.NewScheduler(configYaml1.Name, configYaml1.Game, string(configYaml1.ToYAML()))
+
+			mockRedisClient.EXPECT().
+				HGetAll(opManager.GetOperationKey()).
+				Return(goredis.NewStringStringMapResult(map[string]string{
+					"description": models.OpManagerRollingUpdate,
+				}, nil)).AnyTimes()
+
+			mockRedisClient.EXPECT().
+				TxPipeline().
+				Return(mockPipeline)
+
+			mockPipeline.EXPECT().
+				HMSet(gomock.Any(), gomock.Any()).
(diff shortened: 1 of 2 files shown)


controller/utils.go
```diff
@@ -498,7 +506,18 @@ func waitCreatingPods(
 
 	for {
 		exit := true
+
+		// go select statement chooses "randomly" when two branches are available, sot it can make select ignore
```
DylanGuedes, line 510: nit: "so it"

controller/utils.go
```diff
@@ -189,23 +191,24 @@ func replacePodsAndWait(
...
+	wg.Wait()
+
 	select {
 	case err = <-errChan:
-		logger.Error("operation terminated with error")
-	case <-ctx.Done():
-		logger.Debug("operation timedout/canceled")
-	case <-finishedReplace:
-		logger.Debug("all pods were successfully replaced")
+		logger.WithError(err).Debug("error replacing pods")
```
leohahn, line 207: Shouldn't this be at least a warning?

victor-carvalho: The error is returned up the stack so the caller function can log the error, logging here just for debug

controller/utils.go
```diff
@@ -498,7 +506,18 @@ func waitCreatingPods(
...
+
+		// go select statement chooses "randomly" when two branches are available, so it can make select ignore
+		// ctx.Done() for iterations. This check prevents to enter select if the context was already cancelled.
+		if ctx.Err() != nil {
+			logger.Warn("operation canceled/timeout waiting for rooms to be created")
+			return true, nil
+		}
+
 		select {
+		case <-ctx.Done():
```
gabrielcorado, line 518: Since we're checking if it has been cancelled, is this still required?

controller/controller_test.go
```diff
@@ -6586,6 +6586,7 @@ containers:
 		It("should return error when it fails to create pod", func() {
```
gabrielcorado, line 6586: Do you think it is worth to test both scenarios?
1. When only one worker has to create both pods;
2. When two or more workers take pods to create (what this current version tests);

victor-carvalho: I think it makes sense. I will add another test.

controller/utils.go
victor-carvalho: Just to prevent blocking some time waiting for next loop if canceled.

PR #99, opened 2020-08-04
Validate waiting state when creating pods.

leohahn:
This PR makes some changes:
- It will more reliably check for incorrect images (CrashLoop backoff and ImagePull backoff) when creating new pods and return errors in these cases.
- If creating the pods fails, it will also cleanup Redis and delete the bogus pod as well.


watcher/watcher.go
```diff
@@ -1314,6 +1317,17 @@ func (w *Watcher) EnsureCorrectRooms() error {
...
+	// We need to make sure that at least one goroutine is going to be used for the update of the pods,
+	// otherwise the operation will timeout.
+	goroutinePoolSize := w.Config.GetInt("watcher.goroutinePoolSize")
+	if goroutinePoolSize == 0 {
+		logger.Warn("watcher.goroutinePoolSize was 0, assigning it to 1")
+		goroutinePoolSize = 1
+	}
+
+	// TODO, FIXME(lhahn): remove this.
+	timeoutSec = 120
```
gabrielcorado, line 1329: Any reason why this value is being forced here? Also, I'm not sure if it does have any effect since `timeDur` is already set.

leohahn: Oh, this should be removed, I'll change that.

controller/utils.go
```diff
@@ -289,9 +289,37 @@ func createNewRemoveOldPod(
...
+			redisClient,
+			configYAML,
+			// The function only uses the name of the pod.
+			[]*models.Pod{{Name: newPod.Name}},
+		)
+		if deleteErr != nil {
+			return false, fmt.Errorf("failed to delete bogus pod: %s", deleteErr)
+		}
+		if canceled {
+			return true, nil
```
leeohaddad, line 313: Why not returning `err` anymore here?

leeohaddad: Same question about line 322

models/pod.go
```diff
@@ -417,20 +418,21 @@ func IsPodTerminating(pod *v1.Pod) bool {
...
-func ValidatePodWaitingState(pod *Pod) error {
+// - ErrImagePull
+var InvalidPodWaitingStateErr = goerrors.New("invalid pod waiting state")
 
+func ValidatePodWaitingState(pod *Pod) error {
 	for _, invalidState := range InvalidPodWaitingStates {
 		status := &pod.Status
 		if checkWaitingReason(status, invalidState) {
-			return fmt.Errorf("one or more containers in pod are in %s", invalidState)
+			return fmt.Errorf("%s: one or more containers in pod are in %s", InvalidPodWaitingStateErr, invalidState)
```
leeohaddad, line 433: Maybe you wanted to use %w instead of %s here. Or maybe add .Error() to InvalidPodWaitingStateErr.

watcher/watcher.go
```diff
@@ -280,20 +279,34 @@ func (w *Watcher) reportRoomsStatusesRoutine() {
 }
 
 func (w *Watcher) watchRooms() error {
-	l := w.Logger.WithFields(logrus.Fields{
-		"operation": "watcher.watchRooms.EnsureCorrectRooms",
-	})
-	_, _ = w.WithDownscalingLock(l, w.EnsureCorrectRooms)
+	{
```
leeohaddad, line 282: Curiosity: What is the reason for using this scopes instead of just reusing `l`?

models/pod.go
leeohaddad: Maybe you wanted to use %w instead of %s here. Or maybe add .Error() to InvalidPodWaitingStateErr and replace fmt.Errorf by errors.New + fmt.Sprintf.

watcher/watcher.go
```diff
@@ -803,14 +807,12 @@ func (w *Watcher) AutoScale() error {
 			w.Run = false
 			return err
 		}
-		logger.WithError(err).Error("failed to get scheduler scaling info")
-		return err
+		return fmt.Errorf("failed to get scheduler scaling info: %s", err)
```
leeohaddad, line 810: Maybe you wanted to use %w instead of %s here. Or maybe add .Error() to InvalidPodWaitingStateErr and replace fmt.Errorf by errors.New + fmt.Sprintf.

watcher/watcher.go
```diff
@@ -803,14 +807,12 @@ func (w *Watcher) AutoScale() error {
...
 	}
 
 	err = w.updateOccupiedTimeout(scheduler)
 	if err != nil {
-		logger.WithError(err).Error("failed to update scheduler occupied timeout")
-		return err
+		return fmt.Errorf("failed to update scheduler occupied timeout: %s", err)
```
leeohaddad, line 815: Maybe you wanted to use %w instead of %s here. Or maybe add .Error() to InvalidPodWaitingStateErr and replace fmt.Errorf by errors.New + fmt.Sprintf.

models/pod.go
leeohaddad, line 433: Maybe you wanted to use %w instead of %s here. Or maybe add .Error() to InvalidPodWaitingStateErr & replace fmt.Errorf by errors.New + fmt.Sprintf.

Same comment applies to other Errorf's added in this MR, specially in watcher.go, but I will concentrate it here to avoid repetitive comments.

watcher/watcher.go
```diff
@@ -746,26 +751,26 @@ func (w *Watcher) RemoveDeadRooms() error {
...
 		}
 
-		if timeoutErr == false && err == nil {
-			l := logger.WithFields(logrus.Fields{
+		if len(roomsNoPingSince) > 0 {
+			logger.WithFields(logrus.Fields{
 				"rooms": fmt.Sprintf("%v", roomsNoPingSince),
-			})
-			l.Info("successfully deleted rooms that were not pinging")
+			}).Info("successfully deleted rooms that were not pinging")
```
leeohaddad, line 764: Shouldn't this become a warning, like the log for zombie rooms? Same question applies for rooms that expired occupied timeout.

models/pod.go
leohahn: The current version of go that maestro is being built does not support this feature of using %w.

PR #101, opened 2020-08-11
Add a flag to scheduler forwarders configuration to not return its response

gabrielcorado:
Currently API calls that forward events expect the forward to return a response and error, those values are used to return the API. For example if a forward fails the API call also fails.

This new configuration enable the schedulers to choose if they want to receive forwarders response (and also errors).

In order to not break the current behaviour this flag is set to `true` keeping the current behaviour.

### DIFF
--- a/eventforwarder/forward.go
+++ b/eventforwarder/forward.go
@@ -25,8 +25,9 @@ type Response struct {
 // Forwarder is the struct that defines a forwarder with its EventForwarder as configured
 // in maestro and the metadata configured for each scheduler
 type Forwarder struct {
-	Func     EventForwarder
-	Metadata map[string]interface{}
+	Func           EventForwarder
+	ReturnResponse bool
+	Metadata       map[string]interface{}
 }
 
 func getEnabledForwarders(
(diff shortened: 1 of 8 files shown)


leohahn:
We should have documentation for this feature.

eventforwarder/forward.go
```diff
@@ -137,7 +140,8 @@ func ForwardRoomEvent(
 				}
 				eventWasForwarded = true
 
-				return ForwardEventToForwarders(ctx, enabledForwarders, status, infos, l)
+				res, warning, err = ForwardEventToForwarders(ctx, enabledForwarders, status, infos, l)
```
leohahn, line 143: is this warning variable used somewhere?

models/config_yaml_test.go
```diff
@@ -7,6 +7,44 @@ import (
 )
 
 var _ = Describe("ConfigYaml", func() {
+	Describe("NewCofigYaml", func() {
+		It("should define default value for forwarder", func() {
+			configYaml, _ := NewConfigYAML(`name: scheduler-name
+game: game
```
leohahn, line 13: I've seen that we basically copy the yaml for most test cases in maestro. Maybe having a parameterized function that returns a new yaml config would be less verbose (this comment is more for future commits, not necessarily you have to change this MR).

eventforwarder/forward.go
gabrielcorado: Yes, it is passed to the `reportRPCStatus`.

leohahn:
LGTM

PR #102, opened 2020-08-18
Disable panic scale by default

victor-carvalho:
After some incidents we could see that panic scales is not helping most of the time and sometimes it's actually making things worse so this PR puts this behaviour behind a flag that will be disabled by default but can be enabled if someone really wants it.


leohahn:
LGTM

leohahn:
Thanks for updating the docs!

PR #103, opened 2020-09-03
Use kubernetes informer instead of watcher

victor-carvalho:
Using informers gives us access to many tools in client-go and also takes care of many things that we need to do like worrying about reconnections.

### DIFF
--- a/controller/utils.go
+++ b/controller/utils.go
@@ -256,7 +256,9 @@ func replacePodWorker(
 
 			logger.Infof("pods remaining to replace: %d", len(pods))
 
-			return nil
+			if len(pods) == 0 {
+				return nil
+			}
 		case <-ctx.Done():
 			return nil
 		}
--- a/controller/utils.go
+++ b/controller/utils.go
@@ -255,6 +255,8 @@ func replacePodWorker(
 			}
 
 			logger.Infof("pods remaining to replace: %d", len(pods))
+
+			return nil
 		case <-ctx.Done():
 			return nil
 		}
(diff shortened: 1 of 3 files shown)


controller/utils.go
```diff
@@ -255,6 +255,8 @@ func replacePodWorker(
 			}
 
 			logger.Infof("pods remaining to replace: %d", len(pods))
+
+			return nil
```
leohahn, line 259: is the for loop necessary if you're returning nil here?

victor-carvalho: Sorry, forgot the if.

watcher/watcher.go
```diff
@@ -1513,94 +1520,128 @@ func (w *Watcher) checkIfUsageIsAboveLimit(
...
+	if !ok {
+		// Object is not a string so we skip, shouldn't happen
+		w.Queue.Forget(obj)
+		return
+	}
+
+	err := w.podEventHandler(key)
+	if err != nil {
+		// An error happened so we requeue it so it can processed again later
+		w.Queue.AddRateLimited(key)
```
leohahn, line 1595: Do we want to reprocess in all error cases?

victor-carvalho: Even though I treat some different errors in handler code the only errors that can really happen are redis errors and I think we always want to retry in this cases or redis state can become too different from Kubernetes and that's really bad for maestro.

leohahn:
Does the informer take care of reconnections, etc?

watcher/watcher.go
leohahn: Cool, makes sense.

victor-carvalho:
> Does the informer take care of reconnections, etc?

Yes.

gabrielcorado (approved):

PR #105, opened 2020-09-22
Fix RemoveDeadRooms delete metric reason

victor-carvalho:
When RemoveDeadRooms deletes a room it is reported with a reason as "update". Now RemoveDeadRooms will report the reason as "ping_timeout" when it deletes pods.


leohahn:
LGTM

gabrielcorado (approved):

PR #109, opened 2020-09-25
Move pod state count report to different loop

victor-carvalho:
Because the routine to report pod state count is in the same loop as the watcher's main operations (EnsureCorrectRooms/RemoveDeadRooms/AutoScale) it can not be reported for some time while these operations are running and these are the times the metric is needed most.

The routine was moved to another goroutine running alongside the report of GRU's state.

### DIFF
--- a/watcher/watcher.go
+++ b/watcher/watcher.go
@@ -155,7 +155,7 @@ func (w *Watcher) loadConfigurationDefaults() {
 	w.Config.SetDefault("watcher.autoScalingPeriod", 10)
 	w.Config.SetDefault("watcher.roomsStatusesReportPeriod", 10)
 	w.Config.SetDefault("watcher.ensureCorrectRoomsPeriod", 10*time.Minute)
-	w.Config.SetDefault("watcher.podStatesCountPeriod", 10*time.Second)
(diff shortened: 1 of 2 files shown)


watcher/watcher.go
```diff
@@ -155,7 +155,7 @@ func (w *Watcher) loadConfigurationDefaults() {
 	w.Config.SetDefault("watcher.autoScalingPeriod", 10)
 	w.Config.SetDefault("watcher.roomsStatusesReportPeriod", 10)
 	w.Config.SetDefault("watcher.ensureCorrectRoomsPeriod", 10*time.Minute)
-	w.Config.SetDefault("watcher.podStatesCountPeriod", 1*time.Minute)
+	w.Config.SetDefault("watcher.podStatesCountPeriod", 10*time.Second)
```
gabrielcorado, line 158: Since this operation actually lists all pods form `GetPodMapFromRedis` (which uses an `HGETALL`) and parse them, could we put too much pressure into the Redis? My concern is that because of this constantly running other places where we use the same list pods would have some problem, WDYT? What about keeping this interval larger (for example `30s` or more)?

gabrielcorado (approved):

PR #114, opened 2021-03-24
Change Jaeger configuration to new standard

rsafonseca:
Changes jaeger configuration to env variables for better configurability without future code builds to modify settings
Updates opentracing and jaeger-client libs


victor-carvalho (approved):
LGTM

PR #119, opened 2021-04-09
Use HScan instead of HGetAll to list pods

victor-carvalho:


models/pod.go
```diff
@@ -474,34 +475,47 @@ func GetPodMapRedisKey(schedulerName string) string {
...
+	// The RedisClient interface does not have HSCAN and this is the only way to extract the *redis.Client inside it
+	pageSize := config.GetInt64("podListPageSize")
+	redisKey := GetPodMapRedisKey(schedulerName)
+	podMap = make(map[string]*Pod)
+
+	for cursor := uint64(0); ; {
+		var results []string
+		err := mr.WithSegment(SegmentHScan, func() error {
+			pipeline := redisClient.TxPipeline()
+			cmd := pipeline.HScan(redisKey, 0, "*", pageSize)
```
gabrielcorado, line 492: Shouldn't the second argument be the `counter`?

PR #122, opened 2021-04-19
EnsureCorrectRooms replace only invalid versions

gabrielcorado:
This PR changes how the `EnsureCorrectRooms` (process inside the Watcher). The previous behavior was: list pods that are not sending status information and pods with incorrect versions.

This PR removes that part of replacing pods that are not sending status, and now it will only replace pods with invalid versions. The invalid pods will be removed by the RemoveDeadRooms process and "replaced" by the AutoScale process.

### DIFF
--- a/watcher/watcher.go
+++ b/watcher/watcher.go
@@ -945,6 +945,7 @@ func (w *Watcher) checkState(
 	if scheduler.State == models.StateCreating || scheduler.State == models.StateTerminating {
 		return scaling, nil
 	}
+
 	// Rooms below Min
 	if roomCount.Available() < autoScalingInfo.Min {
 		scaling.Delta = autoScalingInfo.Min - roomCount.Available()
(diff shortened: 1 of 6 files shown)


victor-carvalho (approved):

watcher/watcher.go
```diff
@@ -1247,32 +1223,20 @@ func (w *Watcher) EnsureCorrectRooms() error {
...
-		}
-	}
-
-	// get unregistered pod names
-	var unregisteredPodNames []string
-	if len(unregisteredPods) > 0 {
-		for _, pod := range unregisteredPods {
-			unregisteredPodNames = append(unregisteredPodNames, pod.Name)
-		}
+	if len(incorrectPods) <= 0 {
```
mathnogueira, line 1233: `len()` doesn't return negative numbers. You can replace it by:

```suggestion
	if len(incorrectPods) == 0 {
```

PR #125, opened 2021-05-24
Create interface for access to room redis status

victor-carvalho:


internal/entities/scheduler.go
```diff
@@ -0,0 +1,5 @@
+package entities
+
+type Scheduler struct {
+	Name string
```
gabrielcorado, line 4: Can we change this attribute to `ID`? It might fit GameRoom better.

gabrielcorado (approved):

PR #135, opened 2021-06-07
RoomManager delete room

gabrielcorado:
Adds the function that will remove rooms. It also introduces the SetRoomStatus, which is responsible for transitioning the status for the game rooms.


internal/core/services/room_manager/status.go
```diff
@@ -0,0 +1,60 @@
...
+)
+
+// validStatusTransitions this map has all possible status changes for a game
+// room.
+var validStatusTransitions = map[game_room.GameRoomStatus]map[game_room.GameRoomStatus]struct{}{
+	game_room.GameStatusPending: {
+		game_room.GameStatusReady:   struct{}{},
+		game_room.GameStatusUnready: struct{}{},
+		game_room.GameStatusError:   struct{}{},
+	},
```
victor-carvalho, line 17: When `DeleteRoom` is called in RoomManager it sets the status as terminating. Could we not get a case where it tries to set the status from pending to terminating?

gabrielcorado: I think the best here is to permit any state to transition to terminating since we could have some "force-delete" scenarios. I'll update it.

victor-carvalho (approved):

PR #150, opened 2021-07-16
Add worker manager responsible to start operation workers and keep them in sync

luizmiranda7:
This PR aims to create a service called workerManager that controls all job executions of Maestro NEXT.

Also, in order to bootstrap the first job, we're creating the operationWorker that will be responsible the consume the pending queue of scheduler operations.


internal/core/services/workers_manager/workers_manager.go
```diff
@@ -0,0 +1,159 @@
...
+
+	ticker := time.NewTicker(time.Duration(w.syncOperationWorkersInterval) * time.Second)
+	defer ticker.Stop()
+
+	for w.runSyncOperationWorkers == true {
+		select {
+		case <-ticker.C:
+			err := w.SyncOperationWorkers(ctx)
+			if err != nil {
+				zap.L().Error("scheduled sync operation workers failed", zap.Error(err))
```
gabrielcorado, line 75: Do you think we can recover if there is an error here? If not, wouldn't it be better if we stop everything?

internal/core/services/workers_manager/workers_manager.go
```diff
@@ -0,0 +1,159 @@
...
+		syncOperationWorkersInterval: configs.GetInt(syncOperationWorkersIntervalPath),
+	}
+}
+
+// Function to run a first sync and start a periodically sync worker
+func (w *WorkersManager) Start(ctx context.Context) error {
+
+	err := w.SyncOperationWorkers(ctx)
+	if err != nil {
+		zap.L().Error("initial sync operation workers failed", zap.Error(err))
```
gabrielcorado, line 51: If we have failed to sync the first time, should we keep the process going?

internal/core/services/workers_manager/workers_manager.go
```diff
@@ -0,0 +1,159 @@
...
+	return nil
+}
+
+// Function to start a infinite loop (ticker) that will call
+// (periodically) the SyncOperationWorkers func
+func (w *WorkersManager) startSyncOperationWorkers(ctx context.Context) error {
+
+	w.runSyncOperationWorkers = true
+	sigchan := make(chan os.Signal)
+	signal.Notify(sigchan, os.Interrupt, syscall.SIGINT, syscall.SIGTERM)
```
gabrielcorado, line 65: What about moving this to cmd, where we know things about the environment etc.?

internal/core/services/workers_manager/workers_manager_test.go
```diff
@@ -0,0 +1,228 @@
...
+		}, nil)
+
+		workersManager := NewWorkersManager(configs, schedulerStorage, *operationManager)
+
+		err := workersManager.Start(context.Background())
+		require.NoError(t, err)
+
+		require.Contains(t, workersManager.CurrentWorkers, "zooba-us")
+
+		assertLogMessages(t, recorded, map[zapcore.Level][]string{
```
gabrielcorado, line 63: I'm not sure about making assertions on the logs, it seems a bit fragile.

internal/core/workers/operation_worker.go
```diff
@@ -0,0 +1,86 @@
...
+	signal.Notify(sigchan, os.Interrupt, syscall.SIGINT, syscall.SIGTERM)
+
+	ticker := time.NewTicker(time.Duration(w.syncPeriod) * time.Second)
+	defer ticker.Stop()
+
+	for w.run == true {
+		select {
+		case <-ticker.C:
+			zap.L().Info("Running operation worker", zap.String("scheduler_name", w.scheduler.Name))
+			w.countRuns++
```
gabrielcorado, line 61: Can we remove this since it is being used only on tests?

internal/core/services/workers_manager/workers_manager.go
luizmiranda7: Returning the error

internal/core/services/workers_manager/workers_manager.go
luizmiranda7: Making use of context.Done() channel

internal/core/workers/operation_worker.go
luizmiranda7: Removed!

internal/core/services/workers_manager/workers_manager_test.go
luizmiranda7: Once we have no metrics already defined, I'm keeping the logs assertions.

As soon as we have metrics in place, we can remove it.

gabrielcorado (approved):

PR #155, opened 2021-07-20
Add metrics basic package

luizmiranda7:
This PR aims to add a static package with metrics creation and population functions.

We're assuming that maestro has only Prometheus metrics and will open a HTTP server to serve metrics over a specific route.

Also, this PR covers the metric creation and manipulation with integration tests.


internal/core/monitoring/metrics.go
```diff
@@ -0,0 +1,95 @@
...
+	Name       string
+	Help       string
+	Labels     []string
+	Buckets    []float64
+	Objectives map[float64]float64
+}
+
+// runMetricsServer start a metrics server in other goroutine, and returns a
+// shutdown function.
+func RunMetricsServer(ctx context.Context, configs config.Config) func() error {
```
gabrielcorado, line 29: Not sure about the metrics server initialization being here. It seems like this is the responsibility of the `cmd` or its supportive packages (like `internal/services`). WDYT?

internal/core/monitoring/metrics_test.go
```diff
@@ -0,0 +1,131 @@
...
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+	"github.com/stretchr/testify/require"
+)
+
+func TestCounterCreation(t *testing.T) {
```
gabrielcorado, line 15: Here we're testing that our metrics are being registered, right? It feels like we're testing the Prometheus library. Could we use some process it offers to us? Like gathering metrics and checking if ours is there, we would avoid implementing a metrics server here and could move this test to `unit` since it does integrate only with an external library.

Also, I just want to keep in mind that these tests rely on some global/default states, so I'm not sure if those tests guarantee anything. For example, there is nothing preventing anyone from registering their metric with the same name as here, which would cause this test to fail because of external interference. (not sure how we could improve this tho, we could keep it this way and see if this is something we'll need to tackle later).

internal/core/monitoring/metrics.go
```diff
@@ -0,0 +1,95 @@
...
+)
+
+type MetricOpts struct {
+	Namespace  string
+	Subsystem  string
+	Name       string
+	Help       string
+	Labels     []string
+	Buckets    []float64
+	Objectives map[float64]float64
```
gabrielcorado, line 24: Do you have anything in mind for this attribute? Right now, it is not being used.

internal/core/monitoring/metrics.go
```diff
@@ -0,0 +1,95 @@
...
+}
+
+func CreateLatencyMetric(options *MetricOpts) *prometheus.HistogramVec {
+	return promauto.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Namespace: options.Namespace,
+			Subsystem: options.Subsystem,
+			Name:      options.Name + "_latency",
+			Help:      options.Help + " (latency)",
+			Buckets:   options.Buckets,
```
gabrielcorado, line 71: Do you think it is valuable if we provide a default value for those buckets? We could even remove it from the options and define that all the Latency metrics follow the same buckets. In this way, we could set those values to fit better a specific time unit (like millisecond).

internal/core/monitoring/metrics.go
```diff
@@ -0,0 +1,95 @@
...
+			Namespace: options.Namespace,
+			Subsystem: options.Subsystem,
+			Name:      options.Name + "_gauge",
+			Help:      options.Help + " (gauge)",
+		},
+		options.Labels,
+	)
+}
+
+func ReportLatencyMetric(metric *prometheus.HistogramVec, start time.Time, labels ...string) {
```
gabrielcorado, line 89: Do you think it is valuable to add the time unit to the function name? Something like `ReportLatencyMetricMilliseconds`.

internal/core/monitoring/metrics.go
luizmiranda7: Removing

internal/core/monitoring/metrics.go
luizmiranda7: Moved to internal/services/metrics.go

internal/core/monitoring/metrics.go
luizmiranda7: Yeap, sure!

internal/core/monitoring/metrics_test.go
luizmiranda7: I like the idea to transform it as unit. Still learning how to test it without calling it's API.

gabrielcorado (approved):

PR #157, opened 2021-07-21
Add operation execution worker metrics

luizmiranda7:
This PR aims to add metrics of latency and count for the operation execution worker.

There are 4 main metrics:
* `operation_execution`: Latency metric to measure how long a operation execution is taking;
    * Contains 3 metadatas: Scheduler, Operation, Success
* `operation_on_error`: Latency metric to measure how long a operation error fallback is taking;
    * Contains 3 metadatas: Scheduler, Operation, Success
* `operation_evicted`: Count metric to measure how many operations are being evicted;
    * Contains 3 metadatas: Scheduler, Operation, Reason
* `operation_execution_worker_failed`: Count metric to measure how many operations workers are failing;
    * Contains 2 metadatas: Scheduler, Reason

### DIFF
--- a/internal/core/workers/operation_execution_worker/metrics.go
+++ b/internal/core/workers/operation_execution_worker/metrics.go
@@ -9,9 +9,9 @@ import (
 
 const (
 	LabelNoOperationExecutorFound = "no_operation_executor_found"
-	LabelShouldNotExecute = "should_not_execute"
-	LabelNextOperationFailed = "next_operation_failed"
-	LabelStartOperationFailed = "start_operation_failed"
+	LabelShouldNotExecute         = "should_not_execute"
+	LabelNextOperationFailed      = "next_operation_failed"
+	LabelStartOperationFailed     = "start_operation_failed"
 )
 
 var (
@@ -20,7 +20,6 @@ var (
 		Subsystem: monitoring.SubsystemWorker,
 		Name:      "operation_execution",
 		Help:      "An scheduler operation was executed",
-		Buckets:   monitoring.DefBucketsMs,
 		Labels: []string{
 			monitoring.LabelScheduler,
 			monitoring.LabelOperation,
@@ -34,7 +33,6 @@ var (
 		Subsystem: monitoring.SubsystemWorker,
 		Name:      "operation_on_error",
 		Help:      "An scheduler operation on error fallback was executed",
-		Buckets:   monitoring.DefBucketsMs,
 		Labels: []string{
 			monitoring.LabelScheduler,
 			monitoring.LabelOperation,
@@ -48,7 +46,6 @@ var (
 		Subsystem: monitoring.SubsystemWorker,
 		Name:      "operation_evicted",
 		Help:      "An scheduler operation was evicted",
-		Buckets:   monitoring.DefBucketsMs,
 		Labels: []string{
 			monitoring.LabelScheduler,
 			monitoring.LabelOperation,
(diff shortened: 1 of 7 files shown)


internal/core/workers/operation_execution_worker/metrics.go
```diff
@@ -0,0 +1,105 @@
...
+)
+
+var (
+	operationExecutionLatencyMetric           *prometheus.HistogramVec
+	operationOnErrorLatencyMetric             *prometheus.HistogramVec
+	operationEvictedCountMetric               *prometheus.CounterVec
+	operationExecutionWorkerFailedCountMetric *prometheus.CounterVec
+)
+
+func init() {
```
gabrielcorado, line 25: Why do we need to set the metrics on the package init?

internal/core/workers/operation_execution_worker/metrics.go
```diff
@@ -0,0 +1,105 @@
...
+	resetMetrics()
+}
+
+func resetMetrics() {
+	if operationExecutionLatencyMetric != nil {
+		prometheus.Unregister(operationExecutionLatencyMetric)
```
gabrielcorado, line 31: Not sure why do we need to do this. Using `promauto` must be enough to guarantee a unique definition.

internal/core/workers/operation_execution_worker/metrics.go
luizmiranda7: We could do it on the CMD as well, but once we don't have it, I believe this is the static right place to initialize them.

Also, in order to reset metrics (unit tests) we need to call the resetMetrics function.

internal/core/workers/operation_execution_worker/metrics.go
luizmiranda7: The intention here is not to guarantee the unique definition. But, to only unregister metrics if there's at least one registered.

gabrielcorado: Why do we need to do this? If we left them as package variables, they would be initialized only once.

internal/core/workers/operation_execution_worker/metrics.go
gabrielcorado: My question was more related to why not define them as package variables (since we don't need to check for their failure or perform any extra step). Is the only reason because we are asserting them on our unit tests? If yes, it might be related to [a point I raised on another PR](https://github.com/topfreegames/maestro/pull/155#discussion_r673934833), where we could have trouble relying on "global state", in this case, I think we need to come up with a solution for it instead of adding extra logic to our package, WDYT?

luizmiranda7: Got it, that's the problem on setting metrics on global state and testing them on unit tests that does not clears the global state.

To me, we never came up before with this problem because no repo is testing metrics the way we're. Even prometheus testutils is not enough or does not give us a complete set of tools to test or even assert our metrics.

Totally agree we're relying on global state to fetch our metrics, still I cannot see an alternative given the premises:
* Keep metrics on "global state";
* Assert metrics on test;

The very use of promauto blocks the right assertion once it registers the metric directly to the default register/gather/collector. That lives on global scope.

What other alternatives would you suggest?

internal/core/workers/operation_execution_worker/metrics.go
luizmiranda7: There's no way to clear the metrics between tests, the only way to do it is to unregister and register a new metric in place. That's why we're doing it.

gabrielcorado (approved):

PR #159, opened 2021-07-22
Add worker manager metrics

luizmiranda7:
This PR aims to add metrics of gauge and count for the worker manager events.

There are 3 main metrics:
* `current_workers`: Gauge metric to measure how many workers are alive;
    * Contains 1 metadata: Scheduler
* `restarted_workers`: Counter metric to measure how many workers has been restarted;
    * Contains 1 metadata: Scheduler
* `workers_sync`: Count metric to measure how many workers sync has been executed;
    * Contains no metadata

### DIFF
--- a/internal/core/services/workers_manager/metrics.go
+++ b/internal/core/services/workers_manager/metrics.go
@@ -1,11 +1,22 @@
 package workers_manager
 
 import (
+	"github.com/prometheus/client_golang/prometheus"
 	"github.com/topfreegames/maestro/internal/core/monitoring"
 )
 
 var (
-	currentWorkersGaugeMetricOpts = monitoring.MetricOpts{
+	currentWorkersGaugeMetric     *prometheus.GaugeVec
+	restartedWorkersCounterMetric *prometheus.CounterVec
+	workersSyncCounterMetric      *prometheus.CounterVec
+)
+
+func init() {
+	initMetrics()
+}
+
+func initMetrics() {
+	currentWorkersGaugeMetric = monitoring.CreateGaugeMetric(&monitoring.MetricOpts{
 		Namespace: monitoring.Namespace,
 		Subsystem: monitoring.SubsystemWorker,
 		Name:      "current_workers",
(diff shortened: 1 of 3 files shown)


internal/core/services/workers_manager/metrics.go
```diff
@@ -0,0 +1,54 @@
+package workers_manager
+
+import (
+	"github.com/topfreegames/maestro/internal/core/monitoring"
+)
+
+var (
+	currentWorkersGaugeMetricOpts = monitoring.MetricOpts{
```
gabrielcorado, line 8: nit: Is there any specific reason for defining it as a separated variable instead of inside the `CreateGaugeMetric` call?

internal/core/workers/operation_execution_worker/operation_execution_worker.go
```diff
@@ -101,18 +108,21 @@ func (w *OperationExecutionWorker) Start(ctx context.Context) error {
...
 			loopLogger.Debug("operation execution failed", zap.Error(executionErr))
+
+			onErrorStartTime := time.Now()
 			onErrorErr := executor.OnError(operationContext, op, def, executionErr)
 			if onErrorErr != nil {
 				loopLogger.Error("operation OnError failed", zap.Error(onErrorErr))
 			}
+			ReportOperationOnErrorLatency(onErrorStartTime, w.schedulerName, op.DefinitionName, onErrorErr != nil)
 		}
+		ReportOperationExecutionLatency(executeStartTime, w.schedulerName, op.DefinitionName, executionErr != nil)
```
gabrielcorado, line 119: Shouldn't this report be called right after the `Execute` returns? Here, we're counting the `OnError` time, so in case of error, the reported time for the execution would be `ExecuteTime + OnErrorTime`.

internal/core/workers/operation_execution_worker/metrics.go
```diff
@@ -0,0 +1,90 @@
...
+		Help:      "An scheduler operation execution worker failed and is no longer running",
+		Labels: []string{
+			monitoring.LabelScheduler,
+			monitoring.LabelReason,
+		},
+	}
+	operationExecutionWorkerFailedCountMetric = monitoring.CreateCounterMetric(&operationExecutionWorkerFailedCountMetricOpts)
+)
+
+func ReportOperationExecutionLatency(start time.Time, schedulerName, operationName string, success bool) {
```
gabrielcorado, line 70: nit: We could keep those functions private since no other package will use them.

internal/core/services/workers_manager/metrics.go
luizmiranda7: Copy&Paste reasons rsrs!

gabrielcorado (approved):

PR #161, opened 2021-07-23
Add first version of the worker CMD

gabrielcorado:
We're getting started with the entry point of our workers. This PR does the following:
* Add `wire` as a dependency;
* Create the workers manager initializer using `wire`;
* Fix operations execution worker to follow the `workers.WorkerBuild`;
* Add missing Operation flow Redis initializer;
* Get started with minimum configuration;


cmd/worker/worker.go
```diff
@@ -0,0 +1,78 @@
...
+
+	shutdownMetricsServerFn := service.RunMetricsServer(ctx, config)
+
+	// TODO(gabrielcorado): support multiple workers.
+	workersManager, err := initializeWorker(config, operation_execution_worker.NewOperationExecutionWorker)
+	if err != nil {
+		zap.L().With(zap.Error(err)).Fatal("fail to initialize workers manager")
+	}
+
+	zap.L().Info("initialized workers manager, starting...")
```
luizmiranda7, line 51:
```suggestion
	zap.L().Info("workers manager initialized, starting...")
```

cmd/worker/worker.go
```diff
@@ -0,0 +1,78 @@
...
+	go workersManager.Start(ctx)
+
+	<-ctx.Done()
+	// TODO(gabrielcorado): call the workers manager stop.
```
luizmiranda7, line 55: Why are we not calling it? I mean, we know the current stop implementation has some limits, but still can't a good reason to not call it as is.

PR #162, opened 2021-07-27
Management API CMD

luizmiranda7:
This PR aims to create a new management API command to start the API with all required handlers.

- The initial handler serves a ping endpoint which will (further) serve the room integration.

The command starts 2 HTTP servers:
- Internal HTTP server: contains endpoints for health-check and metrics;
- Management HTTP server: contains the management API endpoints;


internal/service/management_api.go
```diff
@@ -0,0 +1,60 @@
...
+func ProvideManagementMux(ctx context.Context, pingHandler *handlers.PingHandler) *runtime.ServeMux {
+	mux := runtime.NewServeMux()
+	api.RegisterPingHandlerServer(ctx, mux, pingHandler)
+
+	return mux
+}
+
+// RunManagementServer starts HTTP server in other goroutine, and returns a
+// shutdown function. It serves management API endpoints/handlers.
+func RunManagementServer(ctx context.Context, configs config.Config, mux *runtime.ServeMux) func() error {
```
gabrielcorado, line 28: This function looks like it could be used to start other APIs. WDYT about making it more generic?

internal/service/management_api.go
```diff
@@ -0,0 +1,60 @@
...
+	"github.com/slok/go-http-metrics/middleware"
+	"github.com/slok/go-http-metrics/middleware/std"
+	"github.com/topfreegames/maestro/internal/config"
+	"github.com/topfreegames/maestro/internal/core/monitoring"
+	"github.com/topfreegames/maestro/internal/handlers"
+	api "github.com/topfreegames/maestro/protogen/api/v1"
+	"go.uber.org/zap"
+)
+
+func ProvideManagementMux(ctx context.Context, pingHandler *handlers.PingHandler) *runtime.ServeMux {
```
gabrielcorado, line 19: The `cmd/management_api` will be the only one using this function. Do you think we can move it there (as this `service` package will be shared between multiple commands)?

internal/service/management_api.go
```diff
@@ -0,0 +1,60 @@
...
+	mux := runtime.NewServeMux()
+	api.RegisterPingHandlerServer(ctx, mux, pingHandler)
+
+	return mux
+}
+
+// RunManagementServer starts HTTP server in other goroutine, and returns a
+// shutdown function. It serves management API endpoints/handlers.
+func RunManagementServer(ctx context.Context, configs config.Config, mux *runtime.ServeMux) func() error {
+	if !configs.GetBool("management_api.enabled") {
```
gabrielcorado, line 29: Not sure if we want this kind of option since the `management_api` is a command. If we run the `cmd/management_api` with this config as `false`, it won't do anything.

internal/service/internal_api.go
```diff
@@ -0,0 +1,58 @@
...
+	"github.com/topfreegames/maestro/internal/config"
+	"go.uber.org/zap"
+)
+
+// RunInternalServer start a internal server in other goroutine, and returns a
+// shutdown function. The internal server contains handlers for:
+// - health check
+// - metrics
+func RunInternalServer(ctx context.Context, configs config.Config) func() error {
+	if !configs.GetBool("internal_api.enabled") {
```
gabrielcorado, line 18: Do you think there is any case where this will be `false`?

internal/service/management_api.go
luizmiranda7: Actually it depends a lot on management_api configurations... don't think it could be so reusable as we would like.. also, once /internal/service is shared by all CMDs I'm tending to move this function to the CMD.

gabrielcorado: In that case, it makes sense to move it to the CMD.

Makefile
```diff
@@ -39,3 +39,19 @@ wire:
...
+.PHONY: run-management-api
+run-management-api:
+	@go run cmd/management_api/wire_gen.go cmd/management_api/management_api.go
+
+.PHONY: generate
+generate: buf/install
+	@go generate ./gen
+
+.PHONY: buf/install
+buf/install:
```
gabrielcorado, line 51: Can't we move those dependencies definitions to our `tools/` package? In this way, we'll also fix the version we're using.

buf.yaml
```diff
@@ -0,0 +1,8 @@
+
+version: v1beta1
+name: git.topfreegames.com/topfreegames/maestro
```
gabrielcorado, line 3: The correct one is `github.com/topfreegames/maestro`.

api/proto/v1/ping.proto
```diff
@@ -0,0 +1,23 @@
+syntax = "proto3";
+
+package api.v1;
+
+option java_package = "com.topfreegames.maestro.protogen.api.v1";
+option go_package = "git.topfreegames.com/topfreegames/maestro/protogen/api/v1";
```
gabrielcorado, line 6: The package name must be on our `github.com/topfreegames/maestro`.

gabrielcorado (approved):

PR #164, opened 2021-07-28
List schedulers route

luizmiranda7:
This PR aims to add a new HTTP endpoint/route: GET /schedulers, that will return all persisted schedulers.

Also, this PR introduces 2 new concepts:

- SchedulerManager: core service that holds all scheduler business logic;
- SchedulerHandler: HTTP API handler of all scheduler domain requests;
- Scheduler proto file: protobuf file to keep all scheduler API contracts


api/proto/v1/ping.proto
```diff
@@ -2,8 +2,8 @@ syntax = "proto3";
 
 package api.v1;
 
-option java_package = "com.topfreegames.maestro.protogen.api.v1";
-option go_package = "github.com/topfreegames/maestro/protogen/api/v1";
+option java_package = "com.topfreegames.maestro.pkg.api.v1";
+option go_package = "github.com/github.com/grpc-ecosystem/grpc-gateway/v2/runtime/api/v1";
```
gabrielcorado, line 6: Probably a typo here.

gabrielcorado (approved):

PR #167, opened 2021-08-03
WorkerManager.Stop waits until all workers are done 

luizmiranda7:
This PR aims to make `WorkerManager.Stop` to wait until all workers are done.

To do it, WorkerManager controls the workers execution with a instance scoped waiting group. When the Stop function is called, at last operation, the manager waits until all goroutines from the `WaitGroup` are done.


internal/core/workers/mock/mock.go
```diff
@@ -1,20 +1,25 @@
...
 
 type MockWorker struct {
-	Run bool
+	Run           bool
+	SleepDuration time.Duration
 }
 
 func (d *MockWorker) Start(_ context.Context) error {
 	d.Run = true
+	time.Sleep(d.SleepDuration)
```
gabrielcorado, line 15: WDYT about not relying on time to define when the start should return? In this way, we avoid having some "magic numbers".

We could add a `stopCh`, for example, that will block our start until the `Stop` is called (it will work very similar to what our workers are going to implement).

```go
type MockWorker struct {
	StopCh chan struct{}
}

func (d *MockWorker) Start(...) {
	d.Run = true
	<-d.StopCh
	return nil
}

func (d *MockWorker) Stop(...) {
	d.StopCh <- struct{}{}
	d.Run = false
}
```

luizmiranda7: Done!

gabrielcorado (approved):

PR #168, opened 2021-08-04
Fix create scheduler e2e local flow

luizmiranda7:
This PR aims to fix some major problems found when trying to run the NEXT branch locally in order to create a scheduler successfully. The major problems were:

* Operation registry was not being populated, then operation manager was not able to marshal/unmarshal operation definitions;
* Operation executors were not being inject into operation execution worker correctly, then no operation was being executed;
* Missing k3s configuration when running locally was breaking the k8s client;

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -58,7 +58,7 @@ services:
     - POSTGRES_PASSWORD=maestro
 
   redis:
-    image: redis:3.2-alpine
+    image: redis:5.0.10-alpine
     ports:
     - "6379:6379"
 
--- a/internal/core/operations/definition.go
+++ b/internal/core/operations/definition.go
@@ -7,10 +7,10 @@ import (
 )
 
 type DefinitionConstructor func() Definition
-type DefintionConstructors map[string]DefinitionConstructor
+type DefinitionConstructors map[string]DefinitionConstructor
 
-func NewDefinitionConstructors() DefintionConstructors {
-	return DefintionConstructors{}
+func NewDefinitionConstructors() DefinitionConstructors {
+	return DefinitionConstructors{}
 }
 
 // Definition is the operation parameters. It must be able to encode/decode
--- a/internal/core/operations/definition.go
+++ b/internal/core/operations/definition.go
@@ -6,6 +6,13 @@ import (
 	"github.com/topfreegames/maestro/internal/core/entities/operation"
 )
 
+type DefinitionConstructor func() Definition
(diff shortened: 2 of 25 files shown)


internal/core/operations/definition.go
```diff
@@ -6,6 +6,12 @@ import (
 	"github.com/topfreegames/maestro/internal/core/entities/operation"
 )
 
+type DefinitionConstructor func() Definition
+
+func NewDefinitionConstructors() map[string]DefinitionConstructor {
```
gabrielcorado, line 11: Do you think defining a file for those "DefinitionConstructors" would make it easier to read and enable easier changes?

We would initially create a type alias:
```go
type DefintionConstructors map[string]DefinitionConstructor

func NewDefinitionConstructors() DefinitionConstructors {
	return DefintionConstructors(map[string]DefintionConstructor)
}
```

And other places that rely on it would just reference it as `DefintionConstructors`.

docker-compose.yaml
```diff
@@ -58,7 +58,7 @@ services:
     - POSTGRES_PASSWORD=maestro
 
   redis:
-    image: redis:3.2-alpine
+    image: redis:5.0.10-alpine
```
gabrielcorado, line 61: Question: we're changing the version because our integration tests use the `5.0`, is that right?

gabrielcorado (commented):
Have you considered having only one `OperationProvider` which has both `Definition and `Executor`? It seems like having two places to configure error prune. Also, we could think about a way of not having to repeat the operation name. Since it is already present on the `Definition` we could reuse it to reduce duplication and typos.

docker-compose.yaml
luizmiranda7: Yeap, exactly what we're doing. Checking our production redis versions, we're using 6.x redis databases.

luizmiranda7: Yeap, exactly what we're doing. Checking our production redis versions, we're using 6.x redis databases.

internal/core/operations/definition.go
```diff
@@ -6,6 +6,13 @@ import (
 	"github.com/topfreegames/maestro/internal/core/entities/operation"
 )
 
+type DefinitionConstructor func() Definition
+type DefintionConstructors map[string]DefinitionConstructor
```
gabrielcorado, line 10:
```suggestion
type DefinitionConstructors map[string]DefinitionConstructor
```

gabrielcorado (approved):

PR #169, opened 2021-08-04
Create Scheduler route

luizmiranda7:
This PR aims to implement a HTTP/JSON route/endpoint responsible for create schedulers.

The route should be invoked locally by:

```
curl --location --request POST 'http://localhost:8080/schedulers' \
--header 'Content-Type: application/json' \
--data-raw '{
    "name": "zooba-blue",
    "game": "zooba",
    "version": "1.0.0"
}'
```

The request may result in the following things:

* New scheduler on database tables: `schedulers` and `scheduler_version`;
* New operation on redis buckets: `operations:*:*`;
* New operation pending entry on redis zset: `pending_operations:*`.


internal/api/handlers/scheduler_handler.go
```diff
@@ -44,3 +47,35 @@ func (h *SchedulerHandler) ListSchedulers(ctx context.Context, message *api.Empt
...
 
 }
+
+func (h *SchedulerHandler) CreateScheduler(ctx context.Context, request *api.CreateSchedulerRequest) (*api.Scheduler, error) {
+
+	scheduler := h.fromRequestToEntity(request)
+
+	scheduler, err := h.schedulerManager.CreateScheduler(ctx, scheduler)
+	if err != nil {
+		return nil, fmt.Errorf("failed create scheduler: %w", err)
```
gabrielcorado, line 57: Shouldn't we care about the errors we're returning here? otherwise, our external requestors will only receive an "unknown error", making it hard to understand. [gRPC has a list of status](https://grpc.github.io/grpc/core/md_doc_statuscodes.html) code that we can return, we would probably need to use the [status](https://pkg.go.dev/google.golang.org/grpc/status) and [codes](https://pkg.go.dev/google.golang.org/grpc/codes) package.

luizmiranda7: That nice, didn't notice we had such things to work with!

PR #170, opened 2021-08-05
List operations route

luizmiranda7:
This PR aims to implement a HTTP/Json route to fetch all operations of a scheduler.

The route can be invoked with the following curl:

```
$ curl --location --request GET 'http://localhost:8080/schedulers//operations'
```

Returns a map with the following keys:

* Pending operations;
* Active operations;
* Finished operations;


api/proto/v1/schedulers.proto
```diff
@@ -54,3 +61,20 @@ message PortRange {
...
+}
+
+message ListOperationsReply {
+  repeated Operation pending_operations = 1;
+  repeated Operation active_operations = 2;
+  repeated Operation finished_operations = 3;
+}
+
+message Operation {
+  string iD = 1;
```
gabrielcorado, line 76: nit: It seems that the ident is a little bit off here.

internal/api/handlers/scheduler_handler.go
```diff
@@ -79,3 +120,49 @@ func (h *SchedulerHandler) fromEntityToResponse(entity *entities.Scheduler) *api
...
+
+	return &api.Operation{
+		ID:             entity.ID,
+		Status:         status,
+		DefinitionName: entity.DefinitionName,
+		SchedulerName:  entity.SchedulerName,
+	}, nil
+}
+
+func fromOperationStatusToString(status operation.Status) (string, error) {
```
gabrielcorado, line 151: Do you think it makes sense to move this function to something like `.String` for the `operation.Status`? Like what we have on [`GameServerStatus`](https://github.com/topfreegames/maestro/blob/next/internal/core/entities/game_room/game_room.go#L25).

gabrielcorado (commented):
Side note: We need to start documenting our Proto messages and fields. We can add a `buf lint` to our lint step, but we can do it in another PR.

gabrielcorado (approved):

PR #174, opened 2021-08-11
Change remove dead rooms to be async

gabrielcorado:
Context: The RemoveDeadRooms flow is responsible for removing rooms that haven't send ping information or hit the occupied timeout. In some cases, deleting those "malfunctioning" rooms can take a long time, making the entire scheduler very unstable (having a low number of rooms or not being able to update the scheduler).

Since the RemoveDeadRooms don't affect the other parts of the flow (AutoScale and EnsureCorrectRooms), we're making it not block the `watchRooms` flow. In addition to making it async, we're creating a new lock specific for the `RemoveDeadRooms` flow to prevent multiple of them from running in parallel and enable the other parts of the flow to execute.

### DIFF
--- a/watcher/watcher.go
+++ b/watcher/watcher.go
@@ -288,20 +288,27 @@ func (w *Watcher) reportRoomsStatusesRoutine() {
 }
 
 func (w *Watcher) watchRooms() error {
-	l := w.Logger.WithFields(logrus.Fields{
-		"operation": "watcher.watchRooms.EnsureCorrectRooms",
-	})
-	_, _ = w.WithDownscalingLock(l, w.EnsureCorrectRooms)
+	_, _ = w.WithDownscalingLock(
+		w.Logger.WithFields(logrus.Fields{
+			"operation": "watcher.watchRooms.EnsureCorrectRooms",
+		}),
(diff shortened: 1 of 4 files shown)


watcher/watcher.go
```diff
@@ -288,20 +288,27 @@ func (w *Watcher) reportRoomsStatusesRoutine() {
...
+			"operation": "watcher.watchRooms.EnsureCorrectRooms",
+		}),
+		w.EnsureCorrectRooms,
+	)
 
-	l = w.Logger.WithFields(logrus.Fields{
-		"operation": "watcher.watchRooms.RemoveDeadRooms",
-	})
-	_, _ = w.WithDownscalingLock(l, w.RemoveDeadRooms)
+	go w.WithRemoveDeadRoomsLock(
```
luizmiranda7, line 298: WDYT of moving this goroutine to run as first instruction? Above of EnsureCorrectRooms.. makes sense?

luizmiranda7 (approved):

watcher/watcher.go
gabrielcorado: There will be differences only on the first execution. But we can change it.

PR #177, opened 2021-08-12
Add "preventRoomsCreationWithError" option to scheduler

gabrielcorado:
Context: In #120, we introduced a behavior where Maestro would stop scaling or updating if it finds any pod with an error. In practice, we've seen that this is not desired in some situations, like when having an incident where some rooms could be in an error state while the rest keeps working.

We're adding a new configuration option for the scheduler called `preventRoomsCreationWithError`, which has two possible values:
* `true` (which is the default value): Will keep the current behavior of stopping pods creation when there is any error;
* `false`: Will do like previous versions of Maestro, keep scaling and updating the scheduler even if there are pods with error;

### DIFF
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -395,13 +395,13 @@ func ScaleUp(
 		return fmt.Errorf("failed to load scheduler configuration: %w", err)
 	}
 
-	hasNotReadyPods, err := checkNotReadyPods(config, redisClient, scheduler.Name, configYAML.PreventRoomsCreationWithError, mr)
+	notReadyPods, err := hasNotReadyPods(config, redisClient, scheduler.Name, configYAML.PreventRoomsCreationWithError, mr)
 	if err != nil {
 		return fmt.Errorf("failed to list pending or failed pods: %w", err)
 	}
 
-	if hasNotReadyPods {
-		return errors.New("scheduler has pods not ready (pending or with error)")
+	if notReadyPods {
+		return errors.New("scheduler has not ready pods (pending or with error)")
 	}
 
 	amount, err = SetScalingAmount(
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -390,14 +390,18 @@ func ScaleUp(
 		"scaleUpLimit": scaleUpLimit,
 	})
 
-	configYAML, _ := models.NewConfigYAML(scheduler.YAML)
-
-	existPendingPods, err := pendingOrFailedPods(config, redisClient, scheduler.Name, mr)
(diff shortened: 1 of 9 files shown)


controller/controller.go
```diff
@@ -390,14 +390,18 @@ func ScaleUp(
...
 	})
 
-	configYAML, _ := models.NewConfigYAML(scheduler.YAML)
+	configYAML, err := models.NewConfigYAML(scheduler.YAML)
+	if err != nil {
+		return fmt.Errorf("failed to load scheduler configuration: %w", err)
+	}
 
-	existPendingPods, err := pendingOrFailedPods(config, redisClient, scheduler.Name, mr)
+	hasNotReadyPods, err := checkNotReadyPods(config, redisClient, scheduler.Name, configYAML.PreventRoomsCreationWithError, mr)
```
luizmiranda7, line 398: minor: When you say: check not ready pods, I read: "if true, everything is fine, else, there're not ready pods"

WDYT of change the name to hasNotReadyPods? In this way I believe things get more clear.

controller/controller.go
```diff
@@ -390,14 +390,18 @@ func ScaleUp(
...
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to list pending or failed pods: %w", err)
 	}
-	if existPendingPods {
-		return errors.New("there are pending or failed pods")
+
+	if hasNotReadyPods {
+		return errors.New("scheduler has pods not ready (pending or with error)")
```
luizmiranda7, line 404: minor to keep the same signature of the function:
```suggestion
		return errors.New("scheduler has not ready pods (pending or with error)")
```

controller/controller.go
```diff
@@ -390,14 +390,18 @@ func ScaleUp(
...
 	})
 
-	configYAML, _ := models.NewConfigYAML(scheduler.YAML)
+	configYAML, err := models.NewConfigYAML(scheduler.YAML)
+	if err != nil {
+		return fmt.Errorf("failed to load scheduler configuration: %w", err)
+	}
 
-	existPendingPods, err := pendingOrFailedPods(config, redisClient, scheduler.Name, mr)
+	notReadyPods, err := hasNotReadyPods(config, redisClient, scheduler.Name, configYAML.PreventRoomsCreationWithError, mr)
```
luizmiranda7, line 398: Sorry, me again... checking the correct English, we should call it as: `hasNonReadyPods` once `hasNotReadyPods` means that: there's no pod ready.

luizmiranda7 (approved):

PR #178, opened 2021-08-16
Change lint and proto definition structure

gabrielcorado:
This PR consists of a few changes:
* Add `.golangci.yaml` to handle lint configuration;
* No longer running lint on generated files (proto packages, wire, and mocks);
* Buf lint
    * Start running on workflow (through `make lint/protobuf`);
    * Make the necessary lint changes;
* Changed Protobuf structure to follow the API definitions;
    * `schedulers.proto` now only have schedulers actions;
    * Added `operations.proto`, which will handle RPC related to operations;
    * `ping.proto`: The Ping API definition;
    * `messages.proto`: Common messages used across multiple APIs;


internal/api/handlers/management_handler.go
```diff
@@ -39,20 +39,20 @@ import (
 	"google.golang.org/protobuf/types/known/timestamppb"
 )
 
-type SchedulerHandler struct {
+type ManagementHandler struct {
```
luizmiranda7, line 42: If we're going to let all management API handler entry points in this file, I would recommend to let it as dummy as possible, otherwise it will contain logic from several domains that it should not hold. For instance: listing/detailing game rooms and managing schedulers are totally different domains and their treatment should not still in the same handler. WDYT?

gabrielcorado: Do you mean we should have multiple services on top of the "same API", like `SchedulersService` and `OperationsService`, or are you talking about the implementation itself?

luizmiranda7: There's the problem of holding many diff business logic in the same point. We could have multiple services, and multiple handlers, this is a possible solution.

gabrielcorado: Makes sense. My only issue with this approach is that gRPC services don't seem to have a good fit on the resource-oriented API we're designing. For example, let's say the `ListOperations` RPC, in which service/handler would it fall? Looking at its name and what it does, it would probably live into the `OperationsService` or the `OperationHandler`, but currently, we're placing it at the `SchedulersService`.

We don't design our services/handlers based on the API signature to go with this splitting approach.

Another alternative would be to place all of them into the same service and only split them into "logical files", we would have `ManagmentHandler`. Still, it would be implemented in multiple files (like `schedulers.go` and `operations.go`).

luizmiranda7: I believe we should split the operations into another service, indeed! About the alternative, I don't think is the best approach once we still with the big handler with all dependencies injected.

luizmiranda7 (approved):

internal/api/handlers/management_handler.go
gabrielcorado: We've decided to split into two services, `OperationsService` and `SchedulersService`, to keep them very focused on single domains.

luizmiranda7 (approved):

PR #181, opened 2021-08-24
[Management API] Implement add rooms route

luizmiranda7:
## Description

This PR aims to add a new HTTP route to handler "add rooms" calls to a specific scheduler. The HTTP contract follows:

`POST /scheduler/{schedulerName}/add-rooms`

Request payload: `{ amount: int32 }`
Expected successful response: 
- Code: `202`
- Payload: `{ operationId: UUID }`


internal/core/services/scheduler_manager/scheduler_manager.go
```diff
@@ -73,6 +75,23 @@ func (s *SchedulerManager) GetAllSchedulers(ctx context.Context) ([]*entities.Sc
 	return s.schedulerStorage.GetAllSchedulers(ctx)
 }
 
+func (s *SchedulerManager) AddRooms(ctx context.Context, schedulerName string, amount int32) (*operation.Operation, error) {
```
GuilhermBrSp, line 78: @luizmiranda7 would it make sense to have a unit test for this new func? with the initial struct of AddRoomsDefinition that you have created we would be able to write a test for this func using mocks, right?

GuilhermBrSp (approved):

gabrielcorado (approved):

PR #182, opened 2021-08-24
Feature/create update room func for room manager

GuilhermBrSp:
## What ❓ 
Creates a new UpdatRoom function in RoomManager that will be used to update the game room with incoming ping data during ping handler. This PR includes a little refactor that removes SetRoomStatus from RoomStorage interface (port) and its implementation (adapter).

## Obs 👀 
The next PR will add the ping handler along with its proto definition and integration tests.

### DIFF
--- a/internal/core/ports/room_storage.go
+++ b/internal/core/ports/room_storage.go
@@ -32,13 +32,13 @@ import (
 // RoomStorage is an interface for retrieving and updating room status and ping information
 type RoomStorage interface {
 	// GetRoom retrieves a specific room from a scheduler name and roomID
-	// returns an error when the room does not exists
+	// returns an error when the room does not exist
 	GetRoom(ctx context.Context, scheduler string, roomID string) (*game_room.GameRoom, error)
 	// CreateRoom creates a room and returns an error if the room already exists
 	CreateRoom(ctx context.Context, room *game_room.GameRoom) error
-	// UpdateRoom updates a room and returns an error if the room does not exists
+	// UpdateRoom updates a room metadata, status and lastPingAt then publish an status update event. It returns an error if the room does not exist
 	UpdateRoom(ctx context.Context, room *game_room.GameRoom) error
-	// DeleteRoom deletes a room and returns an error if the room does not exists
+	// DeleteRoom deletes a room and returns an error if the room does not exist
 	DeleteRoom(ctx context.Context, scheduler string, roomID string) error
 	// SetRoomStatus sets only the status of a specific room
 	SetRoomStatus(ctx context.Context, scheduler string, roomID string, status game_room.GameRoomStatus) error
--- a/internal/core/ports/room_storage.go
+++ b/internal/core/ports/room_storage.go
@@ -40,9 +40,6 @@ type RoomStorage interface {
 	UpdateRoom(ctx context.Context, room *game_room.GameRoom) error
 	// DeleteRoom deletes a room and returns an error if the room does not exist
 	DeleteRoom(ctx context.Context, scheduler string, roomID string) error
-	// SetRoomStatus sets only the status of a specific room
-	SetRoomStatus(ctx context.Context, scheduler string, roomID string, status game_room.GameRoomStatus) error
-
 	// GetAllRoomIDs gets all room ids in a scheduler
 	GetAllRoomIDs(ctx context.Context, scheduler string) ([]string, error)
 	// GetRoomIDsByLastPing gets all room ids in a scheduler where ping is less than threshold
--- a/internal/core/ports/room_storage.go
+++ b/internal/core/ports/room_storage.go
@@ -36,7 +36,8 @@ type RoomStorage interface {
 	GetRoom(ctx context.Context, scheduler string, roomID string) (*game_room.GameRoom, error)
 	// CreateRoom creates a room and returns an error if the room already exists
 	CreateRoom(ctx context.Context, room *game_room.GameRoom) error
-	// UpdateRoom updates a room metadata, status and lastPingAt then publish an status update event. It returns an error if the room does not exist
+	// UpdateRoom updates a room metadata, status and lastPingAt then publish an status update event
+	// It returns an error if the room does not exist
 	UpdateRoom(ctx context.Context, room *game_room.GameRoom) error
 	// DeleteRoom deletes a room and returns an error if the room does not exist
 	DeleteRoom(ctx context.Context, scheduler string, roomID string) error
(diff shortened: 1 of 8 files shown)


internal/core/services/room_manager/room_manager.go
```diff
@@ -98,10 +98,24 @@ func (m *RoomManager) DeleteRoom(ctx context.Context, gameRoom *game_room.GameRo
...
 		return fmt.Errorf("failed to update game room status to terminating: %w", err)
 	}
 
 	return nil
 }
+
+func (m *RoomManager) UpdateRoom(ctx context.Context, gameRoom *game_room.GameRoom) error {
+	currentGameRoom, err := m.roomStorage.GetRoom(ctx, gameRoom.SchedulerID, gameRoom.ID)
+	if err != nil {
+		return fmt.Errorf("unable to fetch game room from storage: %w", err)
```
GuilhermBrSp, line 114: @gabrielcorado @luizmiranda7 I was inspired by the above functions' error handling when making the code of lines 114 and 118. Do you guys think we should be dealing with those errors differently?

luizmiranda7: Basically, you're encapsulating the error within another, when checking types, if you're planning to handle this error by its type, this wrapping could be a problem. Otherwise, LGTM.

internal/core/ports/room_storage.go
```diff
@@ -32,17 +32,14 @@ import (
 // RoomStorage is an interface for retrieving and updating room status and ping information
 type RoomStorage interface {
 	// GetRoom retrieves a specific room from a scheduler name and roomID
-	// returns an error when the room does not exists
+	// returns an error when the room does not exist
 	GetRoom(ctx context.Context, scheduler string, roomID string) (*game_room.GameRoom, error)
 	// CreateRoom creates a room and returns an error if the room already exists
 	CreateRoom(ctx context.Context, room *game_room.GameRoom) error
-	// UpdateRoom updates a room and returns an error if the room does not exists
+	// UpdateRoom updates a room metadata, status and lastPingAt then publish an status update event. It returns an error if the room does not exist
```
luizmiranda7, line 39:
```suggestion
	// UpdateRoom updates a room metadata, status and lastPingAt then publish an status update event
	// It returns an error if the room does not exist
```

internal/core/services/room_manager/status.go
```diff
@@ -63,19 +63,19 @@ var validStatusTransitions = map[game_room.GameRoomStatus]map[game_room.GameRoom
 	game_room.GameStatusTerminating: {},
 }
 
-// SetRoomStatus changes the game room status in the storage. It takes into
-// count the current status.
-func (m *RoomManager) SetRoomStatus(ctx context.Context, gameRoom *game_room.GameRoom, status game_room.GameRoomStatus) error {
-	transitions, ok := validStatusTransitions[gameRoom.Status]
+// setRoomStatus changes the game room status in the storage. It takes into
+// account the current status.
+func (m *RoomManager) setRoomStatus(ctx context.Context, currentGameRoom *game_room.GameRoom, newGameRoom *game_room.GameRoom) error {
```
luizmiranda7, line 68: I would avoid receiving the "currentGameRoom" as a parameter once someone can make the mistake of calling it twice with the same instance. This could work better if the method searches the room by itself.

internal/core/services/room_manager/room_manager.go
```diff
@@ -98,10 +98,24 @@ func (m *RoomManager) DeleteRoom(ctx context.Context, gameRoom *game_room.GameRo
...
+	}
+	err = m.setRoomStatus(ctx, currentGameRoom, gameRoom)
```
luizmiranda7, line 116: setRoomStatus is doing way more than just setting the room status. If your intent here is just to apply the game room state machine, I would rename it to `validateStatusTransition` and let the update itself living here.

GuilhermBrSp: @luizmiranda7  this isolation definitely makes more sense than the actual implementation! going to change it

internal/core/services/room_manager/room_manager.go
GuilhermBrSp: Not sure if we will need to deal with those specific errors yet, probably going to find out in the next PR. Going to leave those as-is for now.

internal/core/services/room_manager/status.go
GuilhermBrSp: solved in a30c31b

internal/core/services/room_manager/room_manager.go
GuilhermBrSp: solved in a30c31b

luizmiranda7 (approved):

luizmiranda7:
Minor: code coverage is lower... maybe there's something missing on your tests...

internal/core/services/room_manager/room_manager.go
```diff
@@ -98,9 +98,36 @@ func (m *RoomManager) DeleteRoom(ctx context.Context, gameRoom *game_room.GameRo
 		return fmt.Errorf("failed to delete instance on the runtime: %w", err)
 	}
 
-	err = m.SetRoomStatus(ctx, gameRoom, game_room.GameStatusTerminating)
+	newGameRoom := *gameRoom
```
gabrielcorado, line 101: Is there any specific reason for copying the game room? I think it is ok to have this side effect in this function.

internal/core/services/room_manager/room_manager.go
```diff
@@ -98,9 +98,36 @@ func (m *RoomManager) DeleteRoom(ctx context.Context, gameRoom *game_room.GameRo
...
+	if err != nil {
+		return fmt.Errorf("unable to fetch game room from storage: %w", err)
+	}
+
+	err = m.validateRoomStatusTransition(currentGameRoom.Status, gameRoom.Status)
+	if err != nil {
+		return fmt.Errorf("failed when validating game room status transition: %w", err)
+	}
+
+	err = m.roomStorage.UpdateRoom(ctx, gameRoom)
```
gabrielcorado, line 128: Are we expecting the other values to be already filled? (`LastPingAt` and `Metadata`)

GuilhermBrSp: In this case, yes. As I understood, those data will be provided in ping request, right?

internal/core/services/room_manager/room_manager.go
GuilhermBrSp: No specific reason, going to change it!

GuilhermBrSp: 786d396

internal/core/services/room_manager/room_manager.go
gabrielcorado: `Metadata`, yes. The `LastPingAt` is when we receive the ping request, not sure where we're going to "generate it", me it makes sense to be here, WDYT?

GuilhermBrSp: It makes sense to LastPingAt be generated here, definitely!

GuilhermBrSp:
Coverage Increased: 
image

gabrielcorado (approved):

PR #185, opened 2021-08-30
Feature/create rooms ping handler

GuilhermBrSp:
## What ❓ 
Add new `RoomsHandler` for rooms_api, including `UpdateRoomWithPing` function with its proto definition and integrated tests.


internal/api/handlers/rooms_handler.go
```diff
@@ -0,0 +1,99 @@
...
+		return game_room.GameStatusReady, nil
+	case "occupied":
+		return game_room.GameStatusOccupied, nil
+	case "terminating":
+		return game_room.GameStatusTerminating, nil
+	case "error":
+		return game_room.GameStatusError, nil
+	default:
+		return game_room.GameStatusPending, portsErrors.NewErrInvalidArgument("invalid value for GameRoomStatus: %s", value)
+	}
```
luizmiranda7, line 98: We have a "[to string" method](https://github.com/topfreegames/maestro/blob/next/internal/core/entities/game_room/game_room.go#L47-L64) for the status enum, what do you think to place this "from string" in the same place?

internal/api/handlers/rooms_handler_test.go
```diff
@@ -0,0 +1,231 @@
...
+	"net/http"
+	"net/http/httptest"
+	"os"
+	"testing"
+	"time"
+
+	"github.com/topfreegames/maestro/internal/core/entities/game_room"
+	"github.com/topfreegames/maestro/internal/core/ports/errors"
+
+	mock5 "github.com/topfreegames/maestro/internal/adapters/clock/mock"
```
luizmiranda7, line 39: WDYT of calling it clock_mock?

internal/api/handlers/rooms_handler_test.go
```diff
@@ -0,0 +1,231 @@
...
+
+	"github.com/golang/mock/gomock"
+	"github.com/grpc-ecosystem/grpc-gateway/v2/runtime"
+	"github.com/stretchr/testify/require"
+	mock3 "github.com/topfreegames/maestro/internal/adapters/instance_storage/mock"
```
luizmiranda7, line 44: WDYT of calling it as instance_storage_mock?

internal/api/handlers/rooms_handler_test.go
```diff
@@ -0,0 +1,231 @@
...
+	mock2 "github.com/topfreegames/maestro/internal/adapters/port_allocator/mock"
```
luizmiranda7, line 45: Same here...

internal/api/handlers/rooms_handler_test.go
```diff
@@ -0,0 +1,231 @@
...
+	"github.com/topfreegames/maestro/internal/adapters/room_storage/mock"
+	mock4 "github.com/topfreegames/maestro/internal/adapters/runtime/mock"
```
luizmiranda7, line 47: And here!

internal/api/handlers/rooms_handler_test.go
```diff
@@ -0,0 +1,231 @@
...
+		req, err := http.NewRequest(http.MethodPost, "/scheduler/scheduler-name-1/rooms/room-name-1/ping", bytes.NewReader(request))
+		require.NoError(t, err)
+
+		rr := httptest.NewRecorder()
+		mux.ServeHTTP(rr, req)
+
+		require.Equal(t, 500, rr.Code)
+	})
+
+	t.Run("with valid request when the new game room state transition is invalid then it should return with status code 500", func(t *testing.T) {
```
luizmiranda7, line 170: InternalServerError or Conflict? Maybe a UnprocessableEntity... is this taking into count the current maestro ping contract?

internal/api/handlers/rooms_handler.go
GuilhermBrSp: Good ideia

internal/api/handlers/rooms_handler_test.go
GuilhermBrSp: maestro-client currently does not consider the response status code, it would make sense to return a UnprocessableEntity code in this case, but I wasn't able to find an equivalent status code on the GRPC protocol which we are using.

GuilhermBrSp: mestro-client code that i looked upon:
image

internal/api/handlers/rooms_handler_test.go
GuilhermBrSp: f7d5d2b

internal/api/handlers/rooms_handler_test.go
GuilhermBrSp: f7d5d2b

internal/api/handlers/rooms_handler_test.go
GuilhermBrSp: f7d5d2b

internal/api/handlers/rooms_handler_test.go
GuilhermBrSp: f7d5d2b

internal/api/handlers/rooms_handler.go
GuilhermBrSp: f7d5d2b

luizmiranda7 (approved):

internal/core/entities/game_room/game_room.go
```diff
@@ -25,6 +25,8 @@ package game_room
 import (
 	"fmt"
 	"time"
+
+	portsErrors "github.com/topfreegames/maestro/internal/core/ports/errors"
```
gabrielcorado, line 29: We should use those errors only for Ports/Adapters, which is not the case here. If you feel like we need the same kind o package for defining errors on the entities/services, we can set up a different one.

internal/core/services/room_manager/room_manager.go
```diff
@@ -115,7 +115,7 @@ func (m *RoomManager) DeleteRoom(ctx context.Context, gameRoom *game_room.GameRo
 func (m *RoomManager) UpdateRoom(ctx context.Context, gameRoom *game_room.GameRoom) error {
 	currentGameRoom, err := m.roomStorage.GetRoom(ctx, gameRoom.SchedulerID, gameRoom.ID)
 	if err != nil {
-		return fmt.Errorf("unable to fetch game room from storage: %w", err)
+		return err
```
gabrielcorado, line 118: Any reason for dropping this error wrapping?

gabrielcorado (approved):
LGTM. Added some small comments.

internal/core/services/room_manager/room_manager.go
GuilhermBrSp: Yes, we are handling this error at RoomsHandler to return with NotFound 404 when no game room is found.

luizmiranda7 (approved):

PR #186, opened 2021-09-02
Feature/implement add rooms operation

GuilhermBrSp:
## What ❓ 
- Implement AddRooms executor, including tests and dependency injection
- Update RoomManager_CreateRoom function, it now waits for the created game room status to achieve the "ready" state.
- Create a new `roomInitializationTimeoutMilis` config, this will define the timeout value used in CreateRoom waiting logic

### DIFF
--- a/config/rooms-api.local.yaml
+++ b/config/rooms-api.local.yaml
@@ -22,6 +22,3 @@ adapters:
     kubernetes:
       masterUrl: "https://127.0.0.1:6443"
       kubeconfig: "./kubeconfig.yaml"
-services:
-  roomManager:
-    roomInitializationTimeoutMilis: 10000
--- a/config/rooms-api.local.yaml
+++ b/config/rooms-api.local.yaml
@@ -22,3 +22,6 @@ adapters:
     kubernetes:
       masterUrl: "https://127.0.0.1:6443"
       kubeconfig: "./kubeconfig.yaml"
+services:
+  roomManager:
+    roomInitializationTimeoutMilis: 10000
--- a/internal/core/operations/add_rooms/add_rooms_definition.go
+++ b/internal/core/operations/add_rooms/add_rooms_definition.go
@@ -38,11 +38,7 @@ type AddRoomsDefinition struct {
 }
 
 func (d *AddRoomsDefinition) ShouldExecute(_ context.Context, _ []*operation.Operation) bool {
-	if d.Amount <= 10 {
-		return true
-	} else {
-		return false
-	}
+	return d.Amount <= 10
 }
 
 func (d *AddRoomsDefinition) Name() string {
--- a/internal/core/operations/add_rooms/add_rooms_definition.go
+++ b/internal/core/operations/add_rooms/add_rooms_definition.go
@@ -38,7 +38,7 @@ type AddRoomsDefinition struct {
 }
 
 func (d *AddRoomsDefinition) ShouldExecute(_ context.Context, _ []*operation.Operation) bool {
-	return d.Amount <= 10
+	return true
 }
 
 func (d *AddRoomsDefinition) Name() string {
--- a/internal/core/operations/executor.go
+++ b/internal/core/operations/executor.go
@@ -25,19 +25,21 @@ package operations
 import (
 	"context"
 
+	"go.uber.org/zap"
+
 	"github.com/topfreegames/maestro/internal/core/entities/operation"
 )
 
 // Executor is where the actual operation logic is implemented, and it will
 // receive as input its correlated definition.
 type Executor interface {
-	// This is where the operation logic will live; it will receive a context
+	// Execute is where the operation logic will live; it will receive a context
 	// that will be used for deadline and cancelation. This function has only
 	// one return which is the operation error (if any);
-	Execute(ctx context.Context, op *operation.Operation, definition Definition) error
-	// This function is called if Execute returns an error. This will be used
+	Execute(ctx context.Context, op *operation.Operation, definition Definition, logger *zap.Logger) error
+	// OnError is called if Execute returns an error. This will be used
 	// for operations that need to do some cleanup or any process if it fails.
-	OnError(ctx context.Context, op *operation.Operation, definition Definition, executeErr error) error
+	OnError(ctx context.Context, op *operation.Operation, definition Definition, executeErr error, logger *zap.Logger) error
 	// Name returns the executor name. This is used to identify a executor for a
 	// definition.
 	Name() string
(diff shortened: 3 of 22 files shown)


internal/core/services/room_manager/room_manager.go
```diff
@@ -82,6 +85,18 @@ func (m *RoomManager) CreateRoom(ctx context.Context, scheduler entities.Schedul
 		return nil, nil, err
 	}
 
+	// TODO: let each scheduler parametrize its timeout and use this config as fallback if the scheduler timeout value
```
GuilhermBrSp, line 88: @gabrielcorado @luizmiranda7 What do you guys think about this idea? the initialization timeout value of the scheduler's containers will always be the same? would it make sense to be able to define this value for each individual scheduler?

internal/core/operations/executor.go
```diff
@@ -25,19 +25,21 @@ package operations
...
 // Executor is where the actual operation logic is implemented, and it will
 // receive as input its correlated definition.
 type Executor interface {
-	// This is where the operation logic will live; it will receive a context
+	// Execute is where the operation logic will live; it will receive a context
 	// that will be used for deadline and cancelation. This function has only
 	// one return which is the operation error (if any);
-	Execute(ctx context.Context, op *operation.Operation, definition Definition) error
-	// This function is called if Execute returns an error. This will be used
+	Execute(ctx context.Context, op *operation.Operation, definition Definition, logger *zap.Logger) error
```
gabrielcorado, line 39: Any specific reason for adding the logger as an argument on the `Executor`? In other places, we're following the pattern of using the "static" logger instance instead of injecting it (like [here](https://github.com/topfreegames/maestro/blob/next/internal/core/workers/operation_execution_worker/operation_execution_worker.go#L60), for example).

internal/core/services/room_manager/room_manager.go
```diff
@@ -37,15 +38,17 @@ type RoomManager struct {
 	roomStorage     ports.RoomStorage
 	instanceStorage ports.GameRoomInstanceStorage
 	runtime         ports.Runtime
+	configs         config.Config
 }
 
-func NewRoomManager(clock ports.Clock, portAllocator ports.PortAllocator, roomStorage ports.RoomStorage, instanceStorage ports.GameRoomInstanceStorage, runtime ports.Runtime) *RoomManager {
+func NewRoomManager(clock ports.Clock, portAllocator ports.PortAllocator, roomStorage ports.RoomStorage, instanceStorage ports.GameRoomInstanceStorage, runtime ports.Runtime, configs config.Config) *RoomManager {
```
gabrielcorado, line 44: Until now, we were avoiding having references to the "configuration layer" on the core/adapters (like [here](https://github.com/topfreegames/maestro/blob/next/internal/service/adapters.go#L97)) but since we didn't discuss this topic too much in-depth, what's your input on this?

My thoughts: By doing this, it seems to me that we're starting to attach too much of the service to the "infrastructure" the commands are using. Having a service initializer that takes each configuration attribute doesn't seem to scale that well. Still, alternatively, we can have a `RoomManagerOptions` struct to rely on instead of the configuration itself.

internal/core/operations/add_rooms/add_rooms_definition.go
```diff
@@ -38,7 +38,11 @@ type AddRoomsDefinition struct {
 }
 
 func (d *AddRoomsDefinition) ShouldExecute(_ context.Context, _ []*operation.Operation) bool {
-	return true
+	if d.Amount <= 10 {
```
gabrielcorado, line 41: nit: we can do just the return here `return d.Amount <= 10` instead of the if/else clause.

internal/core/operations/executor.go
luizmiranda7: The idea here is to have the same metadata assumed on the OperationExecutionWorker: scheduler name, operation id, and so on...  But I'm not comfortable having it on the interface...

internal/core/services/room_manager/room_manager.go
luizmiranda7: By now I don't see the need of having it for each scheduler, we can watch it in production later on. Basically, we can have a high configured timeout to hold every room addition.

internal/core/services/room_manager/room_manager.go
luizmiranda7: Good idea, actually, devs on project manager are converting the entire YAML configuration to a struct and passing to the service just the scoped piece. We could follow the same approach, WDYT?

In any case, I would recommend opening a tech debt for that once we use the configuration in all over the code.

internal/core/operations/executor.go
GuilhermBrSp: @gabrielcorado, @luizmiranda7 described the initial idea that I had, but seeing it again we already have all of that info inside the executor and we wouldn't need to receive the logger as an argument. For preserving the initial interface definition, I prefer to use the "static" logger

internal/core/services/room_manager/room_manager.go
GuilhermBrSp: @gabrielcorado @luizmiranda7 In this case, the first solution that comes to mind is creating a new port (e.g  RoomManagerOptions or RoomManagerConfig) with its respective adapter to maintain the architecture layers isolated

config/rooms-api.local.yaml
```diff
@@ -22,3 +22,6 @@ adapters:
     kubernetes:
       masterUrl: "https://127.0.0.1:6443"
       kubeconfig: "./kubeconfig.yaml"
+services:
+  roomManager:
+    roomInitializationTimeoutMilis: 10000
```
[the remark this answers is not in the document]
GuilhermBrSp: We don't need it yet, rooms API is not using the CreateRoom function, going to remove it

internal/core/operations/add_rooms/add_rooms_definition.go
GuilhermBrSp: 🤦

GuilhermBrSp: 014ab84

internal/core/services/room_manager/room_manager.go
gabrielcorado: Not sure if we require having one port for each of our services. Also, the configuration struct could live in this package and be initialized by our "infrastructure" code (like we do in the case of the adapter). WDYT?

luizmiranda7 (approved):

gabrielcorado (approved):

PR #189, opened 2021-09-03
Remove rooms executor

gabrielcorado:
This is the first version of the `RemoveRooms` executor, where it will receive an `amount` of rooms to be removed/deleted, and it won't wait until they are completely removed.

This PR also introduces the `ListRoomsWithDeletionPriority`, which will be responsible for returning the rooms that should be deleted first. For now, this function has a simple implementation that returns all rooms without any rule on it.

### DIFF
--- a/internal/core/operations/remove_rooms/remove_rooms_executor.go
+++ b/internal/core/operations/remove_rooms/remove_rooms_executor.go
@@ -24,24 +24,51 @@ package remove_rooms
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/topfreegames/maestro/internal/core/entities/operation"
 	"github.com/topfreegames/maestro/internal/core/operations"
+	"github.com/topfreegames/maestro/internal/core/services/room_manager"
+	"go.uber.org/zap"
 )
 
-type RemoveRoomsExecutor struct{}
-
-func NewExecutor() *RemoveRoomsExecutor {
-	return &RemoveRoomsExecutor{}
+type RemoveRoomsExecutor struct {
+	roomManager *room_manager.RoomManager
 }
 
-func (e *RemoveRoomsExecutor) Execute(_ context.Context, _ *operation.Operation, _ operations.Definition) error {
-	// TODO(gabrielcorado): implement it
+func NewExecutor(roomManager *room_manager.RoomManager) *RemoveRoomsExecutor {
+	return &RemoveRoomsExecutor{roomManager}
+}
+
+func (e *RemoveRoomsExecutor) Execute(ctx context.Context, op *operation.Operation, definition operations.Definition) error {
+	removeDefiniton := definition.(*RemoveRoomsDefinition)
+	rooms, err := e.roomManager.ListRoomsWithDeletionPriority(ctx, op.SchedulerName, removeDefiniton.Amount)
+	if err != nil {
+		return fmt.Errorf("failed to list rooms to delete: %w", err)
+	}
+
+	logger := zap.L().With(
+		zap.String("scheduler_name", op.SchedulerName),
+		zap.String("operation_definition", definition.Name()),
+		zap.String("operation_id", op.ID),
+	)
+
+	logger.Debug("start deleting rooms", zap.Int("amonut", len(rooms)))
+
+	for _, room := range rooms {
+		err = e.roomManager.DeleteRoom(ctx, room)
+		if err != nil {
+			reportDeletionFailedTotal(op.SchedulerName, op.ID)
+			logger.Warn("failed to remove rooms", zap.Error(err))
+		}
+	}
+
+	logger.Debug("finished deleting rooms")
 	return nil
 }
 
+// OnError will do nothing.
 func (e *RemoveRoomsExecutor) OnError(_ context.Context, _ *operation.Operation, _ operations.Definition, _ error) error {
-	// TODO(gabrielcorado): implement it
 	return nil
 }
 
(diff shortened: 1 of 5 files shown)


internal/core/services/room_manager/room_manager_test.go
```diff
@@ -218,3 +219,70 @@ func TestRoomManager_UpdateRoom(t *testing.T) {
...
+	)
+
+	t.Run("when there enough rooms it should return the specified number", func(t *testing.T) {
+		schedulerName := "test-scheduler"
+		availableRooms := []*game_room.GameRoom{
+			{ID: "first-room", SchedulerID: schedulerName, Status: game_room.GameStatusReady},
+			{ID: "second-room", SchedulerID: schedulerName, Status: game_room.GameStatusReady},
+			{ID: "third-room", SchedulerID: schedulerName, Status: game_room.GameStatusReady},
+			{ID: "forth-room", SchedulerID: schedulerName, Status: game_room.GameStatusReady},
+		}
```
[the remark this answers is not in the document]
gabrielcorado: This test ensures that even if there are more rooms, the function will only return the `amount` specified.

internal/core/operations/remove_rooms/remove_rooms_executor.go
```diff
@@ -24,24 +24,51 @@ package remove_rooms
...
-	// TODO(gabrielcorado): implement it
+func (e *RemoveRoomsExecutor) Execute(ctx context.Context, op *operation.Operation, definition operations.Definition) error {
+	removeDefiniton := definition.(*RemoveRoomsDefinition)
+	rooms, err := e.roomManager.ListRoomsWithDeletionPriority(ctx, op.SchedulerName, removeDefiniton.Amount)
+	if err != nil {
+		return fmt.Errorf("failed to list rooms to delete: %w", err)
+	}
+
+	logger := zap.L().With(
+		zap.String("scheduler_name", op.SchedulerName),
```
GuilhermBrSp, line 51: @gabrielcorado what do you think about including the `service` key in this logger as we are doing in other places? Ex: zap.String("service", "worker")

internal/core/operations/remove_rooms/remove_rooms_executor_test.go
```diff
@@ -0,0 +1,121 @@
...
+	"github.com/topfreegames/maestro/internal/adapters/room_storage/mock"
+	runtime_mock "github.com/topfreegames/maestro/internal/adapters/runtime/mock"
+
+	"github.com/topfreegames/maestro/internal/core/entities/game_room"
+	"github.com/topfreegames/maestro/internal/core/entities/operation"
+	porterrors "github.com/topfreegames/maestro/internal/core/ports/errors"
+	"github.com/topfreegames/maestro/internal/core/services/room_manager"
+)
+
+func TestExecute(t *testing.T) {
```
GuilhermBrSp, line 46: @gabrielcorado missing test case when `ListRoomsWithDeletionPriority` method returns with an error, what do you think about including this one here?

internal/core/services/room_manager/room_manager.go
```diff
@@ -132,3 +132,33 @@ func (m *RoomManager) UpdateRoom(ctx context.Context, gameRoom *game_room.GameRo
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to list scheduler rooms: %w", err)
+	}
+
+	var result []*game_room.GameRoom
+	for _, roomID := range schedulerRoomsIDs {
+		room, err := m.roomStorage.GetRoom(ctx, schedulerName, roomID)
+		if err != nil {
+			// TODO(gabrielcorado): should we fail the entire process because of
+			// a room missing?
```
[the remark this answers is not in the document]
GuilhermBrSp: In what case we will have a missing room with an existent roomID from the scheduler?

internal/core/operations/remove_rooms/remove_rooms_executor_test.go
GuilhermBrSp, line 46: @gabrielcorado what do you think about adding an extra test case here: when ListRoomsWithDeletionPriority returns with an error, it's one of the branches of the execution flow, right?

internal/core/services/room_manager/room_manager.go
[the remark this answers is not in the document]
gabrielcorado: No, because we list them all, and they start getting more of their details.

My point here was more related to if this happens, we have an inconsistent state. In this case, what should we do:
* recover somehow? let's say clean the room with error from the list;
* add into a metric and then just measure those inconsistencies (seems the best first version here, altho I don't want to leave too many inconsistencies and deal with them later);

internal/core/operations/remove_rooms/remove_rooms_executor_test.go
gabrielcorado: It makes sense. I'll add it.

internal/core/operations/remove_rooms/remove_rooms_executor.go
gabrielcorado: Does it make sense in this file to add this reference? The operation itself doesn't know who is actually calling it. If we assume where it is going to run, I'd be in favor of leaving without it and, since we are deploying our workers in different commands, we can add those labels into the global instance (which we derive from);

GuilhermBrSp: I was thinking more about how it will be when making queries for logs, say that we want to see logs filtering the worker cmd, without this label would It be possible? I saw that we were populating this `service` label in some logger parents, I'm worried that those generated logs would "escape" from the standard we are creating with those labels. Adding this label in the global instance seems to be the better approach here.

GuilhermBrSp (approved):
Need to fix the lint

PR #190, opened 2021-09-03
Feature/implement remove rooms handler

GuilhermBrSp:
Implement RemoveRooms func for SchedulerManager and SchedulerHandler, including the remove-rooms proto definition


internal/api/handlers/schedulers_handler.go
```diff
@@ -102,6 +102,21 @@ func (h *SchedulersHandler) AddRooms(ctx context.Context, request *api.AddRoomsR
 	}, nil
 }
 
+func (h *SchedulersHandler) RemoveRooms(ctx context.Context, request *api.RemoveRoomsRequest) (*api.RemoveRoomsResponse, error) {
+
+	operation, err := h.schedulerManager.RemoveRooms(ctx, request.GetSchedulerName(), request.GetAmount())
+	if errors.Is(err, portsErrors.ErrNotFound) {
```
gabrielcorado, line 108: Note (not really actionable): I think we are starting to "leak" those internal errors (port errors) into the layers we shouldn't (like the API handlers here). We might need to discuss introducing a `errors` package for the services or develop a different solution.

internal/core/operations/remove_rooms/remove_rooms_definition.go
```diff
@@ -34,7 +34,7 @@ import (
 const OperationName = "remove_rooms"
 
 type RemoveRoomsDefinition struct {
-	Amount int `json:"amount"`
+	Amount int32 `json:"amount"`
```
gabrielcorado, line 37: Question: Is the only reason for changing it to `int32` is to make the ProtoBuf conversion easier? If yes, should we do it? I think we shouldn't enforce anything in the core because of the API definition.

internal/api/handlers/schedulers_handler.go
GuilhermBrSp: @gabrielcorado Either way we would have our API layer dealing with internal errors from the Core layer since our services are core domain logic, I don't see a problem with that. But I get your point about having a "translation" of internal errors to more "semantic" ones, defining and isolating the possible errors that our `services` could return, I like this idea.

internal/core/operations/remove_rooms/remove_rooms_definition.go
GuilhermBrSp: ffdf53ba009fb7ff41b42989410c87307e736633

internal/api/handlers/schedulers_handler.go
gabrielcorado: You're right, and the API still relies on the core errors. However, we could rely on errors from the closest layer (in this case, the services) instead of some deep errors (like the ports one).

Anyway, WDYT of taking it as technical debt for now, and we can tackle it later when we have more samples of this use case?

gabrielcorado (approved):

internal/api/handlers/schedulers_handler.go
GuilhermBrSp: I Agree! :)

PR #191, opened 2021-09-09
Add scheduler events model + storage

luizmiranda7:
## Description

This PR aims to add a scheduler event model + persistence layer. The event will consist of the following fields:
* Name: upper and snake case values;
* Scheduler name;
* Created at: the date and time when the event was created;
* Metadata: a map of values to give more context about the event;

Maestro will store the events in a Redis layer


storage/redis/redis_scheduler_event_storage.go
```diff
@@ -0,0 +1,79 @@
...
+	redisClient interfaces.RedisClient
+}
+
+func NewRedisSchedulerEventStorage(redisClient interfaces.RedisClient) *RedisSchedulerEventStorage {
+	return &RedisSchedulerEventStorage {
+		redisClient: redisClient,
+	}
+}
+
+func (this *RedisSchedulerEventStorage) PersistSchedulerEvent(event *models.SchedulerEvent) error {
```
gabrielcorado, line 29: WDYT about keeping the events healthy by removing old entries (for example, doing a `ZREMRANGEBYRANK key MAXAGE -2`) while persisting in new events.

storage/redis/redis_scheduler_event_storage.go
```diff
@@ -0,0 +1,79 @@
...
+	return nil
+}
+
+// Load loads scheduler events from the redis using the scheduler name and a page
+func (this *RedisSchedulerEventStorage) LoadSchedulerEvents(schedulerName string, page int) ([]*models.SchedulerEvent, error) {
+	eventsString, err := this.redisClient.ZRangeByScore(getSchedulerEventKey(schedulerName), redis.ZRangeBy{
+		Min:    "-inf",
+		Max:    "+inf",
+		Count:  30,
+		Offset: int64((page-1)*30 + 1),
```
gabrielcorado, line 57: Nit: Make the page size a constant.

storage/redis/redis_scheduler_event_storage.go
```diff
@@ -0,0 +1,79 @@
...
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/go-redis/redis"
+	"github.com/topfreegames/extensions/redis/interfaces"
+	"github.com/topfreegames/maestro/models"
+)
+
+type RedisSchedulerEventStorage struct {
```
gabrielcorado, line 19: The interface is a little different from the implementation. We check if the struct implements it at compile-time by adding lines like `var _ SchedulerEventStorage = (*RedisSchedulerEventStorage)(nil)`. [Like we do on next](https://github.com/topfreegames/maestro/blob/next/internal/adapters/runtime/kubernetes/kubernetes.go#L30).

plugins/grpc/forwarder.go
```diff
@@ -384,3 +384,5 @@ func NewForwarder(config *viper.Viper, logger log.FieldLogger) (eventforwarder.E
 	}
 	return g, nil
 }
+
+func main(){}
```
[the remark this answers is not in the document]
luizmiranda7: The compiler was giving a warning because of forwarder.go is in the main package with no main func.

storage/redis/redis_scheduler_event_storage.go
luizmiranda7: I made something different here:

ZREMRANGEBYSCORE key -inf (<score of 14 days ago>

gabrielcorado (approved):

PR #198, opened 2021-09-20
Create standard env vars on rooms pods

GuilhermBrSp:
## What ❓ 
We have identified that two default game room containers env vars were missing while creating those in maestro next.
This PR add two default env vars on game rooms containers while creating it: `MAESTRO_SCHEDULER_NAME` and `MAESTRO_ROOM_ID`.

### DIFF
--- a/internal/adapters/runtime/kubernetes/game_room_convert.go
+++ b/internal/adapters/runtime/kubernetes/game_room_convert.go
@@ -24,6 +24,7 @@ package kubernetes
 
 import (
 	"fmt"
+	uuid "github.com/satori/go.uuid"
 	"strings"
 
 	"github.com/topfreegames/maestro/internal/core/entities/game_room"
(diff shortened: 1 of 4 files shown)


internal/adapters/runtime/kubernetes/game_room_convert.go
```diff
@@ -61,10 +62,12 @@ var invalidPodWaitingStates = []string{
 }
 
 func convertGameRoomSpec(schedulerID string, gameRoomSpec game_room.Spec) (*v1.Pod, error) {
+	randID := strings.SplitN(uuid.NewV4().String(), "-", 2)[0]
```
GuilhermBrSp, line 65: inspired by https://github.com/topfreegames/maestro/blob/master/models/game_room.go#L158

luizmiranda7 (approved):

internal/adapters/runtime/kubernetes/game_room_convert.go
gabrielcorado: We should be careful with the name of the pods we generate (that's why we were delegating to Kubernetes to do it).

WDYT about using the same logic the `GenerateName`? We would be just changing to our side so that we can place our environment variable. Ref: https://github.com/kubernetes/apiserver/blob/master/pkg/storage/names/generate.go#L40

gabrielcorado: Alternatively, we can add an environment variable that references the Pod resource field. An example of this in YAML format would be:
```yaml
- name: MAESTRO_ROOM_ID
  valueFrom:
    fieldRef:
      fieldPath: status.podIP
```

Ref: https://pkg.go.dev/k8s.io/api/core/v1#EnvVarSource

GuilhermBrSp: 3c1b0a8770b93a11143f0d240fa8153e58e31262

PR #203, opened 2021-09-22
Runtime watcher update game room instance when receive added event

gabrielcorado:
Implementing the `Start`, `Stop`, and `IsRunning` of the runtime watcher. Implementing processing of `InstanceEventTypeAdded` event.

This PR also adds a new function to the `RoomManager` used to update the game room instance (`UpdateRoomInstance`).

### DIFF
--- a/internal/core/workers/runtime_watcher_worker/runtime_watcher_worker.go
+++ b/internal/core/workers/runtime_watcher_worker/runtime_watcher_worker.go
@@ -68,6 +68,8 @@ func (w *runtimeWatcherWorker) Start(ctx context.Context) error {
 	}
 
 	w.ctx, w.cancelFunc = context.WithCancel(ctx)
+	defer w.cancelFunc()
+
 	resultChan := watcher.ResultChan()
 	for {
 		select {
--- a/internal/core/workers/runtime_watcher_worker/runtime_watcher_worker.go
+++ b/internal/core/workers/runtime_watcher_worker/runtime_watcher_worker.go
@@ -24,9 +24,14 @@ package runtime_watcher_worker
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/topfreegames/maestro/internal/core/entities"
+	"github.com/topfreegames/maestro/internal/core/entities/game_room"
+	"github.com/topfreegames/maestro/internal/core/ports"
+	"github.com/topfreegames/maestro/internal/core/services/room_manager"
 	"github.com/topfreegames/maestro/internal/core/workers"
+	"go.uber.org/zap"
 )
 
 var _ workers.Worker = (*runtimeWatcherWorker)(nil)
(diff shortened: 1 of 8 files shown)


internal/core/workers/runtime_watcher_worker/runtime_watcher_worker_test.go
```diff
@@ -0,0 +1,141 @@
...
+		config := room_manager.RoomManagerConfig{RoomInitializationTimeoutMillis: time.Millisecond * 1000}
+		roomManager := room_manager.NewRoomManager(fakeClock, portAllocator, roomStorage, instanceStorage, runtime, config)
+
+		return mockCtrl, instanceStorage, runtime, &workers.WorkerOptions{
+			Runtime:     runtime,
+			RoomManager: roomManager,
+		}
+	}
+
+	t.Run("fails to start watcher", func(t *testing.T) {
```
GuilhermBrSp, line 65: @gabrielcorado WDYT about adding one test case for when `w.roomManager.UpdateRoomInstance` call fails?

GuilhermBrSp (approved):

internal/core/workers/runtime_watcher_worker/runtime_watcher_worker_test.go
```diff
@@ -0,0 +1,141 @@
...
+		mockCtrl, _, runtime, workerOptions := workerOptions(t)
```
GuilhermBrSp, line 66: Also, is there some way that we can test the `IsRunning` method?

internal/core/workers/runtime_watcher_worker/runtime_watcher_worker.go
```diff
@@ -24,29 +24,84 @@ package runtime_watcher_worker
...
 var _ workers.Worker = (*runtimeWatcherWorker)(nil)
 
 // runtimeWatcherWorker is a work that will watch for changes on the Runtime
 // and apply to keep the Maestro state up-to-date. It is not expected to perform
 // any action over the game rooms or the scheduler.
-type runtimeWatcherWorker struct{}
+type runtimeWatcherWorker struct {
+	scheduler   *entities.Scheduler
+	roomManager *room_manager.RoomManager
+	// TODO(gabrielcorado): should we access the port directly? do we need to
```
GuilhermBrSp, line 45: @gabrielcorado  I didn't get your point here

internal/core/workers/runtime_watcher_worker/runtime_watcher_worker_test.go
gabrielcorado: Sure.

internal/core/workers/runtime_watcher_worker/runtime_watcher_worker_test.go
gabrielcorado: I've added a check on the current tests.

internal/core/workers/runtime_watcher_worker/runtime_watcher_worker.go
gabrielcorado: Until now, we're only accessing ports from core services. The workers are not services, so should we add this "interface" to some of our services?

PR #205, opened 2021-09-23
Runtime watcher instance deleted event

gabrielcorado:
Implements the process of instance deleted events. It consists of cleaning the room state, which is done by a newly introduced `CleanRoomState` function from `RoomManager`.

NOTE: Right now, this PR is pointing to the branch `feature/runtime-pod-created` (which is the #203 PR). After the #203 is closed, this one should be rebased and the target changed to `next`.

### DIFF
--- a/internal/core/workers/runtime_watcher_worker/runtime_watcher_worker_test.go
+++ b/internal/core/workers/runtime_watcher_worker/runtime_watcher_worker_test.go
@@ -44,7 +44,7 @@ import (
 )
 
 func TestProcessRuntimeEvents(t *testing.T) {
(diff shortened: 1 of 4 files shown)


GuilhermBrSp (approved):

internal/core/workers/runtime_watcher_worker/runtime_watcher_worker_test.go
```diff
@@ -138,4 +138,63 @@ func TestProcessRuntimeEvents(t *testing.T) {
 			return true
 		}, time.Second, time.Millisecond)
 	})
+
+	t.Run("updates clean room state on delete event", func(t *testing.T) {
```
GuilhermBrSp, line 142: @gabrielcorado WDYT about adding a test case for when `w.roomManager.CleanRoomState` fails?

gabrielcorado: Sure.

PR #208, opened 2021-09-27
Feature: operation cancelation e2e test

luizmiranda7:
## Description

This PR aims to add an E2E test over the cancel operation feature. Basically, it follows the script:

1. Creates a scheduler;
2. Evaluates the scheduler is "online";
3. Enqueues a test operation; when executing it, it sleeps during the given time (in seconds);
4. Ensures the test operation is running;
5. Calls the Management API to cancel the test operation;
6. Evaluates the operation is finished;

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -48,7 +48,7 @@ run/e2e-tests: deps/stop build/worker build/management-api build/rooms-api build
 
 .PHONY: build/worker
 build/worker:
-	@rm -f ./bin/worker-*
+	@rm -f ./bin/worker-* || true
 	@go build -o ./bin/worker ./cmd/worker
 	@env GOOS=linux GOARCH=amd64 go build -o ./bin/worker-linux-x86_64 ./cmd/worker
 
@@ -58,7 +58,7 @@ run/worker: build/worker
 
 .PHONY: build/management-api
 build/management-api:
-	@rm -f ./bin/management-api-*
+	@rm -f ./bin/management-api-* || true
 	@go build -o ./bin/management-api ./cmd/management_api
 	@env GOOS=linux GOARCH=amd64 go build -o ./bin/management-api-linux-x86_64 ./cmd/management_api
 
@@ -68,7 +68,7 @@ run/management-api: build/management-api
 
 .PHONY: build/rooms-api
 build/rooms-api:
-	@rm -f ./bin/rooms-api-*
+	@rm -f ./bin/rooms-api-* || true
 	@go build -o ./bin/rooms-api ./cmd/rooms_api
 	@env GOOS=linux GOARCH=amd64 go build -o ./bin/rooms-api-linux-x86_64 ./cmd/rooms_api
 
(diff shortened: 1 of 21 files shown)


internal/core/operations/test_operation/test_operation_executor.go
```diff
@@ -0,0 +1,68 @@
...
+func NewExecutor() *TestOperationExecutor {
+	return &TestOperationExecutor{}
+}
+
+func (e *TestOperationExecutor) Execute(ctx context.Context, _op *operation.Operation, definition operations.Definition) error {
+	testOperationDefinition := definition.(*TestOperationDefinition)
+
+	zap.L().Sugar().Infof("sleeping routine for %d seconds", testOperationDefinition.SleepSeconds)
+
+	ticker := time.NewTicker(1 * time.Second)
```
gabrielcorado, line 45: Any specific reason for using a ticker here? We could remove the necessity of the `for` below by using a select with a timer:

```go
select {
case <-time.After(time.Second * testOperationDefinition.SleepSeconds):
case <-ctx.Done():
	return ctx.Err()
}
```

The `select` will block the execution until it receives something in any of the channels.

e2e/framework/wrappers.go
```diff
@@ -26,13 +26,14 @@ import (
...
 	"github.com/go-redis/redis"
+	redisV8 "github.com/go-redis/redis/v8"
 
 	"github.com/topfreegames/maestro/e2e/framework/maestro"
 
 	"k8s.io/client-go/kubernetes"
 )
 
-func WithClients(t *testing.T, testCase func(apiClient *APIClient, kubeClient kubernetes.Interface, redisClient *redis.Client, maestro *maestro.MaestroInstance)) {
+func WithClients(t *testing.T, testCase func(apiClient *APIClient, kubeClient kubernetes.Interface, redisClient *redis.Client, redisClientV8 *redisV8.Client, maestro *maestro.MaestroInstance)) {
```
gabrielcorado, line 36: Can we set a `TODO` here to remove the "old" client when all the ports are migrated into the v8?

Makefile
```diff
@@ -48,35 +48,39 @@ run/e2e-tests: deps/stop
 
 .PHONY: build/worker
 build/worker:
-	@rm -f ./bin/worker
+	@rm -f ./bin/worker-*
```
gabrielcorado, line 51: Performing an `rm` with a wildcard that doesn't match any file, returns an error similar to this:
```
No matches for wildcard './bin/worker-*'. See `help expand`.
```

To not break, we can change the command to always exit with 0, meaning it won't break the target. It can be done by adding a `|| true` at the end, having something like `@rm -f ./bin/worker-* || true`.

internal/core/operations/test_operation/test_operation_executor.go
luizmiranda7: I need the context to be evaluated every second. So I need this sleep to look (second by second) to ctx.Done() channel

GuilhermBrSp (approved):

PR #214, opened 2021-10-01
Compose runtime and ping status

gabrielcorado:
This PR changes how we defined the Game Room status. It is now composed of both: Ping status and Runtime status.

A few other changes were necessary due to this core change:
- `RoomStorage` changes;
    - Base room key is now Hash, which contains the room metadata and the ping status;
    - `UpdateRoom` now only updates ping status;
    - New `UpdateRoomStatus` function that updates only the room status;
- `RoomManager` changes;
    - `UpdateInstance` and `UpdateRoom` now calls internal `updateGameRoomStatus` which is going to update the final room status if necessary;
- Rooms ping API now updates the ping status instead of the game room status;

NOTE: This PR is not going to change the e2e (although it might break it). The change is going to be done in future PR.

### DIFF
--- a/internal/api/handlers/rooms_handler_test.go
+++ b/internal/api/handlers/rooms_handler_test.go
@@ -52,13 +52,7 @@ import (
 func TestRoomsHandler_UpdateRoomWithPing(t *testing.T) {
 	dirPath, _ := os.Getwd()
 
-	gameRoom := &game_room.GameRoom{
-		ID:          "",
-		SchedulerID: "",
-		Status:      game_room.GameStatusReady,
-		Metadata:    nil,
-		LastPingAt:  time.Time{},
-	}
+	instance := &game_room.Instance{Status: game_room.InstanceStatus{Type: game_room.InstanceReady}}
 
 	validRequests, _ := ioutil.ReadFile(dirPath + "/fixtures/valid-ping-data-list.json")
 	var validRawRequests []*json.RawMessage
@@ -71,7 +65,6 @@ func TestRoomsHandler_UpdateRoomWithPing(t *testing.T) {
 	require.NoError(t, err)
 
 	t.Run("with valid request and existent game room it should return status code 200 with success = true", func(t *testing.T) {
-
 		mockCtrl := gomock.NewController(t)
 		defer mockCtrl.Finish()
 		clockMock := clock_mock.NewFakeClock(time.Now())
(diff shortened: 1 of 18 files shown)


internal/api/handlers/rooms_handler_test.go
```diff
@@ -185,8 +192,19 @@ func TestRoomsHandler_UpdateRoomWithPing(t *testing.T) {
 		require.NoError(t, err)
 
 		for _, invalidStateRawRequest := range invalidStateRawRequests {
-
-			roomStorageMock.EXPECT().GetRoom(gomock.Any(), gomock.Any(), gomock.Any()).Return(gameRoom, nil)
+			// TODO(gabrielcorado): since we're exposing the room manager
```
GuilhermBrSp, line 195: @gabrielcorado I didn't understand this TODO here, could you help me understand?

internal/core/services/room_manager/room_manager.go
```diff
@@ -101,6 +101,7 @@ func (m *RoomManager) CreateRoom(ctx context.Context, scheduler entities.Schedul
 	return room, instance, err
 }
 
+// TODO(gabrielcorado): should we "force" the room status to be "Terminating"?
```
GuilhermBrSp, line 104: @gabrielcorado  In this case, the runtime watcher would already update its status to terminating if the deletion succeeds, right? in this case I don't think we should be forcing this status here

GuilhermBrSp (approved):

internal/core/services/room_manager/status.go
```diff
@@ -77,6 +135,40 @@ func (m *RoomManager) validateRoomStatusTransition(currentStatus game_room.GameR
 	return nil
 }
 
+// updateGameRoomStatus updates the game based on the ping and runtime status.
+func (m *RoomManager) updateGameRoomStatus(ctx context.Context, schedulerId, gameRoomId string) error {
```
GuilhermBrSp, line 139: @gabrielcorado could we have a specific test for this method? i noticed that we are not covering those first four if's in room_manager tests

internal/api/handlers/rooms_handler_test.go
gabrielcorado: We have to implement a `DoAndReturn`, which for me seems like a "test implementation" for this test. The TODO is not clear what has to be done here, and I'm going to update its description.

internal/core/services/room_manager/room_manager.go
gabrielcorado: > In this case, the runtime watcher would already update its status to terminating if the deletion succeeds, right?
Yes. The only scenario where this is not true is when the runtime watcher is not running.

internal/core/services/room_manager/status.go
gabrielcorado: Sure. Some use cases are being tested on the `UpdateRoom` test, but I'll add tests for this function.

PR #216, opened 2021-10-05
Feature/support rooms forwarder on ping api

GuilhermBrSp:
## What ❓ 
This PR adds event forwarding on room manager's `UpdateRoom` method (ping), including the first version of EventsForwarder interface and a dummy implementation of MatchmakerForwarder.


internal/core/ports/events_forwarder.go
```diff
@@ -0,0 +1,39 @@
...
+
+	"github.com/topfreegames/maestro/internal/core/entities/game_room"
+)
+
+type EventsForwarder interface {
+	ForwardRoomEvent(ctx context.Context, room *game_room.GameRoom, status string, eventType string, metadata map[string]interface{}) (res *Response, err error)
+}
+
+// Response is the struct that defines a forwarder response
+type Response struct {
```
gabrielcorado, line 36: WDYT about having a more specific struct name? `Response` could be misleading. What about `EventsForwarderResponse`?

internal/adapters/events_forwarder/matchmaker_forwarder/matchmaker_forwarder.go
```diff
@@ -0,0 +1,40 @@
...
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+package matchmaker_forwarder
```
gabrielcorado, line 23: If you're thinking about matchmaker here, we should call it `grpcForwarder` instead. But this forwarder could be called something like a `NoopForwarder` because it won't do anything, only return, WDYT?

internal/core/ports/events_forwarder.go
```diff
@@ -0,0 +1,39 @@
...
+package ports
+
+import (
+	"context"
+
+	"github.com/topfreegames/maestro/internal/core/entities/game_room"
+)
+
+type EventsForwarder interface {
+	ForwardRoomEvent(ctx context.Context, room *game_room.GameRoom, status string, eventType string, metadata map[string]interface{}) (res *Response, err error)
```
gabrielcorado, line 32: Returning just `[]byte, error` should be sufficient here? I'm thinking that introducing a `response.Code` might overexpose the `EventForwarders` since it should be responsible for telling if the function returned an error or not.

PR #232, opened 2021-11-09
Adding ping metadata value on API that list game rooms

lyracampos:
The purpose of this PR is add **Pint Metadata Value** from each game room listed on API

- added property **LastPingMetada**  on **Room** struct
- changes methods that **Details** and **List** Game Rooms to return new property on API
- tests adjustments to attend new implementation

### DIFF
--- a/api/room_handler_test.go
+++ b/api/room_handler_test.go
@@ -1,5 +1,7 @@
 // maestro
+//go:build unit
 // +build unit
+
 // https://github.com/topfreegames/maestro
 //
 // Licensed under the MIT license:
@@ -1284,16 +1286,19 @@ forwarders:
 					Return(redis.NewStringStringMapResult(map[string]string{
 						"status":   "ready",
 						"lastPing": "1632405900",
+						"metadata": `{"region":"us","type":"sometype"}`,
 					}, nil))
 				mockRedisClient.EXPECT().HGetAll("scheduler:scheduler-name:rooms:test-ready-2").
 					Return(redis.NewStringStringMapResult(map[string]string{
 						"status":   "ready",
 						"lastPing": "1632405900",
+						"metadata": `{"region":"us","type":"sometype"}`,
 					}, nil))
 				mockRedisClient.EXPECT().HGetAll("scheduler:scheduler-name:rooms:test-ready-3").
 					Return(redis.NewStringStringMapResult(map[string]string{
 						"status":   "ready",
 						"lastPing": "1632405900",
+						"metadata": `{"region":"us","type":"sometype"}`,
 					}, nil))
 
 				mockRedisClient.EXPECT().HGet("scheduler:scheduler-name:podMap", "test-ready-3").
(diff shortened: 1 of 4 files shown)


models/room.go
```diff
@@ -557,12 +558,19 @@ func GetRoomDetails(redisClient interfaces.RedisClient, schedulerName string, ro
 
 	lastPingAt, _ := strconv.ParseInt(roomData["lastPing"], 10, 64)
 	status := roomData["status"]
+	var lastPingMetadata map[string]interface{}
+	err = json.Unmarshal([]byte(roomData["metadata"]), &lastPingMetadata)
```
henriqueoelze, line 562: nit: WDYT about just merge this line with the If? Just to keep it easy to read, like: 
if err := json.Unmarshal([]byte(roomData["metadata"]), &lastPingMetadata); err != nil {
}

api/room_handler_test.go
```diff
@@ -1585,6 +1603,7 @@ forwarders:
 				Return(redis.NewStringStringMapResult(map[string]string{
 					"status":   "ready",
 					"lastPing": "1632405900",
+					"metadata": `{"region":"us","type":"sometype"}`,
```
henriqueoelze, line 1606: We added the metadata in all mocks, but we had a problem in the past during the serialization of the data into API. Is it possible to check if the metadata will be "correct" after we call the API?

Also, WDYT about add a scenario where metadata is just en empty {} (more commons scenario in prod)

api/room_handler_test.go
```diff
@@ -1,5 +1,7 @@
 // maestro
+//go:build unit
```
GuilhermBrSp, line 2: @lyracampos I think this was auto-added, right? you can remove it

henriqueoelze (commented):
Just two small comments. Besides those, great work folks! 🚀

models/room_test.go
```diff
@@ -1,5 +1,7 @@
 // maestro
+//go:build unit
```
GuilhermBrSp, line 2: @lyracampos same thing here

GuilhermBrSp (approved):

models/room.go
```diff
@@ -557,12 +558,19 @@ func GetRoomDetails(redisClient interfaces.RedisClient, schedulerName string, ro
 
 	lastPingAt, _ := strconv.ParseInt(roomData["lastPing"], 10, 64)
 	status := roomData["status"]
+	var lastPingMetadata map[string]interface{}
```
GuilhermBrSp, line 561: @lyracampos  i think we can add a test case here for when the "metadata" key does not exist on roomData, this is a scenario that will happen when the incoming ping does not have metadata when persisting it (you can check line 279 in this file). For this scenario, I don't think this function should return an error in this scenario, it would be best for it to return empty metadata, WDYT?.

GuilhermBrSp (changes requested):
We have to check the logic of missing metadata key

GuilhermBrSp (approved):

caiordjesus (approved):
LGTM!

PR #234, opened 2021-11-17
Next feat/get scheduler

caiordjesus:
### What was done?
- Implemented getScheduler API. It receives an optional parameter (version). Also, integration tests for it.

### Why was done?
- It wasn't possible to get informations about an specific scheduler, neither an specific version of it.

### How was done?
- Decided for expanding to a new function "GetSchedulerByVersion" that receives the version parameter. If version is not passed, it calls the "GetScheduler" function, which returns the latest version of the scheduler.

To be discussed: the actual scheduler message does not brings specific informations, like yaml and rollback version. Maybe we should add new properties to the message?


internal/adapters/scheduler_storage/pg/pg.go
```diff
@@ -52,7 +52,16 @@ SELECT
 	v.rolling_update_status, v.rollback_version
 FROM schedulers s join scheduler_versions v
 	ON s.name=v.name AND v.version=s.version
-WHERE s.name = ?`
+WHERE s.name = ?
+order by created_at desc`
+	queryGetSchedulerByVersion = `
+SELECT
```
lyracampos, line 58: @caiordjesus queryGetScheduler and queryGetSchedulerByVersion are identical queries, in but queryGetSchedulerByVersion hasn't order by by created_at. I think it was because the query QueryGetSchedulerByVersion should returns one row, right? if not i think you can use order by like in first query.

proto/api/v1/messages.proto
```diff
@@ -35,6 +96,28 @@ message Scheduler {
   google.protobuf.Timestamp created_at = 6;
   // Max surge of rooms 
   string max_surge = 7;
+  // GameRoom spec
+  Spec spec = 8;
+}
+
+// Scheduler message used in the "ListScheduler version" definition. The "spec" is not implemented
+// on this message since it's unecessary for the list function
+message ListedScheduler {
```
lyracampos, line 105: Appearlly this message is a structure to use like a list and not a list by itself.

normally the message are create to define a data structure and should not be associate with and action (verb), beside that is not a good practice to conjugate verb to to express an action. When you use like that you are breaking the semantics definitions. 

I think you could rename this message to be more expressive but note associate with an action. What do you think?

GuilhermBrSp (approved):

PR #235, opened 2021-11-18
Feature/scheduler update endpoint

lyracampos:
### what was done?

* Created new endpoint to update a scheduler on  **Management-Api**. this update doesn't work synchronously. It just enqueue a operation to be executed by **Operation Worker**

### why was done?
* Nowadays Maestro Next doesn't have feature to Update Schedule on Management-API. 

### what was done?
* a new method to expose new endpoint PUT - api/v1/schedulers/scheduler_name
* a new method on domain context to scheduler update flow orchestrate
* it was necessary to implement an interface to abstract UpdateSchedulerConfig from scheduler_manager to avoid circular references between services/scheduler_manager and operations/update_scheduler.


cmd/rooms_api/wire_gen.go
```diff
@@ -1,7 +1,8 @@
 // Code generated by Wire. DO NOT EDIT.
 
 //go:generate go run github.com/google/wire/cmd/wire
-//+build !wireinject
+//go:build !wireinject
```
caiordjesus, line 4: Every time we run "make generate" this is created, but do we actually want to merge this? What does it mean?

e2e/suites/management/utils.go
```diff
@@ -43,7 +43,7 @@ func createSchedulerAndWaitForIt(
 	kubeclient kubernetes.Interface,
 	gruCommand []string) (string, error) {
 	schedulerName := framework.GenerateSchedulerName()
-	createRequest := &maestroApiV1.CreateSchedulerRequest{
+	createRequest := &maestroApiV1.SchedulerRequest{
```
caiordjesus, line 46: Why this was changed? I think this MIGHT break all e2e tests, no?

internal/core/services/scheduler_manager/scheduler_manager.go
```diff
@@ -154,6 +155,26 @@ func (s *SchedulerManager) UpdateSchedulerConfig(ctx context.Context, scheduler
...
+	if err != nil {
+		return nil, err
+	}
+
+	_, err = s.schedulerStorage.GetScheduler(ctx, scheduler.Name)
+	if err != nil {
+		return nil, fmt.Errorf("no scheduler found to be updated: %w", err)
+	}
+
+	scheduler.State = entities.StateCreating //TODO (guilhermelyra): what correct value should be set for state in this case?
```
caiordjesus, line 169: As we discussed before, I think we should get together and review the status and maybe create a machine state. Maybe a status like "updating" would make sense?

caiordjesus (commented):
- I have a concern about the existent statuses (as I commented, maybe we should review then);
- I have a concern about the change on the "e2e/suites/management/utils.go" change on function. Is it related to the update endpoint?

e2e/suites/management/utils.go
lyracampos: > Why this changed? I think this MIGHT break all e2e tests, no?

good! I had renamed this message to make them more generic, but made rollback and forgot to change this point.

lyracampos:
> * I have a concern about the existent statuses (as I commented, maybe we should review then);
> * I have a concern about the change on the "e2e/suites/management/utils.go" change on function. Is it related to the update endpoint?

lyracampos:
> * I have a concern about the existent statuses (as I commented, maybe we should review then);
> * I have a concern about the change on the "e2e/suites/management/utils.go" change on function. Is it related to the update endpoint?



> * I have a concern about the existent statuses (as I commented, maybe we should review then);
> * I have a concern about the change on the "e2e/suites/management/utils.go" change on function. Is it related to the update endpoint?

1- I agree @caiordjesus. I think we could hold on MR while we review this point. 
2- You're right @caiordjesus! I made rollback could disregard

lyracampos:
> * I have a concern about the existent statuses (as I commented, maybe we should review then);
> * I have a concern about the change on the "e2e/suites/management/utils.go" change on function. Is it related to the update endpoint?

I agree @caiordjesus. I think we could hold on MR while we review this point.
You're right @caiordjesus! I made rollback in "e2e../utils.go" could disregard

internal/core/operations/update_scheduler/update_scheduler_executor.go
```diff
@@ -32,16 +32,19 @@ import (
 	"github.com/topfreegames/maestro/internal/core/entities/operation"
 	"github.com/topfreegames/maestro/internal/core/operations"
 	"github.com/topfreegames/maestro/internal/core/services/room_manager"
-	"github.com/topfreegames/maestro/internal/core/services/scheduler_manager"
 	"go.uber.org/zap"
 )
 
+type SchedulerManager interface {
```
GuilhermBrSp, line 38: @lyracampos why did you create this interface?  regardless of the reason, I don't see this file being the proper place to define I, right?

lyracampos: @GuilhermBrSp,It was necessary to create this interface to avoid circular dependency/reference between update scheduler and scheduler manager.

yes, i agree that this interface can be wrote in a specific file. I'm not sure about where because we don't have this kind of implementation yet. 
I note that interfaces are placed in 'internal/core/ports' but in my point of view this new interface aren't a port interface concept.

I saw in some github projects open source that devs put interface into same folder that implementation ex: update_scheduler.go and update_scheduler_impl.go but for me it smeels like java implementations hahaha.

my suggestion here is create new folder in /internal/core/services/interfaces to put interfaces that abstract services implementations. What do you think?

GuilhermBrSp: @lyracampos  I got the circular dependency problem, i strongly agree on creating domain service interfaces for two reasons:
- Solve this type of circular dependency problem
- Those interfaces will enable us to create mocks for these services and use them in handlers tests. Currently, in our handler tests, we have to mock all dependencies of the services that it uses, generating a level of mocking that shouldn't exist in those scenarios.

Regarding where we should declare those interfaces, I like your suggestion to place it inside `/internal/core/services/interfaces`

internal/core/services/scheduler_manager/scheduler_manager.go
lyracampos: @caiordjesus we have decided that state property from scheduler doesn't aggreegate value today. for now we are going to follow setting this value only with creating. Soon we'll see with this property will be removed or not.

GuilhermBrSp (approved):
@lyracampos i think the only change that is missing is the removal of `version` field in request

PR #236, opened 2021-11-23
using original value for new generated version to ensure that format …

lyracampos:
This merge request is only to ensure that value version continue in original format that it was when scheduler was created.


caiordjesus (approved):
Awesome!
Understood based on: https://pkg.go.dev/github.com/hashicorp/go-version

lyracampos:
> Awesome!
> Understood based on: https://pkg.go.dev/github.com/hashicorp/go-version

thanks for sharing the documentation <3

PR #237, opened 2021-11-24
feature: getSchedulerVersions endpoint created.

caiordjesus:
### What was done?
- Created GetSchedulerVersions endpoint and tests.
### Why was done?
- We didn't have a way of fetching from the client side the older versions of an scheduler deployed on maestro.
### How was done?
- Created "GET /schedulers/{scheduler_name}/versions" on management-api


internal/api/handlers/schedulers_handler.go
```diff
@@ -82,6 +82,21 @@ func (h *SchedulersHandler) GetScheduler(ctx context.Context, request *api.GetSc
 	return &api.GetSchedulerResponse{Scheduler: h.fromEntitySchedulerToResponse(scheduler)}, nil
 }
 
+func (h *SchedulersHandler) GetSchedulerVersions(ctx context.Context, request *api.GetSchedulerVersionsRequest) (*api.GetSchedulerVersionsResponse, error) {
+	var versions []*entities.SchedulerVersion
```
GuilhermBrSp, line 86: @caiordjesus Couldn't we declare and assign those vars using `versions, err :=` instead of making it separately?

internal/core/services/scheduler_manager/scheduler_manager.go
```diff
@@ -88,6 +88,10 @@ func (s *SchedulerManager) GetScheduler(ctx context.Context, schedulerName, vers
 	})
 }
 
+func (s *SchedulerManager) GetSchedulerVersions(ctx context.Context, schedulerName string) ([]*entities.SchedulerVersion, error) {
```
GuilhermBrSp, line 91: @caiordjesus i think the test for this new guy is missing

caiordjesus: Actually the test for GetScheduler was missing too!
Just added on the last commit. Can you please take a look again?

internal/api/handlers/schedulers_handler.go
caiordjesus: You're absolutely right!

GuilhermBrSp (approved):

lyracampos (approved):

PR #238, opened 2021-11-30
Fix concurrency problem in update scheduler executor

GuilhermBrSp:
## What 👀 
Adds a mechanism in update rooms executor to prevent rooms that ate already being replaced to be replaced more than one time. Also implements a TODO that existed in DeleteRoom method in RoonManager.

## Why ❓ 
Upon implementing the e2e test for the update scheduler, we found that the update scheduler executor is trying to replace the same room more than one time, which leads to a higher number of game rooms created after the operation finishes. This is not the behavior that we expect for the update, we expect each room to be replaced one time only, that means, if we have a current scheduler with 4 rooms, after a major version update, we expect the scheduler to have 4 rooms (with the new version) after the operation finishes.

## How 0️⃣ 1️⃣  
- Using a sync map to manage rooms that are being replaced and prevent them from being replaced again.
- Update DeleteRoom method to wait for the status of the room to be Terminating.
- Updated tests and add some new test cases.

### DIFF
--- a/internal/core/operations/update_scheduler/update_scheduler_executor.go
+++ b/internal/core/operations/update_scheduler/update_scheduler_executor.go
@@ -37,8 +37,9 @@ import (
 )
 
 type UpdateSchedulerExecutor struct {
-	roomManager      *room_manager.RoomManager
-	schedulerManager interfaces.SchedulerManager
+	roomManager        *room_manager.RoomManager
+	schedulerManager   interfaces.SchedulerManager
+	roomsBeingReplaced sync.Map
 }
 
 func NewExecutor(roomManager *room_manager.RoomManager, schedulerManager interfaces.SchedulerManager) *UpdateSchedulerExecutor {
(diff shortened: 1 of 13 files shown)


internal/core/operations/update_scheduler/update_scheduler_executor.go
```diff
@@ -37,8 +37,9 @@ import (
 )
 
 type UpdateSchedulerExecutor struct {
-	roomManager      *room_manager.RoomManager
-	schedulerManager interfaces.SchedulerManager
+	roomManager        *room_manager.RoomManager
+	schedulerManager   interfaces.SchedulerManager
+	roomsBeingReplaced sync.Map
```
GuilhermBrSp, line 42: In my first implementation, i was using a build-in map with a Mutex to manage concurrency, but then I discovered this awesome map that simplifies the code and can be used for what we are doing here.

lyracampos (approved):

caiordjesus (approved):

PR #244, opened 2021-12-17
implements GrantLease method for operationLeaseStorage port

caiordjesus:
### What was done?
- Implemented method GrantLease for operation_lease_storage port and unit tests for it.
### Why?
- To be able to understand which operations are running, and which expired, we first grant a lease for the specific operation.
### How?
- Using redis sorted set. We add a score (unixtime) with the information of when the operation expires.

### DIFF
--- a/internal/adapters/operation_lease/redis/redis.go
+++ b/internal/adapters/operation_lease/redis/redis.go
@@ -97,7 +97,7 @@ func (r *redisOperationLeaseStorage) existsOperationLease(ctx context.Context, s
 		if err == redis.Nil {
 			return false, nil
 		}
-		return false, errors.NewErrUnexpected("failed to list active operations for \"%s\"", schedulerName).WithError(err)
+		return false, errors.NewErrUnexpected("failed on finding operationsLease for \"%s\" and operationID \"%s\"", schedulerName, operationId).WithError(err)
 	}
 	return true, nil
 }
--- a/internal/adapters/operation_lease/redis/redis.go
+++ b/internal/adapters/operation_lease/redis/redis.go
@@ -92,9 +92,12 @@ func (r *redisOperationLeaseStorage) buildSchedulerOperationLeaseKey(schedulerNa
 }
 
 func (r *redisOperationLeaseStorage) existsOperationLease(ctx context.Context, schedulerName, operationId string) (bool, error) {
-	operationLeaseList, _, err := r.client.ZScan(ctx, r.buildSchedulerOperationLeaseKey(schedulerName), 0, operationId, 0).Result()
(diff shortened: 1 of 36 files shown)


internal/adapters/operation_lease/redis/redis.go
```diff
@@ -64,3 +86,15 @@ func (r *redisOperationLeaseStorage) FetchLeaseTTL(ctx context.Context, schedule
 func (r *redisOperationLeaseStorage) ListExpiredLeases(ctx context.Context, schedulerName string, maxLease time.Time) ([]operation.OperationLease, error) {
 	return nil, nil
 }
+
+func (r *redisOperationLeaseStorage) buildSchedulerOperationLeaseKey(schedulerName string) string {
+	return fmt.Sprintf("operations:%s:operationsLease", schedulerName)
+}
+
+func (r *redisOperationLeaseStorage) existsOperationLease(ctx context.Context, schedulerName, operationId string) (bool, error) {
+	operationLeaseList, _, err := r.client.ZScan(ctx, r.buildSchedulerOperationLeaseKey(schedulerName), 0, operationId, 0).Result()
```
GuilhermBrSp, line 95: @caiordjesus WDYT about using ZSCORE for checking the existence here? it's time complexity is constant in comparison with Zscan that have a linear complexity in worst case. Ref: https://redis.io/commands/zscore

GuilhermBrSp:
```
operationLeaseScore, err := client.ZScore(ctx, r.buildSchedulerOperationLeaseKey(schedulerName), operationID).Result()
```

GuilhermBrSp: se o erro for diferente de nil, significa que o valor que estamos buscando não existe ainda, então você poderia retornar apenas err != nil

internal/adapters/operation_lease/redis/redis.go
```diff
@@ -64,3 +86,15 @@ func (r *redisOperationLeaseStorage) FetchLeaseTTL(ctx context.Context, schedule
...
+	if err != nil {
+		return false, errors.NewErrUnexpected("failed to list active operations for \"%s\"", schedulerName).WithError(err)
```
GuilhermBrSp, line 97: This message should inform about operation lease and not active operations, right?

internal/adapters/operation_lease/redis/redis.go
caiordjesus: Awesome insight!!
Just made the change!

internal/adapters/operation_lease/redis/redis.go
caiordjesus: You are right! Just made the change! Thanks!!

GuilhermBrSp (approved):

PR #245, opened 2021-12-21
implements RevokeLease method on OperationLeaseStorage adapter

caiordjesus:
### What?
- Implements RevokeLease method on OperationLeaseStorage adapter.
### Why?
- After a lease is granted to an operation of a scheduler, we need a method to revoke it (we won't be updating it forever, neither letting it expired, since expired means that the operation might be stuck)
### How?
- The method verifies if the operationLease exists, and if so, removes it from the redis zset.

### DIFF
--- a/internal/adapters/operation_lease/redis/redis.go
+++ b/internal/adapters/operation_lease/redis/redis.go
@@ -78,10 +78,10 @@ func (r *redisOperationLeaseStorage) RevokeLease(ctx context.Context, schedulerN
 	}
 
 	if !existsLease {
-		return errors.NewErrNotFound("Lease of scheduler \"%s\" and operationId \"%s\" trying to be revoked does not exist", schedulerName, operationID)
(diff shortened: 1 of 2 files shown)


internal/adapters/operation_lease/redis/redis.go
```diff
@@ -72,6 +72,19 @@ func (r *redisOperationLeaseStorage) GrantLease(ctx context.Context, schedulerNa
...
+	existsLease, err := r.existsOperationLease(ctx, schedulerName, operationID)
+	if err != nil {
+		return err
+	}
+
+	if !existsLease {
+		return errors.NewErrNotFound("Lease of scheduler \"%s\" and operationId \"%s\" does not exist", schedulerName, operationID)
+	}
+
+	_, err = r.client.ZRem(ctx, r.buildSchedulerOperationLeaseKey(schedulerName), operationID).Result()
```
GuilhermBrSp, line 84: just for curiosity, what does ZRem returns as the first argument?

GuilhermBrSp (approved):

lyracampos (approved):

PR #247, opened 2021-12-27
Fix project typos

GuilhermBrSp:
## What ❓ 
Fix project typos

## Why 🤔 
Currently, there are some typos in the project, including comments, methods names, var names. We need to maintain the housekeeping of the project by fixing those. 

## How 👀 
Using Goland features to detect project typos and fix them.


internal/core/operations/remove_rooms/remove_rooms_executor_test.go
```diff
@@ -34,14 +34,14 @@ import (
...
+	clockmock "github.com/topfreegames/maestro/internal/adapters/clock/mock"
+	eventsforwardermock "github.com/topfreegames/maestro/internal/adapters/events_forwarder/mock"
+	instancestoragemock "github.com/topfreegames/maestro/internal/adapters/instance_storage/mock"
+	roomstoragemock "github.com/topfreegames/maestro/internal/adapters/room_storage/mock"
 
-	port_allocator_mock "github.com/topfreegames/maestro/internal/adapters/port_allocator/mock"
+	portallocatormock "github.com/topfreegames/maestro/internal/adapters/port_allocator/mock"
 	"github.com/topfreegames/maestro/internal/adapters/room_storage/mock"
-	runtime_mock "github.com/topfreegames/maestro/internal/adapters/runtime/mock"
+	runtimemock "github.com/topfreegames/maestro/internal/adapters/runtime/mock"
```
GuilhermBrSp, line 44: @caiordjesus @lyracampos In Maestro's Next we are using three types of naming regarding import alias, with `snake_case`, `camelCase` and `packagenaming`. After searching a bit it seems that the correct convention to follow is using the same convention as package naming: https://youtrack.jetbrains.com/issue/GO-5942.
We can even check kubernetes project adopting this convention: https://github.com/kubernetes/kubernetes/pull/35258/files.

What do you guys think about adopting this import alias naming convention?

lyracampos: I think that is the best way

lyracampos (approved):

PR #248, opened 2021-12-28
Update golang version to 1.17

GuilhermBrSp:
What ❓
Update golang version in go.mod

Why 🤔
Since we are creating maestro next from scratch, we decided to use the latest LTS go version in the project.

How 👀
Updating go version in go mod file


e2e/go.mod
```diff
@@ -14,4 +14,70 @@ require (
 	k8s.io/client-go v0.22.1
 )
 
+require (
```
GuilhermBrSp, line 17: In golang v 1.17 the module graph has been changed to enable pruning and lazy loading. The second require block contains indirect dependencies. Ref: https://stackoverflow.com/questions/69432877/why-there-are-two-require-blocks-in-go-mod-since-go-1-17, https://go.dev/doc/go1.17#go-command

lyracampos (approved):

caiordjesus (approved):
LGTM!

PR #249, opened 2021-12-28
Always build maestro components images while running e2e tests

GuilhermBrSp:
What ❓
Configure e2e tests execution to always build maestro components images.

Why 🤔
Currently, when executing e2e tests via makefile, images built previously (rooms API, worker, etc) aren't built again. This can lead up to problems when changing the code since the changes will not be "loaded" because the image already exists and it will not build again. Currently, we have to delete images manually so it builds it again with new changes. We need to guarantee that the image used in e2e tests will be built while running it, preventing these problems to happen.

How 👀
Running docker commands with --build flag for maestro components.

### DIFF
--- a/e2e/framework/maestro/components/management_api.go
+++ b/e2e/framework/maestro/components/management_api.go
@@ -46,7 +46,7 @@ func ProvideManagementApi(maestroPath string) (*ManagementApiServer, error) {
 	identifier := strings.ToLower("test-something")
 
 	compose := tc.NewLocalDockerCompose(composeFilePaths, identifier)
-	composeErr := compose.WithCommand([]string{"up", "-d", "management-api"}).Invoke()
+	composeErr := compose.WithCommand([]string{"up", "-d", "--build", "management-api"}).Invoke()
 
 	if composeErr.Error != nil {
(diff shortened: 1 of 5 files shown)


e2e/framework/maestro/components/management_api.go
```diff
@@ -46,7 +46,7 @@ func ProvideManagementApi(maestroPath string) (*ManagementApiServer, error) {
 	identifier := strings.ToLower("test-something")
 
 	compose := tc.NewLocalDockerCompose(composeFilePaths, identifier)
-	composeErr := compose.WithCommand([]string{"up", "-d", "management-api"}).Invoke()
+	composeErr := compose.WithCommand([]string{"up", "-d", "--build", "management-api"}).Invoke()
```
GuilhermBrSp, line 49: @caiordjesus @lyracampos I decided to let the image building here instead of placing it in the makefile, this way we ensure the image building even when running e2e tests in isolation (without makefile)

caiordjesus: Nice!!

caiordjesus (approved):
LGTM!

lyracampos (approved):

PR #250, opened 2021-12-28
migrate buf version from v1beta1 to v1

caiordjesus:
### What was done?
Upgrade [buf](https://docs.buf.build/introduction) version from v1beta1 to v1.
### Why was done?
To use a stable version on the project, we want to leave beta version behind. Also, the v1 comes with a lot of improvements that might help the future of the project.
### How was done?
Using the [migration guide](https://docs.buf.build/configuration/v1beta1-migration-guide).
### Next steps
A natural next step would be pushing maestro api v1 to [BSR](https://docs.buf.build/bsr/introduction). This would bring improvements such as documented on the buf page.
Example: https://buf.build/caio-rodrigues-22/petapis/docs/main/pet.v1


caiordjesus:
Actually Buf does not have an official 1.0 release (only release candidates). Would be prudent to wait till the official release.
https://github.com/bufbuild/buf/releases?page=1

lyracampos:
@caiordjesus a doubt,I saw that it has enough TODO declarations. When we are going work with them?

lyracampos (approved):

caiordjesus:
> @caiordjesus a doubt,I saw that it has enough TODO declarations. When we are going work with them?

They depend on using the BSR. Since BSR is very new, we could think about using it somewhere in the future if we think it's safe to do so.

PR #251, opened 2021-12-30
bump mock. v1.5 => v1.6

caiordjesus:


internal/adapters/runtime/kubernetes/game_room_watcher_test.go
```diff
@@ -134,7 +134,7 @@ func TestGameRoomsWatch(t *testing.T) {
 			}
 
 			return false
-		}, time.Minute, time.Second)
```
GuilhermBrSp, line 137: @caiordjesus Why do we have to increase this number? does the new mockgen version force us?

caiordjesus: Nope, it does not, but sometimes my tests were not passing because of it. Even here in the github workflow. That's why I increased it...

GuilhermBrSp (approved):

PR #252, opened 2022-01-03
Implement ListExpiredLeases method including unitary tests

GuilhermBrSp:
## What❓
Implement ListExpiredLeases method logic to retrieve leases that have expired. 

## Why 🤔 
Following the Operations lease feature plan, we need to provide some method in the repository that can retrieve operations lease that have expired (using a maxTtl argument as reference)

## How 👀 
Using [ZRANGEBYSCORE](https://redis.io/commands/zrangebyscore) to retrieve the list of leases whose ttl (score) resides in the range [-inf,  maxTtl]


caiordjesus (approved):
LGTM!

lyracampos (approved):

PR #254, opened 2022-01-03
Adds Operation Manager lease methods

caiordjesus:
### What?
Adds operation Manager lease methods: **GrantLease**, **RevokeLease** and **StartLeaseRenewGoRoutine**
### Why?
The worker should not communicate directly with redis-manipulation methods (repositories). The operation manager is responsible for this work. That said, we needed methods implemented on the manager being called by the worker.
### How?
GrantLease and RevokeLease are straightforward. The StartLeaseRenewGoRoutine calls a goroutine that will keep tracking the operation lease, renewing it from time to time before cancelation of finished status.

### DIFF
--- a/internal/core/services/operation_manager/operation_manager.go
+++ b/internal/core/services/operation_manager/operation_manager.go
@@ -92,15 +92,17 @@ type OperationManager struct {
 	operationCancelFunctions        *OperationCancelFunctions
 	flow                            ports.OperationFlow
 	storage                         ports.OperationStorage
+	leaseStorage                    ports.OperationLeaseStorage
 	operationDefinitionConstructors map[string]operations.DefinitionConstructor
 }
 
-func New(flow ports.OperationFlow, storage ports.OperationStorage, operationDefinitionConstructors map[string]operations.DefinitionConstructor) *OperationManager {
+func New(flow ports.OperationFlow, storage ports.OperationStorage, operationDefinitionConstructors map[string]operations.DefinitionConstructor, leaseStorage ports.OperationLeaseStorage) *OperationManager {
 	return &OperationManager{
 		flow:                            flow,
 		storage:                         storage,
 		operationDefinitionConstructors: operationDefinitionConstructors,
 		operationCancelFunctions:        NewOperationCancelFunctions(),
+		leaseStorage:                    leaseStorage,
 	}
 }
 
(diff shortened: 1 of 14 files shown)


internal/core/services/operation_manager/operation_manager_test.go
```diff
@@ -485,3 +500,235 @@ func TestWatchOperationCancellationRequests(t *testing.T) {
...
+
+		operationLeaseStorage.EXPECT().RevokeLease(ctx, schedulerName, operationID).Return(errors.New("error"))
+		err := opManager.RevokeLease(ctx, op)
+		require.Error(t, err)
+		require.Equal(t, err.Error(), "failed to revoke lease to operation: error")
+	})
+}
+
+func TestStartLeaseRenewGoRoutine(t *testing.T) {
+	t.Run("Renews lease not executed since op.status = finished", func(t *testing.T) {
```
GuilhermBrSp, line 618: @caiordjesus What do you think about having a test covering the scenario in which the operation status updates to `finished` after the renew goroutine starts? This scenario here already initializes the goroutine with the finalized status before the goroutine starts.

internal/core/services/operation_manager/operation_manager.go
```diff
@@ -271,3 +273,61 @@ func (om *OperationManager) cancelOperation(ctx context.Context, schedulerName,
...
+func (om *OperationManager) RevokeLease(ctx context.Context, operation *operation.Operation) error {
+	err := om.leaseStorage.RevokeLease(ctx, operation.SchedulerName, operation.ID)
+	if err != nil {
+		return fmt.Errorf("failed to revoke lease to operation: %w", err)
+	}
+
+	return nil
+}
+
+func (om *OperationManager) StartLeaseRenewGoRoutine(operationCtx context.Context, op *operation.Operation, ttl time.Duration) {
```
GuilhermBrSp, line 295: @caiordjesus currently we are receiving the ttl in both renew lease and grant lease methods as an argument. This means that the worker will have the responsibility of knowing the value it will pass to those methods. I was thinking that the best approach for setting this value would be isolating it into a config level parameter (same as the room initialization timeout on `internal/core/services/room_manager/room_manager.go:108`) instead of creating a constant value or just a hardcoded one.  With this in view, it would be necessary to include a new dependency of a config property similar to what we do in `RoomManager`, but with this ttl value being passed as an argument this config dependency would be injected in the `worker` and not the `operation manager`. Since we are already defined that the operation manager being responsible for encapsulating this kind of logic, what do you think about having it in there (doesn't need to be in this PR)?

internal/core/services/operation_manager/operation_manager_test.go
caiordjesus: Good idea! Just did it!

GuilhermBrSp (approved):

lyracampos (approved):

PR #256, opened 2022-01-04
Include lease data in operation listing

GuilhermBrSp:
## What ❓ 
Add Lease field in Operation struct and modify ListSchedulerActiveOperations method from OperationManager to include lease information in its response.

## Why 🤔 
Get operations endpoint will now include lease information, so we need to change its internals to include this lease data.

## How 👀 
- Minor refactor in operation_manager file
- New field in Operation struct
- Update ListSchedulerActiveOperations method and improve its tst

### DIFF
--- a/internal/core/services/operation_manager/operation_cancel_functions.go
+++ b/internal/core/services/operation_manager/operation_cancel_functions.go
@@ -1,3 +1,25 @@
+// MIT License
+//
+// Copyright (c) 2021 TFG Co
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in all
+// copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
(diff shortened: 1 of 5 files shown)


internal/core/services/operation_manager/operation_cancel_functions.go
```diff
@@ -0,0 +1,59 @@
+package operation_manager
+
+import (
```
GuilhermBrSp, line 3: This entire file was just a extraction refactor, no changes here

internal/core/services/operation_manager/operation_manager_test.go
```diff
@@ -415,8 +477,8 @@ func TestListSchedulerFinishedOperations(t *testing.T) {
 		}
 
 		schedulerName := "test-scheduler"
-		operationStorage.EXPECT().ListSchedulerActiveOperations(ctx, schedulerName).Return(operationsResult, nil)
-		operations, err := opManager.ListSchedulerActiveOperations(ctx, schedulerName)
+		operationStorage.EXPECT().ListSchedulerFinishedOperations(ctx, schedulerName).Return(operationsResult, nil)
```
GuilhermBrSp, line 480: this test was testing the wrong method

internal/core/services/operation_manager/operation_cancel_functions.go
caiordjesus: Nice!!

internal/core/services/operation_manager/operation_manager.go
```diff
@@ -331,3 +263,37 @@ func (om *OperationManager) StartLeaseRenewGoRoutine(operationCtx context.Contex
 		}
 	}()
 }
+
+func (om *OperationManager) addOperationsLeaseData(ctx context.Context, schedulerName string, ops []*operation.Operation) error {
+	for _, op := range ops {
+		ttl, err := om.leaseStorage.FetchLeaseTTL(ctx, schedulerName, op.ID)
```
caiordjesus, line 269: Something that crossed my mind is that what we want for the API is to list all operations.
So we go to redis once, get all operations, then go back to the server and sort it in lists.
What we added makes us go to redis n times (being n the number of operations on redis). Since we have to list all operations anyways, do you think it would be bad to have a method that get all operation leases, so that we can iterate through them changing those 1+n redis queries to only 2 (all operations, then all leases)?

GuilhermBrSp: It wouldn't be bad if we had this method to return all operation leases, but I question if this is an actual necessity for the current model. You can see that this method iterates over all active operations **for a given scheduler**, and in the current model (one worker model) this number will always be 1, so I'm not sure if we need this optimization now. Even with the multiple workers model, the active status is always a temporary status and not a final one (like StatusFinished for example), this means that this number does not scale as the finished operations do. Anyway, we are still querying all operations (finished, pending and active) in the storage methods, since we are using pagination, it would be nice to use some kind of "LIMIT" at least in the finished operations query (since it can get really huge). I'm going to search if we can limit those queries using redis.

caiordjesus (approved):
LGTM!

lyracampos (approved):

PR #257, opened 2022-01-06
Refactor/create FetchOperationsLease method

GuilhermBrSp:
## What ❓ 
Create new variadic function FetchOperationsLease on operation lease storage to retrieve a list of operation lease list given an arbitrary number of op ids arguments.

## Why 🤔 
Currently, we were only providing one method for retrieving just the lease ttl, with this function we seek to optimize the number of redis queries made.

## How 👀 
-  create FetchOperationsLease with unitary tests
- update redis version to 6.2.0 (the ZMSCORE method which we are using only exist from version 6.2.0 onwards)
- small fix to the GetRedisConnection method, which was using redis logical database indexes that don't exist ( >= 16) in scenarios where the test file contains more than 16 test cases (16 connections).
- properly configure mockgen for operation lease interface and implementation


gen/gen.go
```diff
@@ -34,6 +34,7 @@ package gen
 //go:generate go run github.com/golang/mock/mockgen -source=../internal/core/ports/room_storage.go -destination=../internal/adapters/room_storage/mock/mock.go -package=mock
 //go:generate go run github.com/golang/mock/mockgen -source=../internal/core/ports/instance_storage.go -destination=../internal/adapters/instance_storage/mock/mock.go -package=mock
 //go:generate go run github.com/golang/mock/mockgen -source=../internal/core/ports/operation_storage.go -destination=../internal/adapters/operation_storage/mock/mock.go -package=mock
+//go:generate go run github.com/golang/mock/mockgen -source=../internal/core/ports/operation_lease_storage.go -destination=../internal/adapters/operation_lease/mock/mock.go -package=mock
```
GuilhermBrSp, line 37: @lyracampos @caiordjesus Every time that we create a new interface that we want to mock, we have to add its target here so the `make generate` command can generate the mock file automatically.

internal/core/ports/operation_lease_storage.go
```diff
@@ -36,8 +36,10 @@ type OperationLeaseStorage interface {
 	RevokeLease(ctx context.Context, schedulerName, operationID string) error
 	// RenewLease renews an active lease adding the specified TTL to it. NOTE: The lease must exist (by calling GrantLease) before renewing it.
 	RenewLease(ctx context.Context, schedulerName, operationID string, ttl time.Duration) error
-	// FetchLeaseTTL fetches the time the lease will expire given the operation.
+	// FetchLeaseTTL fetches the time the lease will expire given the operation. Return error if the lease does not exist.
```
GuilhermBrSp, line 39: @lyracampos @caiordjesus What do you guys think about letting the FetchLeaseTTL method here? should we delete it or not?

GuilhermBrSp: I still se advantage in having a method like this here, with new `FetchOperationsLease` method we reduce the number of redis query to 1, but the time complexity of the method it uses is linear [O(n)](https://redis.io/commands/zmscore) while the method FetchLeaseTTL is constant [O(1)](https://redis.io/commands/zscore). In a scenario where we want to reduce the overall query time (given a lease list) and not the number of connections, we could use FetchLeaseTTL method launching a goroutine for each operation id. The average time complexity would be O(1) vs the O(n) of FetchOperationsLease.

gen/gen.go
caiordjesus: Awesome! Thanks for the information!

internal/core/ports/operation_lease_storage.go
caiordjesus: 100% agree!

internal/core/services/operation_manager/operation_manager.go
```diff
@@ -267,13 +267,22 @@ func (om *OperationManager) StartLeaseRenewGoRoutine(operationCtx context.Contex
...
+		opIds = append(opIds, op.ID)
+	}
+
+	leases, err := om.leaseStorage.FetchOperationsLease(ctx, schedulerName, opIds...)
+	if err != nil {
+		return fmt.Errorf("failed to fetch operations lease for scheduler %s: %w", schedulerName, err)
+	}
+
+	for _, lease := range leases {
+		opMap[lease.OperationID].Lease = lease
```
caiordjesus, line 283: WDYT about creating a method "setLease()" for the entity operation and use it here? Maybe it would be a good start for future validations

GuilhermBrSp: https://github.com/topfreegames/maestro/pull/257/commits/f452b6b4816c76362929e2c3e3109afc4fc5c655

internal/core/services/operation_manager/operation_manager.go
```diff
@@ -267,13 +267,22 @@ func (om *OperationManager) StartLeaseRenewGoRoutine(operationCtx context.Contex
...
-		ttl, err := om.leaseStorage.FetchLeaseTTL(ctx, schedulerName, op.ID)
-		if err != nil {
-			return fmt.Errorf("failed to fetch lease data for scheduler %s operation %s: %w", op.ID, schedulerName, err)
-		}
-		op.Lease = operation.OperationLease{OperationID: op.ID, Ttl: ttl}
+		opMap[op.ID] = op
+		opIds = append(opIds, op.ID)
+	}
+
+	leases, err := om.leaseStorage.FetchOperationsLease(ctx, schedulerName, opIds...)
```
caiordjesus, line 277: Although it might be unusual (a rare case) but wouldn't be prudent to check if len(opIds) > 0 before calling this method? Actually seems a validation we could do even before calling **addOperationsLeaseData**...

GuilhermBrSp: Definitly!

caiordjesus (approved):
LGTM!

PR #258, opened 2022-01-06
Feature/update list operations api to include lease info

GuilhermBrSp:
What ❓
Update mapping function in operations handler to parse and include lease info in list operations response.

Why 🤔
Now with the lease mechanism, it would be good to include this kind of information in the operation listing api.

How 👀
Update OperationsHandler mapping function and tests


caiordjesus (approved):
LGTM!

PR #259, opened 2022-01-11
Feature/create e2e tests for op lease

GuilhermBrSp:
## What ❓ 
Add operation lease feature e2e tests to ensure the lease keeps being renewed properly when the operation is executing.

## Why 🤔 
We have the operations lease (core) feature fully implemented, so we need to provide some e2e tests to validate the feature and explore edge cases fully.

## How 👀 
- Exposing lease storage env vars in docker file and adding its value in docker-compose.
- Creating e2e tests for operations lease.
- Adding a fix to RevokeLease method call (it was using the wrong context).
- Adding a minor fix to the default sorting order of operation listing endpoint.


internal/api/handlers/operations_handler.go
```diff
@@ -152,7 +152,7 @@ func sortOperationsByCreatedAt(operations []*operation.Operation, order string)
 }
 
 func extractSortingParameters(orderBy string) (string, error) {
-	sortingOrder := "asc"
+	sortingOrder := "desc"
```
GuilhermBrSp, line 155: @caiordjesus @lyracampos We were using the wrong sorting order here, we want to show the latest operations first (decrescent order) if the sorting parameters aren't provided

internal/core/workers/operation_execution_worker/operation_execution_worker.go
```diff
@@ -154,7 +154,7 @@ func (w *OperationExecutionWorker) Start(ctx context.Context) error {
 		if err != nil {
 			loopLogger.Error("failed to finish operation", zap.Error(err))
 		}
-		err = w.operationManager.RevokeLease(operationContext, op)
+		err = w.operationManager.RevokeLease(ctx, op)
```
GuilhermBrSp, line 157: @caiordjesus @lyracampos  We can not use the operation context for this method since the worker calls this method and is not attached to the operation execution. Using operation context here causes an error in which the lease is not revoked if the operation is canceled, it fails.

lyracampos (approved):

PR #261, opened 2022-01-13
implement validations to scheduler aggregate entities.

lyracampos:
### What ❓ 
Adding to maestro a new way to validate structs and use this validations in our use cases.

### Why 🤔 
Nowadays, Maestro Next doesn't have a well-defined way to validate domain entities. 
As Maestro Next has been growing fast, it is necessary to have theses validation's definitions well-defined, solid, generic and that can be usage by all entities.

### How 👀 
First of all we change validation plugin from [validator](https://github.com/go-validator/validator/tree/v2) to [playgroud/validator](https://github.com/go-playground/validator). This new pluging give to us more flexibility to create custom validations and also it's have a lot of built-in validations ready to use.

Also was created a **validations package** it's concern is about manager objects, create and register custom validations.
To validate Scheduler Aggregate Entities we used some of built-in validations and have to create some to customized validations too:

- VersionSemanticValidation: check if scheduler version is well-formatted following semantic lógic
- PortProtocolValidation: check if ports protocol is supported by kubernetes
- ImagePullPolicyValidation: check if policy is supported by kubernetes

With all validations ready we use them in our use cases, like CreateScheduler and UpdateScheduler

Also was written unit tests to cover the new scenarios.


internal/api/handlers/schedulers_handler.go
```diff
@@ -162,24 +159,24 @@ func (h *SchedulersHandler) UpdateScheduler(ctx context.Context, request *api.Up
...
-		Spec: game_room.Spec{
-			Version:                request.GetVersion(),
-			TerminationGracePeriod: time.Duration(request.GetTerminationGracePeriod()),
-			Affinity:               request.GetAffinity(),
-			Toleration:             request.GetToleration(),
-			Containers:             h.fromApiContainers(request.GetContainers()),
-		},
-	}
+func (h *SchedulersHandler) fromApiCreateSchedulerRequestToEntity(request *api.CreateSchedulerRequest) (*entities.Scheduler, error) {
+	scheduler, err := entities.NewScheduler(
```
GuilhermBrSp, line 163: Maybe we can return result of this method directly? we are not using `scheduler` or `err` variables, right?

internal/core/entities/game_room/spec.go
```diff
@@ -22,15 +22,34 @@
...
+	"time"
+)
 
 type Spec struct {
-	Version                string        `validate:"min=1"`
-	TerminationGracePeriod time.Duration `validate:"min=1"`
-	Containers             []Container   `validate:"min=1"`
+	Version                string        `validate:"required,semantic_version"`
+	TerminationGracePeriod time.Duration `validate:"gt=0"`
+	Containers             []Container   `validate:"required,dive"`
```
GuilhermBrSp, line 34: @lyracampos whats does `dive` means here?

internal/core/entities/scheduler.go
```diff
@@ -43,12 +44,28 @@ const (
...
-	State           string
+	Name            string `validate:"required"`
+	Game            string `validate:"required"`
+	State           string `validate:"required"`
 	RollbackVersion string
 	Spec            game_room.Spec
 	PortRange       *PortRange
 	CreatedAt       time.Time
-	MaxSurge        string `validate:"min=1"`
+	MaxSurge        string `validate:"required"`
```
GuilhermBrSp, line 54: @lyracampos if i remeber correctly, this maxsurge field can be either an empty string, or an integer number, or a percentage (number with suffix %, e.g 5%), what do you think about including this validation here? you can see how we parse it with more details here: internal/core/services/room_manager/room_manager.go:278

internal/core/entities/scheduler.go
```diff
@@ -43,12 +44,28 @@ const (
...
+		Game:      game,
+		State:     state,
+		Spec:      spec,
+		PortRange: portRange,
+		MaxSurge:  maxSurge}
+	return scheduler, scheduler.Validate()
+}
+
+func (s *Scheduler) Validate() error {
+	err := validations.Validate.Struct(s)
```
GuilhermBrSp, line 69: @lyracampos when we run this validate method, does it validates inner structs also? like PortRange and Spec?

internal/core/services/scheduler_manager/scheduler_manager.go
```diff
@@ -69,7 +66,7 @@ func (s *SchedulerManager) CreateScheduler(ctx context.Context, scheduler *entit
 
 	operation, err := s.operationManager.CreateOperation(ctx, scheduler.Name, &create_scheduler.CreateSchedulerDefinition{})
 	if err != nil {
-		return nil, fmt.Errorf("failed to schedule 'create scheduler' operation: %w", err)
+		return nil, fmt.Errorf("failed create schedule - fails operation: %w", err)
```
GuilhermBrSp, line 69: Here, the error message should be failing in creating the operation, right?

GuilhermBrSp (commented):
I think it is missing to run a `go mod tidy` to remove the validation lib we are not using anymore

internal/core/entities/game_room/spec.go
lyracampos: @GuilhermBrSp  **dive** means that the slice validations will be checked automatically.
in this case, all container slice validations will be checked whenever the scheduler is being validated

if you don't use dive in slice you'll need to call validations one by one for each slice item

internal/core/entities/scheduler.go
lyracampos: @GuilhermBrSp good! I'll create a new custom validation for this field

internal/core/entities/scheduler.go
lyracampos: @GuilhermBrSp yes it does! all of structs will be checked when scheduler validate is call

caiordjesus (approved):
LGTM!

GuilhermBrSp (approved):

PR #262, opened 2022-01-18
adds forwarder (present on scheduler as list)

caiordjesus:
### What?
We want to have forwarder options configured on Scheduler. That said, to enable forwarders on maestro, such as the feature we had on v9.
### Why?
To forward events from the game rooms being orchestrated by Maestro.
### How?
Creating a new entity called Forwarder, and adding it (as a list) to the Scheduler entity. Also, refactoring the pg adapter under scheduler_storage to parse unmarshall this information from the scheduler


internal/adapters/scheduler_storage/pg/scheduler_test.go
```diff
@@ -27,14 +27,35 @@ package pg
...
+	err := validations.RegisterValidations()
+	if err != nil {
+		t.Errorf("unexpected error %d'", err)
+	}
 	t.Run("valid schedulers", func(t *testing.T) {
+		fwd := &forwarder.Forwarder{
+			Name:    "fwd",
+			Enabled: true,
+			FwdType: forwarder.TypeGrpc,
+			Address: "address",
```
caiordjesus, line 51: @lyracampos @GuilhermBrSp This line is different from the RFC, but I think it makes a lot more sense (as we discussed before). WDYT?

internal/adapters/scheduler_storage/pg/scheduler_test.go
```diff
@@ -27,14 +27,35 @@ package pg
...
+
+	"github.com/topfreegames/maestro/internal/validations"
 
 	"github.com/stretchr/testify/require"
 	"github.com/topfreegames/maestro/internal/core/entities"
 	"github.com/topfreegames/maestro/internal/core/entities/game_room"
 )
 
 func TestScheduler_ToScheduler(t *testing.T) {
+	err := validations.RegisterValidations()
```
lyracampos, line 42: @caiordjesus why did you register the validations here?

internal/adapters/scheduler_storage/pg/scheduler_test.go
lyracampos: @caiordjesus are you talking about the name of FwdOptions property? 
i think you should use **Options** only, because you are already in forwarder package context. 

but if you don't want to use only Options, I suggest you to use **ForwarderOptions** instead, because Fwd doesn't means anything.

internal/api/handlers/schedulers_handler_test.go
```diff
@@ -54,7 +56,10 @@ import (
 )
 
 func TestGetAllSchedulers(t *testing.T) {
-
+	err := validations.RegisterValidations()
```
lyracampos, line 59: @caiordjesus  why did you register the validations here? I think it's not necessary.

internal/api/handlers/schedulers_handler_test.go
```diff
@@ -161,7 +166,10 @@ func TestGetAllSchedulers(t *testing.T) {
 }
 
 func TestGetScheduler(t *testing.T) {
-
+	err := validations.RegisterValidations()
```
lyracampos, line 169: @caiordjesus here too. I think it's not necessary to register validations here.

internal/core/entities/forwarder/forwarder.go
```diff
@@ -0,0 +1,37 @@
...
+package forwarder
+
+type FwdType string
+
+const (
+	TypeGrpc FwdType = "gRPC"
+)
+
+type Forwarder struct {
+	Name    string `validate:"required"`
```
lyracampos, line 32: @caiordjesus I think you could use Type instead of FwdType and Options instead of FwdOptions. Because you are already in a Forwarder Struct and it's already give us a forwarder context.

Avoid use words like Fwd because it's doesn't exists in real world.

internal/core/entities/scheduler.go
```diff
@@ -52,16 +53,19 @@ type Scheduler struct {
 	PortRange       *PortRange
 	CreatedAt       time.Time
 	MaxSurge        string `validate:"required,max_surge"`
+	Forwarders      []*forwarder.Forwarder
```
lyracampos, line 56: @caiordjesus use Dive decorator to indicate that Forwarders[]'s will be validate when Scheduler is being validated

internal/core/entities/forwarder/forwarder.go
caiordjesus: Nice, thx!!

internal/core/entities/scheduler.go
caiordjesus: Done! Thx

lyracampos (approved):
LGTM 👍

PR #266, opened 2022-01-24
produceEvent forwarder event service

caiordjesus:
### What?
Refactor forwarder event service
### Why?
We have initially implemented some forwarder tasks using an initial version of the EventForwarder interface, now that we have a final version of RFC, we need to update the EventsService implementation according to it.
### How?
Changing produceEvent method on events_forwarder_service and method called together


internal/core/services/events_forwarder/events_forwarder_service.go
```diff
@@ -24,60 +24,169 @@ package events_forwarder
...
 func (es *EventsForwarderService) ProduceEvent(ctx context.Context, event *events.Event) error {
-	room := &game_room.GameRoom{}
-	instance := &game_room.Instance{}
-
-	switch event.Name {
-	case events.RoomEvent:
-		return es.forwardRoomEvent(ctx, room, instance, map[string]interface{}{}, "")
-	case events.PlayerEvent:
-		return es.forwardPlayerEvent(ctx, room, map[string]interface{}{}, "")
+	var eventType string
```
lyracampos, line 67: To be more readable, you can switch if condition to avoid else statement, like this: 

if _, ok := event.Attributes["eventType"]; !ok { 
	return errors.New("eventAttributes must contain key \"eventType\"")
}

eventType := event.Attributes["eventType"].(string)
…..

internal/core/services/events_forwarder/events_forwarder_service.go
```diff
@@ -24,60 +24,169 @@ package events_forwarder
...
+	if _, ok := event.Attributes["eventType"]; ok {
+		eventType = event.Attributes["eventType"].(string)
+	} else {
+		return errors.New("eventAttributes must contain key \"eventType\"")
 	}
-	return nil
 
+	var scheduler *entities.Scheduler
+	var err error
+	scheduler, _ = es.schedulerCache.GetScheduler(ctx, event.SchedulerID)
```
lyracampos, line 76: to avoid code smell (Bloaters), I think you could create an internal func to get scheduler and there you put logic to find scheduler, get from cache etc...

then just replace the lines between 76 to 86 to
scheduler := es.getScheduler(ctx, event.SchedulerID)

internal/core/services/events_forwarder/events_forwarder_service.go
```diff
@@ -24,60 +24,169 @@ package events_forwarder
...
+	eventType string,
+	instance game_room.Instance,
+	scheduler entities.Scheduler,
+	_forwarder *forwarder.Forwarder,
+) error {
+	var selectedPort int32
+	if len(instance.Address.Ports) > 0 {
+		selectedPort = instance.Address.Ports[0].Port
+	} else {
+		return errors.New("no room port found to forward roomEvent")
```
lyracampos, line 128: @caiordjesus to help troubleshooting you could add more information to enrich the log. Maybe you can add schedulerName or ForwarderName

internal/core/services/events_forwarder/events_forwarder_service.go
```diff
@@ -24,60 +24,169 @@ package events_forwarder
...
-	err := es.eventsForwarder.ForwardRoomEventObsolete(ctx, room, instance, attributes, options)
+func (es *EventsForwarderService) forwardRoomEvent(
+	ctx context.Context,
+	event *events.Event,
+	eventType string,
+	instance game_room.Instance,
+	scheduler entities.Scheduler,
+	_forwarder *forwarder.Forwarder,
+) error {
+	var selectedPort int32
```
lyracampos, line 124: @caiordjesus here I think you could switch conditions too like

if len(instance.Address.Ports) <= 0 {
		return errors.New("no room port found to forward roomEvent")	
	}
selectedPort := instance.Address.Ports[0].Port
….

internal/core/services/events_forwarder/events_forwarder_service.go
```diff
@@ -24,60 +24,169 @@ package events_forwarder
...
+	if len(instance.Address.Ports) > 0 {
+		selectedPort = instance.Address.Ports[0].Port
+	} else {
+		return errors.New("no room port found to forward roomEvent")
+	}
+
+	roomEvent := events.RoomEventType(eventType)
+
+	var pingType events.RoomPingEventType
```
lyracampos, line 133: @caiordjesus here you could change if declarations to avoid statement like previous comments

internal/core/services/events_forwarder/events_forwarder_service.go
```diff
@@ -24,60 +24,169 @@ package events_forwarder
...
+	if roomEvent == events.Ping {
+		if _type, ok := event.Attributes["pingType"]; ok {
```
lyracampos, line 135: @caiordjesus here you could change if declarations to avoid statement like previous comments

internal/core/services/events_forwarder/events_forwarder_service.go
```diff
@@ -24,60 +24,169 @@ package events_forwarder
...
+		EventType: roomEvent,
+		PingType:  &pingType,
+		Other:     event.Attributes,
+	}
+	err := es.eventsForwarder.ForwardRoomEvent(ctx, roomAttributes, *_forwarder)
 	if err != nil {
-		reportRoomEventForwardingFailed(room.SchedulerID)
-		es.logger.Error(fmt.Sprintf("Failed to forward room events for room %s and scheduler %s", room.ID, room.SchedulerID), zap.Error(err))
+		reportRoomEventForwardingFailed(event.SchedulerID)
+		es.logger.Error(fmt.Sprintf("Failed to forward room events for room %s and scheduler %s", event.RoomID, event.SchedulerID), zap.Error(err))
```
lyracampos, line 155: @caiordjesus I think you could add forwarder Name in log.Error. Once scheduler can have a list of forwarder.

internal/core/services/events_forwarder/events_forwarder_service.go
```diff
@@ -24,60 +24,169 @@ package events_forwarder
...
 
-func (es *EventsForwarderService) forwardPlayerEvent(ctx context.Context, room *game_room.GameRoom, attributes map[string]interface{}, options interface{}) error {
-	err := es.eventsForwarder.ForwardPlayerEventObsolete(ctx, room, attributes, options)
+func (es *EventsForwarderService) forwardPlayerEvent(
+	ctx context.Context,
+	event *events.Event,
+	eventType string,
+	_forwarder *forwarder.Forwarder,
+) error {
+	var playerId string
```
lyracampos, line 169: @caiordjesus here you could change if declarations to avoid statement like previous comments

internal/core/services/events_forwarder/events_forwarder_service.go
caiordjesus: Nice!!

internal/core/services/events_forwarder/events_forwarder_service.go
caiordjesus: A lot better for sure! Thanks!!

internal/core/services/events_forwarder/events_forwarder_service.go
caiordjesus: 100%

internal/core/services/events_forwarder/events_forwarder_service.go
```diff
@@ -190,3 +174,20 @@ func (es *EventsForwarderService) forwardPlayerEvent(
 	reportPlayerEventForwardingSuccess(event.SchedulerID)
 	return nil
 }
+
+func (es *EventsForwarderService) getScheduler(ctx context.Context, schedulerName string) (*entities.Scheduler, error) {
+	var scheduler *entities.Scheduler
+	var err error
+	scheduler, _ = es.schedulerCache.GetScheduler(ctx, schedulerName)
```
lyracampos, line 181: @caiordjesus I think is important to verify if schedulerCache.GetScheduler return an error. Because you are testing if scheduler was found in cache, but we can receive an error too, like connection timeout or something like that.

Could you try:

- remove lines 179 and 180
- scheduler, err := es.schedulerCache.GetScheduler(ctx, schedulerName)
- test if returns err
- ....

caiordjesus: Thx!

internal/core/services/events_forwarder/events_forwarder_service.go
```diff
@@ -24,60 +24,171 @@ package events_forwarder
...
+	eventType := event.Attributes["eventType"].(string)
 
+	scheduler, err := es.getScheduler(ctx, event.SchedulerID)
+	if err != nil {
+		return err
+	}
+
+	forwarderList := scheduler.Forwarders
+	if len(forwarderList) > 0 {
+		for _, _forwarder := range forwarderList {
```
GuilhermBrSp, line 79: @caiordjesus is there any reason for using underscore in the `forwarder` variable name in this loop?

GuilhermBrSp: If not, I would remove it to make it more readable, here, and in the methods arguments below

internal/core/services/events_forwarder/events_forwarder_service.go
```diff
@@ -24,60 +24,171 @@ package events_forwarder
...
+			switch event.Name {
+			case events.RoomEvent:
+				instance, err := es.instanceStorage.GetInstance(ctx, event.SchedulerID, event.RoomID)
+				if err != nil {
+					es.logger.Error(fmt.Sprintf("Failed to get instance for room \"%v\" from scheduler \"%v\" info", event.RoomID, event.SchedulerID), zap.Error(err))
+					return err
+				}
+				err = es.forwardRoomEvent(ctx, event, eventType, *instance, *scheduler, _forwarder)
```
GuilhermBrSp, line 87: We can return the es.forwardRoomEvent method value here, no need for checking error since we are doing nothing

internal/core/services/events_forwarder/events_forwarder_service.go
```diff
@@ -24,60 +24,171 @@ package events_forwarder
...
+				if err != nil {
+					return err
+				}
+			case events.PlayerEvent:
+				err = es.forwardPlayerEvent(ctx, event, eventType, _forwarder)
+				if err != nil {
```
GuilhermBrSp, line 93: same here

config/worker.local.yaml
```diff
@@ -31,6 +31,9 @@ adapters:
   instanceStorage:
     redis:
       url: "redis://localhost:6379/0"
```
GuilhermBrSp, line 33: @caiordjesus you need to add scheduler storage and scheduler cache configs in rooms-api YAML file as well

internal/core/services/events_forwarder/events_forwarder_service.go
caiordjesus: forwarder was shadowing forwarder entity package. I could either give an alias to the imported package or for the item list. Both options seemed less readable than simply using an underscore

internal/core/services/events_forwarder/events_forwarder_service.go
caiordjesus: Actually we need since it's a loop. If I put return there, I would forward only for the first forwarder of the list

internal/core/services/events_forwarder/events_forwarder_service.go
```diff
@@ -24,60 +24,171 @@ package events_forwarder
...
+	instance game_room.Instance,
+	scheduler entities.Scheduler,
+	_forwarder *forwarder.Forwarder,
+) error {
+	if len(instance.Address.Ports) == 0 {
+		return fmt.Errorf("no room port found to forward roomEvent. Forwarder name: \"%v\", Scheduler: \"%v\"", _forwarder.Name, event.SchedulerID)
+	}
+	selectedPort := instance.Address.Ports[0].Port
+
+	roomEvent := events.RoomEventType(eventType)
```
GuilhermBrSp, line 118: @caiordjesus you can pass any string for this eventType value and the code will not break, what about creating and using some methods for converting enums instead of using weak typecast here? We can use something like the convert methods from GameRoomPingStatus internal/core/entities/game_room/game_room.go:103, what do you think?

config/worker.local.yaml
caiordjesus: Nice! Just did it!

internal/core/services/events_forwarder/events_forwarder_service.go
caiordjesus: Nice, I'll make this improve right now!

config/rooms-api.local.yaml
```diff
@@ -18,6 +18,12 @@ adapters:
   instanceStorage:
     redis:
       url: "redis://localhost:6379/0"
+  schedulerStorage:
```
lyracampos, line 21: WDYT to reorganize the configurations in contexts? 
You could create scheduler context and put all scheduler configuration there.

would be like this:

```
scheduler:
       storage:
             postgres:
                   url: "postgres://user:pwd@localhost:5432/maestro?sslmode=disable"
        cache:
             redis:
                    url: "redis://localhost:6379/0"
```

config/runtime-watcher.local.yaml
```diff
@@ -31,6 +31,9 @@ adapters:
   instanceStorage:
     redis:
       url: "redis://localhost:6379/0"
+  schedulerCache:
```
lyracampos, line 34: @caiordjesus here too

lyracampos (approved):

config/rooms-api.local.yaml
caiordjesus: It's a change of pattern used now on next and will require some refactoring. Do you think this could be done on another PR?

GuilhermBrSp (approved):

PR #272, opened 2022-02-01
creates switch active version operation

caiordjesus:
### WHAT?
adds switch active version operation and tests
### WHY?
We want to have an operation that switches the active scheduler version to a specified one, switching the GRUs and the database registry.
### HOW?
implementing a new operation (like update scheduler operation), responsible for replacing the GRUs using the max surge.  If something goes wrong, the operation tries to delete all created rooms and recreate the deleted ones


internal/core/operations/switch_active_version/switch_active_version_executor.go
```diff
@@ -0,0 +1,197 @@
...
+
+// Execute the process of switching a scheduler active version consists of the following:
+// 1. Creates "replace" goroutines (same number as MaxSurge);
+// 2. Each goroutine will listen to a channel and create a new room using the
+//    new configuration. After the room is ready, it will then delete the room
+//    being replaced;
+// 3. List all game rooms that need to be replaced and produce them into the
+//    replace goroutines channel;
+// 4. Switch the active version
+func (e *SwitchActiveVersionExecutor) Execute(ctx context.Context, op *operation.Operation, definition operations.Definition) error {
```
GuilhermBrSp, line 65: @caiordjesus what do you think about renaming this method receiver to `s` or `se`? Ref: https://blog.devgenius.io/how-should-you-name-your-receivers-in-go-aec60abd7f67

internal/core/operations/switch_active_version/switch_active_version_executor.go
```diff
@@ -0,0 +1,197 @@
...
+	logger := zap.L().With(
+		zap.String("scheduler_name", op.SchedulerName),
+		zap.String("operation_definition", definition.Name()),
+		zap.String("operation_id", op.ID),
+	)
+	logger.Debug("start switching scheduler active version")
+
+	updateDefinition := definition.(*SwitchActiveVersionDefinition)
```
GuilhermBrSp, line 73: @caiordjesus what do you think about adding a safe check here in this type cast?

internal/core/operations/switch_active_version/switch_active_version_executor.go
```diff
@@ -0,0 +1,197 @@
...
+
+	roomsChan := make(chan *game_room.GameRoom)
+	var maxSurgeWg sync.WaitGroup
+
+	maxSurgeWg.Add(maxSurgeNum)
+	for i := 0; i < maxSurgeNum; i++ {
+		go e.replaceRoom(logger, &maxSurgeWg, roomsChan, e.roomManager, *scheduler)
+	}
+
+roomsListLoop:
```
GuilhermBrSp, line 89: @caiordjesus What about isolating this loop execution and logic into a separate private method,something like `se.startReplaceLoop(ctx context.Context, scheduler *entities.Scheduler, maxSurgeNum int, roomsChan chan *game_room.GameRoom) error` or `se.executeReplaceLoop(ctx context.Context, scheduler *entities.Scheduler, maxSurgeNum int, roomsChan chan *game_room.GameRoom) error`

internal/core/services/scheduler_manager/scheduler_manager.go
```diff
@@ -205,3 +205,19 @@ func isMajorVersionUpdate(currentScheduler, newScheduler *entities.Scheduler) bo
 		),
 	)
 }
+
```
GuilhermBrSp, line 208: @caiordjesus Missing unitary tests for these new guys

internal/core/operations/switch_active_version/switch_active_version_executor_test.go
```diff
@@ -0,0 +1,382 @@
...
+	if err != nil {
+		t.Errorf("unexpected error %d'", err)
+	}
+
+	newScheduler := newValidScheduler()
+	definition := &switch_active_version.SwitchActiveVersionDefinition{
+		NewActiveScheduler: newScheduler,
+	}
+
+	t.Run("should succeed - Execute on error if operation finishes (no created rooms)", func(t *testing.T) {
```
GuilhermBrSp, line 271: @caiordjesus I think the major feature of the OnError logic is to remove rooms that were created, right? what about adding a test case exploring this scenario here?

internal/core/operations/switch_active_version/switch_active_version_executor_test.go
```diff
@@ -0,0 +1,382 @@
...
+
+	newScheduler := newValidScheduler()
+	newScheduler.PortRange.Start = 1000
+	newScheduler.MaxSurge = "3"
+
+	definition := &switch_active_version.SwitchActiveVersionDefinition{
+		NewActiveScheduler: newScheduler,
+	}
+
+	t.Run("should succeed - Execute switch active version operation", func(t *testing.T) {
```
GuilhermBrSp, line 95: @caiordjesus WDYT about adding scenarios for when the `ListRoomsWithDeletionPriority` fails?

GuilhermBrSp: @caiordjesus Also, one important scenario is when the game room creation fails during replacement, we don't want to end the operation with success when this happens, right?

internal/core/operations/switch_active_version/switch_active_version_executor.go
```diff
@@ -0,0 +1,197 @@
...
+	ctx := context.Background()
+
+	for {
+		room, ok := <-roomsChan
+		if !ok {
+			return
+		}
+
+		gameRoom, _, err := roomManager.CreateRoom(ctx, scheduler)
+		if err != nil {
```
GuilhermBrSp, line 160: @caiordjesus The current code won't make the operation fail if we have an error creating the new game room, I think we need to change this logic to make the execution return with an error when this happens, right?

GuilhermBrSp (changes requested):
The operation isn't returning with error when the replacement fails

internal/core/operations/switch_active_version/switch_active_version_executor_test.go
```diff
@@ -0,0 +1,382 @@
...
+	"time"
+
+	"github.com/topfreegames/maestro/internal/core/operations/switch_active_version"
+
+	"github.com/topfreegames/maestro/internal/core/services/interfaces"
+	mockeventsservice "github.com/topfreegames/maestro/internal/core/services/interfaces/mock/events_service"
+
+	"github.com/golang/mock/gomock"
+	"github.com/stretchr/testify/require"
+	clock_mock "github.com/topfreegames/maestro/internal/adapters/clock/mock"
```
GuilhermBrSp, line 39: @caiordjesus Just a simple refactor, can we rename these import alias to remove underscores? we are now following the golang convention, ref:  https://youtrack.jetbrains.com/issue/GO-5942, https://github.com/kubernetes/kubernetes/pull/35258/files

internal/core/operations/switch_active_version/switch_active_version_executor.go
caiordjesus: Thanks for the article! I'll read right now!

caiordjesus: What do you think about using "ex", standing for executor?

GuilhermBrSp: Awesome!

internal/core/operations/switch_active_version/switch_active_version_executor.go
```diff
@@ -0,0 +1,233 @@
...
+
+func (ex *SwitchActiveVersionExecutor) startReplaceRoomsLoop(ctx context.Context, logger *zap.Logger, maxSurgeNum int, scheduler entities.Scheduler) error {
+	logger.Debug("replacing rooms loop - start")
+	roomsChan := make(chan *game_room.GameRoom)
+	var maxSurgeWg sync.WaitGroup
+	errs, ctx := errgroup.WithContext(ctx)
+
+	maxSurgeWg.Add(maxSurgeNum)
+	for i := 0; i < maxSurgeNum; i++ {
+		errs.Go(func() error {
```
GuilhermBrSp, line 183: I didn't know about this guy here, it seems to be perfect for what we wanted.

internal/core/operations/switch_active_version/switch_active_version_executor.go
```diff
@@ -0,0 +1,233 @@
...
+			err := ex.replaceRoom(logger, &maxSurgeWg, roomsChan, ex.roomManager, scheduler)
```
GuilhermBrSp, line 184: Can't we just return ex.replaceRoom(logger, &maxSurgeWg, roomsChan, ex.roomManager, scheduler) since we are doing nothing here?

internal/core/operations/switch_active_version/switch_active_version_executor.go
```diff
@@ -0,0 +1,233 @@
...
+
+func (ex *SwitchActiveVersionExecutor) replaceRoom(logger *zap.Logger, wg *sync.WaitGroup, roomsChan chan *game_room.GameRoom, roomManager interfaces.RoomManager, scheduler entities.Scheduler) error {
+	defer wg.Done()
+
+	// we're going to use a separated context for each replaceRoom since we
+	// don't want to cancel the replace in the middle (like creating a room and
+	// then left the old one (without deleting it).
+	ctx := context.Background()
+
+	for {
```
GuilhermBrSp, line 133: @caiordjesus I see value in logging if we have some error in creating or deleting rooms here

internal/core/operations/switch_active_version/switch_active_version_executor.go
```diff
@@ -0,0 +1,233 @@
...
+		return fmt.Errorf("the definition is invalid. Should be type SwitchActiveVersionDefinition")
+	}
+	scheduler := &updateDefinition.NewActiveScheduler
+
+	maxSurgeNum, err := ex.roomManager.SchedulerMaxSurge(ctx, scheduler)
+	if err != nil {
+		return fmt.Errorf("error fetching scheduler max surge: %w", err)
+	}
+
+	err = ex.startReplaceRoomsLoop(ctx, logger, maxSurgeNum, *scheduler)
```
GuilhermBrSp, line 87: WDYT about adding a log informing that the replace fails here if it does?

internal/core/operations/switch_active_version/switch_active_version_executor.go
```diff
@@ -0,0 +1,233 @@
...
+	"go.uber.org/zap"
+	"golang.org/x/sync/errgroup"
+)
+
+type SwitchActiveVersionExecutor struct {
+	roomManager        interfaces.RoomManager
+	schedulerManager   interfaces.SchedulerManager
+	roomsBeingReplaced *sync.Map
+	newCreatedRooms    map[string][]*game_room.GameRoom
+	locker             sync.Mutex
```
GuilhermBrSp, line 44: @caiordjesus why didn't we use sync.Map here?

GuilhermBrSp, line 44: @caio what about renaming this var to `newCreatedRoomsMutex` or `newCreatedRoomsLocker`?

GuilhermBrSp (approved):

internal/core/operations/switch_active_version/switch_active_version_executor.go
caiordjesus: You're right!

internal/core/operations/switch_active_version/switch_active_version_executor.go
caiordjesus: Would be very hard to append to list in key, I would have to load items from list, typecast as the game_room list, append the value and store again. Maybe it wouldn't work since between the time I load from map to the time I store, the key could've changed. I think it's more readable and maintainable this way

PR #274, opened 2022-02-03
Feature/rename update scheduler endpoint to create new version

GuilhermBrSp:
## What ❓
Refactor rename update scheduler endpoint to new scheduler version, including changes in the handler, proto definition, and scheduler manager.

## Why 🤔
With the new scheduler version and active versions concept, the update scheduler will no longer be used.

How 👀
- Refactor proto definition
- Refactor scheduler handler and scheduler manager
- Create extra test cases for increasing coverage in scheduler_manager
- Fix CreateScheduler method from the scheduler storage port, it was no longer creating the initial version in the db and that was breaking e2e tests.


internal/core/services/scheduler_manager/scheduler_manager_test.go
```diff
@@ -46,6 +46,68 @@ import (
 	"github.com/topfreegames/maestro/internal/validations"
 )
 
+func TestCreateScheduler(t *testing.T) {
```
GuilhermBrSp, line 49: @lyracampos @caiordjesus There is no new behaviour in scheduler_manager, here I just added missing test cases and scenarios to improve coverage of scheduler manager to 100%

lyracampos (approved):

PR #275, opened 2022-02-04
Feature/refactor update operation to new scheduler version

GuilhermBrSp:
## What ❓ 
Refactor the update scheduler operation and rename it to create new scheduler version.
This new operation concept consists in changing the operation responsibility. instead of making an update scheduler at database and apply the new spec in kubernetes, this new operation will assume responsibility for:

1. check if is scheduler spec is a new major version
2. check if GRU is stable (GRU stable means that GRU can send ping to Maestro) 
3. add new scheduler version at database 
4. enqueue new operation 'Switch Active Version'

## Why 🤔 
Thir PR is part of the new update scheduler refactor that we are working on right now.

## How
- Create new methods for scheduler manager
- Create new validation method in room manager
- Refactor create scheduler update with new name, new logic and tests.


lyracampos (approved):

internal/core/operations/newschedulerversion/new_scheduler_version_executor.go
```diff
@@ -0,0 +1,143 @@
...
+		return fmt.Errorf("error creating new scheduler version in db: %w", err)
+	}
+	return nil
+}
+
+func (ex *CreateNewSchedulerVersionExecutor) enqueueSwitchActiveVersionOperation(ctx context.Context, newScheduler *entities.Scheduler, logger *zap.Logger) (*operation.Operation, error) {
+	switchActiveVersionOp, err := ex.schedulerManager.EnqueueSwitchActiveVersionOperation(ctx, newScheduler)
+	if err != nil {
+		// TODO(guilhermbrsp): Maybe we should rollback the creation of the new scheduler version if some error happens here
+		logger.Error("error enqueuing switch active version operation", zap.Error(err))
```
GuilhermBrSp, line 123: @lyracampos @caiordjesus What do you guys think about this? should we rollback the creation of new scheduler here or in the OnError method?

internal/core/services/scheduler_manager/scheduler_manager.go
```diff
@@ -123,72 +146,48 @@ func (s *SchedulerManager) RemoveRooms(ctx context.Context, schedulerName string
...
-	err := scheduler.Validate()
-	if err != nil {
-		return false, err
-	}
-
+func (s *SchedulerManager) EnqueueNewSchedulerVersionOperation(ctx context.Context, scheduler *entities.Scheduler) (*operation.Operation, error) {
 	currentScheduler, err := s.schedulerStorage.GetScheduler(ctx, scheduler.Name)
-	if err != nil {
-		return false, fmt.Errorf("error fetching scheduler: %w", err)
+	if err != nil || currentScheduler == nil {
```
caiordjesus, line 151: Why is this condition "currentScheduler == nil" necessary? The GetScheduler method returns nil, err in case the query on database returns 0 rows...

internal/core/operations/newschedulerversion/new_scheduler_version_executor.go
caiordjesus: Just for record: As we discussed before, we should open a pg transaction before creating the new version, create it using the transaction, and only commit after everything works well (and rollback if anything goes wrong)

internal/core/services/room_manager/room_manager.go
```diff
@@ -310,6 +310,19 @@ func (m *RoomManager) SchedulerMaxSurge(ctx context.Context, scheduler *entities
 	return int(math.Max(minSchedulerMaxSurge, absoluteNum)), nil
 }
 
+func (m *RoomManager) ValidateGameRoomCreation(ctx context.Context, scheduler *entities.Scheduler) error {
+	gameRoom, _, err := m.CreateRoom(ctx, *scheduler)
```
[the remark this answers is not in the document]
GuilhermBrSp: It creates the room suing the runtime and them waits for it status to be `ready`

internal/core/services/scheduler_manager/scheduler_manager.go
GuilhermBrSp: This nil check is unnecessary

internal/core/operations/newschedulerversion/new_scheduler_version_executor.go
```diff
@@ -0,0 +1,143 @@
...
+			return err
+		}
+	}
+
+	err = ex.createNewSchedulerVersion(ctx, newScheduler, logger)
+	if err != nil {
+		return err
+	}
+
+	switchActiveVersionOp, err := ex.enqueueSwitchActiveVersionOperation(ctx, newScheduler, logger)
```
caiordjesus, line 93: WDYT about enqueuing the operation with a information like "isMajorChange"? Also, we would have to change on the "switch active version operation" to only switch pods on minor versions (I believe it's changing the pods nevertheless the version change"

caiordjesus (changes requested):
It's not 100% related to this PR, but we forgot to include the logic that only changes pods on major changes

internal/core/operations/newschedulerversion/new_scheduler_version_executor.go
lyracampos: @GuilhermBrSp I think we could do the rollback in this method. In my view, it's more contextualized. If you put it on onError, you may have to check the type of error to do rollback or not. I think it would be weird.

internal/core/operations/switch_active_version/switch_active_version_executor.go
```diff
@@ -78,19 +78,22 @@ func (ex *SwitchActiveVersionExecutor) Execute(ctx context.Context, op *operatio
 		return fmt.Errorf("the definition is invalid. Should be type SwitchActiveVersionDefinition")
 	}
 	scheduler := &updateDefinition.NewActiveScheduler
+	replacePods := updateDefinition.ReplacePods
```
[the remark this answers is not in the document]
caiordjesus: Agree with you guys 100%

lyracampos (approved):

caiordjesus (approved):

PR #279, opened 2022-02-11
expose switch active version to rest/grpc services.

lyracampos:
**What?**
Create endpoint to expose feature Switch Active Version to be called from grpc/rest-api

**Why?**
So that end users can define which version is active for the scheduler. 

**How**?
As the core feature already exists in domain, we only create a proto definition with messages and service to switch version. Also, was created an new endpoint which enables http calls.


internal/adapters/scheduler_storage/pg/scheduler_version_test.go
```diff
@@ -0,0 +1,23 @@
+// MIT License
```
GuilhermBrSp, line 1: @lyracampos I guess we are not using this file anymore

internal/core/entities/scheduler_test.go
```diff
@@ -115,3 +115,47 @@ func TestNewScheduler(t *testing.T) {
...
 	})
 }
+
+func TestIsMajorVersion(t *testing.T) {
+	tests := map[string]struct {
+		currentScheduler *entities.Scheduler
+		newScheduler     *entities.Scheduler
+		expected         bool
+	}{
+		"port range should be a major update": {
```
GuilhermBrSp, line 125: Hmmmm... using table-driven tests here, right? I like this approach and I think we should adopt it in more tests, what do you guys think? @lyracampos @caiordjesus ref: https://dev.to/boncheff/table-driven-unit-tests-in-go-407b, https://dave.cheney.net/2019/05/07/prefer-table-driven-tests

GuilhermBrSp (changes requested):
Need to remove unused file

GuilhermBrSp (approved):

internal/core/entities/scheduler_test.go
lyracampos: This tests already exists before. I just moved it to new file. I didn't know Table Driven but I found it very interesting. for me we adopted

caiordjesus: Very nice!!!

caiordjesus (approved):
LGTM!

PR #280, opened 2022-02-11
switch active version e2e test

caiordjesus:
### What?
Implement e2e tests for switch active version use case.
### Why?
Sometimes, the client wants to rollback a scheduler version to an older version. To be able to do that, we needed a 
functionality to switch the active version of a scheduler. To guarantee the whole thing is happening as expected, we implement an e2e test.
### How?
Following the pattern of other e2e tests.

### DIFF
--- a/internal/api/handlers/schedulers_handler.go
+++ b/internal/api/handlers/schedulers_handler.go
@@ -71,7 +71,14 @@ func (h *SchedulersHandler) ListSchedulers(ctx context.Context, message *api.Lis
 func (h *SchedulersHandler) GetScheduler(ctx context.Context, request *api.GetSchedulerRequest) (*api.GetSchedulerResponse, error) {
 	var scheduler *entities.Scheduler
 	var err error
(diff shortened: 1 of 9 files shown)


internal/api/handlers/schedulers_handler.go
```diff
@@ -71,7 +71,14 @@ func (h *SchedulersHandler) ListSchedulers(ctx context.Context, message *api.Lis
 func (h *SchedulersHandler) GetScheduler(ctx context.Context, request *api.GetSchedulerRequest) (*api.GetSchedulerResponse, error) {
 	var scheduler *entities.Scheduler
 	var err error
-	scheduler, err = h.schedulerManager.GetScheduler(ctx, request.GetSchedulerName(), request.GetVersion())
+
```
GuilhermBrSp, line 74: Unnecessary blank lines here

GuilhermBrSp (approved):

PR #281, opened 2022-02-14
Refactor/create operations manager port

GuilhermBrSp:
## What ❓ 
Creating new primary port for OperationManager, refactor operation_manager service, creating a new file to place both primary and secondary ports (https://alistair.cockburn.us/hexagonal-architecture/) grouped by contexts. Refactor tests to use new operation manager mock.

## Why 🤔 
Currently, we can't generate mocks for our domain services, due to the fact that that lib we are using for generating mocks can only generate mock for interfaces and not concrete implementations. This causes us to mock service internals instead of the service itself while testing components that depend on the service (e.g handlers and executors).
This can lead to really huge test mocking and preparation in components that should be mocking those services. [https://github.com/golang/mock/issues/332.](https://github.com/golang/mock/issues/332). To solve this problem we decided to do a major refactor to create the OperationManager interface, which will work as a primary port.

## How 👀 
- Create primary port for OperationManager.
- Group operations ports into a single file.
- Create mock for OperationManager.
- Refactor tests to use operation manager mock.


internal/api/handlers/ping_handler_test.go
```diff
@@ -44,7 +44,7 @@ func TestPingHandler(t *testing.T) {
 	t.Run("with valid request", func(t *testing.T) {
 
 		mockCtrl := gomock.NewController(t)
-		defer mockCtrl.Finish()
+
```
GuilhermBrSp, line 47: Doc from the mock lib:
```
New in go1.14+, if you are passing a *testing.T into this function you no longer need to call ctrl.Finish() in your test methods
```
We don't need to call.Finish() anymore

PR #289, opened 2022-02-16
Refactor/create single entrypoint

guilhermocc:
## What ❓ 
Create single entrypoint for maestro, refactor maestro components initialization logic and update makefile commands

## Why 🤔 
After a lot of discussions, the team has decided that the current multi-entrypoint model of maestro makes it harder for launching new maestro versions and creating a deployment. We have defined that maestro should follow the single entry point model, using command parameters for switching between maestro components. This will also enable us to have a single docker image for the project instead of 4.

## How 👁️ 
- Create new command chain with a single entrypoint for maestro
- Made some refactor in maestro components initialization and migration.


caiordjesus (approved):
LGTM!

PR #290, opened 2022-02-17
Add make help

arthur29:
### Why

Currently, Maestro doesn't have a quick way to show the make targets and what it does.

### What

This PR proposes to add the target `help` in the Makefile that parses the commands and shows documentation about them.

### How

To do that a target `help` was added to Makefile with an `awk` command that parses the file searching for a regular expression that represents targets.


caiordjesus (approved):
AWESOME GOOD TO ME!

PR #292, opened 2022-02-17
Create first version of maestro image with single entrypoint

guilhermocc:
## What ❓
Create a new dockerfile for maestro and update e2e setup

## Why 🤔
Now that maestro has a single entrypoint, we need to create the new docker image and update e2e tests setup to use this new image.

## How 👀 
- Create new dockerfile for maestro using multi-stage build
- Create dockerfile for maestro e2e tests without building the project (to optimize e2e tests preparation time)
- Update config ports

## Obs 🗒️ 
⚠️ ⚠️ ⚠️  This PR depends on https://github.com/topfreegames/maestro/pull/289 ⚠️ ⚠️ ⚠️


lyracampos (approved):

Dockerfile
```diff
@@ -18,26 +18,36 @@
...
+
+ARG maestroComponent
+ENV MAESTRO_COMPONENT=$maestroComponent
 
 WORKDIR /app
 
-COPY ./bin/management-api-linux-x86_64 /app/management-api
-COPY ./config /app/config
+COPY --from=build-env /app/maestro /app/maestro
+COPY --from=build-env /app/config /app/config
```
caiordjesus, line 49: Why do we need to copy the config folder if it only have local config?

caiordjesus (approved):

PR #295, opened 2022-02-21
Add parameters name, game and version to list schedulers route

arthur29:
### Why
Currently the route `/schedulers` doesn't accept any argument. But it is needed to filter `schedulers` by game.

### How
This PR proposes to change the `schedulers.proto` adding URL parameters to the `ListSchedulers` method and changing `handlers/schedulers_handler.go` to pass these parameters as filters to the `GetSchedulersWithFilters` method.

### What
This PR adds three URL parameters to the route `/schedulers` `name` that is the scheduler name, `game` that is the scheduler `game`, and `version` that is the scheduler `version`.


internal/api/handlers/schedulers_handler.go
```diff
@@ -53,7 +54,12 @@ func ProvideSchedulersHandler(schedulerManager *scheduler_manager.SchedulerManag
 }
 
 func (h *SchedulersHandler) ListSchedulers(ctx context.Context, message *api.ListSchedulersRequest) (*api.ListSchedulersResponse, error) {
-	schedulerEntities, err := h.schedulerManager.GetSchedulersWithFilter(ctx, nil)
+	schedulerFilter := &filters.SchedulerFilter{
+		Name:    message.Name,
```
caiordjesus, line 58: WDYT about using "message.GetName()" instead of directly accessing the property? It's a good practice since these "GetX()" methods are auto-generated and take care of nil values and stuff. If you look at the other handlers, they're using this pattern too 🤟

internal/api/handlers/schedulers_handler_test.go
```diff
@@ -55,16 +57,20 @@ import (
 
 func TestListSchedulers(t *testing.T) {
```
caiordjesus, line 58: WDYT about adding a test case where the "GetSchedulersWithFilter" method returns an error?

internal/api/handlers/schedulers_handler.go
arthur29: Good catch, thank you!

caiordjesus (approved):
Clean!

guilhermocc (approved):

PR #297, opened 2022-02-22
fix(kube config): checks if inCluster

caiordjesus:
### Why?
To deploy maestro next, we have to check if we are inCluster


arthur29 (approved):
LGTM

PR #298, opened 2022-02-22
Feature/add active version in version listing

guilhermocc:
## What ❓ 
Add active version field in scheduler version listing.

## Why 🤔 
Currently, this endpoint does not inform what is the active scheduler, since now an active scheduler can be any version, we need to include this information in the endpoint response.

## How
- Update SchedulerStorage postgres adapter
- Update ListScheduler message in proto definition
- Update SchedulerHandler and tests


caiordjesus (approved):
LGTM

internal/adapters/scheduler_storage/pg/scheduler_version.go
```diff
@@ -29,12 +29,14 @@ import (
 
 type SchedulerVersion struct {
 	Version   string      `db:"version"`
+	IsActive  bool        `db:"is_active"`
```
lyracampos, line 32: @guilhermocc IsActive(is_active) is just a logical property, so we won't have a column in the database to represent it, right?

lyracampos (approved):

lyracampos:
👍

internal/adapters/scheduler_storage/pg/scheduler_version.go
guilhermocc: exactly! in the query, we use an alias `is_active` for this calculated query result, so I had to add this annotation here

PR #299, opened 2022-02-22
feature/scheduler info services

lyracampos:
**What?**
We've implemented a new feature in Maestro to list information from Schedulers such as name, game and current state, also information from Game Rooms such as total from ready, occupied, pending and terminating rooms.

**Why?**
For now this feature is to attend a demand from Maestro Next and Hub Dashboard integration.

**How?**
- Creating use case in domain to list Schedulers and Game Rooms information.  The purpose of this use case is to merge data from Schedulers (postgres) with data from Game Rooms (redis).
- Creating new endpoint in management-api to expose this functionality.

to list schedulers information by game:
`http://localhost:8080/schedulers/info?game=battleroyale`

to list information for all schedulers
`http://localhost:8080/schedulers/info`


caiordjesus (approved):

internal/core/entities/scheduler_info.go
```diff
@@ -0,0 +1,37 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+package entities
+
+type SchedulerInfo struct {
+	Name             string
+	Game             string
+	State            string
+	RoomsReady       int64
```
guilhermocc, line 29: @lyracampos why aren't we just using `int` here? according to go tour: "When you need an integer value you should use `int` unless you have a specific reason to use a sized or unsigned integer type." https://go.dev/tour/basics/11

guilhermocc (approved):

PR #300, opened 2022-02-23
e2e refactor utils to return scheduler instead scheduler name

arthur29:
### Why
Currently, e2e utils functions return the scheduler name, but we need other fields to use along with tests.

### What
Return entire Scheduler in utils functions.

### How
Refactoring utils functions to return a scheduler.
Refactoring tests to use scheduler instead of scheduler name.


lyracampos (approved):
👍

guilhermocc (approved):

PR #301, opened 2022-02-23
Return error when removing rooms fails

guilhermocc:
## What ❓ 
Return error in remove_room operation executor when some error occurs while deleting game rooms.

## Why 🤔 
Currently, the remove_rooms operation does not fail if some error occurred while removing rooms, this behavior makes the visibility worse since we are "masking" that the execution that doesn't succeed.

## How
- Update executor logic and tests


lyracampos (approved):
Maestrinho ❤️ ❤️ ❤️

caiordjesus (approved):
MESTRINHO!!! <3 <3

PR #303, opened 2022-02-23
chore(rooms ports): unify room ports on single file

caiordjesus:
### What?
Unify room ports on a single file (room_storage, room manager, etc).
### Why?
Today we have ports spread across Maestro Next. Through various files, we find different interfaces. That said, for each one of them we need to specify a mock file. Unifying those ports makes it easier to auto-generate mocks, use those mocks and respect the hexagonal architecture (having interfaces under the services folder is weird).
### How?
Migrating interfaces from others room_x files to room_ports.go and generating mock for this new file. Also, removing unnecessary older mocks files (such as room_storage mock that was under "adapters" folder).


lyracampos (commented):
@caiordjesus take a look everywhere in the project that are injecting the RoomManager service and change to the RoomManager port.

For example in rooms_handler

internal/core/ports/room_ports.go
```diff
@@ -24,11 +24,23 @@ package ports
...
 import (
 	"context"
+	"sync"
 	"time"
 
+	"github.com/topfreegames/maestro/internal/core/entities"
 	"github.com/topfreegames/maestro/internal/core/entities/game_room"
 )
 
+type RoomManager interface {
```
guilhermocc, line 34: @caiordjesus what do you think about separating the port types with comments as we did on operations ports?
```
// Primary ports (input, driving ports)
...
// Secondary ports (output, driven ports)
...
```

internal/core/ports/room_ports.go
```diff
@@ -24,11 +24,23 @@ package ports
...
+	DeleteRoomAndWaitForRoomTerminated(ctx context.Context, gameRoom *game_room.GameRoom) error
```
guilhermocc, line 35: Also, could we add comments describing these interface methods as we are doing with other primaries/secondary ports?

guilhermocc (approved):

PR #304, opened 2022-02-24
Fix/make add rooms fail on error

guilhermocc:
## What ❓ 
Update add_rooms operation executor to fail if there is some error while creating new rooms. Also implements OnError method to clean previously created rooms if the operation fails.

## Why 🤔 
Currently, the add_rooms operation does not fail if some error occurred while creating rooms, this behavior makes the visibility worse since we are "masking" that the execution that doesn't succeed.

## How 👀 
- Implement OnError method to clean created rooms
- Use errgroup to fail the operation if some error occurs while the rooms are being created
- Update unitary and e2e test cases
- Remove necessary waitGroup from switch_active_version operation executor


lyracampos (approved):
👍

PR #307, opened 2022-03-02
Move events service interface to ports folder

guilhermocc:
## What ❓ 
Move events service interface to ports folder.

## Why 🤔 
Following the ports refactor we are doing, we will no longer have the `interfaces` package in services folder, so we need to do this refactor with events service.

## How 👀 
- Simple package refactor for event service, agregate events ports and adapters
image
image

- Agregate operation adapters
image


lyracampos (approved):
👍

caiordjesus (approved):
LGTM!

PR #308, opened 2022-03-03
Aggregate scheduler ports, refactor scheduler adapters

guilhermocc:
## What ❓ 
Aggregate scheduler ports and refactor scheduler adapters into a single folder.

## Why 🤔 
Following the ports refactor we are doing, we will no longer have the interfaces package in services folder, so we need to do this refactor with scheduler manager.

## Obs ✍️ 
Another PR will be opened to update tests that are using SchedulerManager implementation without mock.

## How 👀 
image
image


lyracampos (approved):
👍

PR #309, opened 2022-03-03
chore(runtime watcher, rooms adapters): troubleshooting improvements

caiordjesus:
### What?
Improve watcher logs and rooms adapters methods.
### Why?
When deployed, runtime watcher is not behaving as expected. Since reproducing the same bugs locally is not always easy, we want to be able to better troubleshoot what is happening.
### How?
Generating more log messages and improving runtime-related methods on rooms adapters to handle errors.

### DIFF
--- a/internal/core/workers/runtime_watcher_worker/runtime_watcher_worker.go
+++ b/internal/core/workers/runtime_watcher_worker/runtime_watcher_worker.go
@@ -93,13 +93,16 @@ func (w *runtimeWatcherWorker) IsRunning() bool {
 }
 
 func (w *runtimeWatcherWorker) processEvent(ctx context.Context, event game_room.InstanceEvent) error {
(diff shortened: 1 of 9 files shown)


internal/core/workers/runtime_watcher_worker/runtime_watcher_worker.go
```diff
@@ -93,13 +93,16 @@ func (w *runtimeWatcherWorker) IsRunning() bool {
 }
 
 func (w *runtimeWatcherWorker) processEvent(ctx context.Context, event game_room.InstanceEvent) error {
+	w.logger.Sugar().Infof("processing event: %++v", event)
```
guilhermocc, line 96: @caiordjesus why `%++v`?

guilhermocc (approved):

PR #310, opened 2022-03-03
Feature/implement create new version executor OnError method

guilhermocc:
## What ❓ 
Implement `OnError` method for new_scheduler_version operation executor. If some error occurs in its execution, the `OnError` method will ensure that any room created for validating new scheduler version will be deleted.

## Why 🤔 
Currently, we are not dealing with the case in which the operation is canceled.

## How 👀 
- Delete validation method from room manager
- Update new_scheduler_version executor OnError method


caiordjesus (approved):
LGTM!

PR #312, opened 2022-03-04
chore(validations): create validation for kube resource name

caiordjesus:
### What?
- Create validation methods for kube resources names.
### Why?
- Since containers and schedulers are completely related to Kubernetes pods and namespaces, we need to validate if they follow the kubernetes rules for naming, otherwise Maestro will not be able to create those resources, resulting in an error.
### How?
- Following the RFC 1123 through kubernetes documentation. https://kubernetes.io/docs/concepts/overview/working-with-objects/names/

### DIFF
--- a/internal/core/validations/validations.go
+++ b/internal/core/validations/validations.go
@@ -23,6 +23,7 @@
 package validations
 
 import (
+	"regexp"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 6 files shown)


internal/core/validations/validations.go
```diff
@@ -78,3 +79,31 @@ func IsVersionValid(version string) bool {
 	_, err := semver.NewVersion(version)
 	return err == nil
 }
+
+// IsNameValidOnKube check if name is valid for kubernetes resources (RFC 1123)
+// Since regexp does not support lookbehind, we test directly the last character before
+// matching the string name.
+func IsNameValidOnKube(name string) bool {
```
lyracampos, line 86: @caiordjesus I think that we can assume some pattern here like the previous validations which use _**is/are**_ as prefix,  _**field**_ and then _**action**_ as suffix

examples:
Is MaxSurge _**Valid**_
Is ImagePullPolicy _**Supported**_
Is Protocol _**Supported**_
Is Version _**Valid**_
Is Name _**Valid**_ OnKube

WDYT about use IsKubeResourceNameValid?

lyracampos (approved):
👍

PR #315, opened 2022-03-04
Change CreateScheduler and NewSchedulerVersion contracts to reflect entities

arthur29:
### Why
`CreateScheduler` and `NewSchedulerVersion` have different request contracts than `GetScheduler` and `ListSchedulers` responses. 
### What
Unify contracts between all scheduler routes.
### How
Change `schedulers.proto` and `messages.proto` and reflect all changes on the code.


guilhermocc (approved):
@arthur29 the fix for get scheduler not bringing back the forwarder will be implemented in the next PRs?

PR #318, opened 2022-03-04
validate if scheduler and operations exists when try to cancel operation

lyracampos:
### What?
Validating if the scheduler and the operation exist to be cancelled

### Why?
Because when user try to cancel and inexistent operation we are returning 200 instead of 404 when scheduler or operation doesn't exists from storage.

### How?
Implementing scheduler and operation validation and return error if someone doesn't found from storage
Implementing unit tests that didn't exist before


guilhermocc (approved):

arthur29 (approved):
LGTM

PR #321, opened 2022-03-07
adds upsert instance in create room method

caiordjesus:
### What?
Adds upsert instance method on `CreateRoomAndWaitForReadiness`.

### Why?
We were relaying on watcher to call this method, but there's no specific reason to not do that at this time (with the instance creation)

### How?
Calling the method `UpsertInstance` from `instanceStorage` right after the creation of the instance


arthur29 (approved):
LGTM

guilhermocc (approved):

PR #327, opened 2022-03-08
New major version: validation room should not be forwarded

caiordjesus:
### What?
Don't forward validation room pings.
### Why?
Because during an update, a match could try to use validation rooms
### How?
Storing a flag with the game room on Redis, and fetching this room before producing event (inside produce event method). If game room have flag true, event is not produced

### DIFF
--- a/e2e/suites/management/create_new_scheduler_version_test.go
+++ b/e2e/suites/management/create_new_scheduler_version_test.go
@@ -129,13 +129,10 @@ func TestCreateNewSchedulerVersion(t *testing.T) {
 		t.Run("Should Succeed - create major change, all pods are changed", func(t *testing.T) {
 			t.Parallel()
 
-			scheduler, err := createSchedulerWithRoomsAndWaitForIt(t, maestro, managementApiClient, game, kubeClient)
-
-			podsBeforeUpdate, err := kubeClient.CoreV1().Pods(scheduler.Name).List(context.Background(), metav1.ListOptions{})
-			require.NoError(t, err)
+			schedulerName, roomNameBeforeUpdate := createSchedulerWithForwarderAndRooms(t, maestro, kubeClient, managementApiClient, maestro.ServerMocks.GrpcForwarderAddress)
 
 			updateRequest := &maestroApiV1.NewSchedulerVersionRequest{
-				Name:     scheduler.Name,
+				Name:     schedulerName,
 				Game:     "test",
 				MaxSurge: "10%",
 				Spec: &maestrov1.Spec{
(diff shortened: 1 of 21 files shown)


e2e/suites/management/create_new_scheduler_version_test.go
```diff
@@ -181,6 +183,53 @@ func TestCreateNewSchedulerVersion(t *testing.T) {
...
+			// Wait till we have 3 pods (the third one is the validation one). Then, try to forward event.
+			// Since we don't forward event from validation room, it should return true even though we're not
+			// mocking the gRPC call.
+			require.Eventually(t, func() bool {
+				pods, err := kubeClient.CoreV1().Pods(scheduler.Name).List(context.Background(), metav1.ListOptions{})
+				require.NoError(t, err)
+				require.NotEmpty(t, pods.Items)
+
+				if len(pods.Items) == 3 {
+					validationRoom := pods.Items[2].ObjectMeta.Name
```
guilhermocc, line 195: @caiordjesus the validation pod will always be the third pod in this listing?

internal/core/operations/add_rooms/add_rooms_executor_test.go
```diff
@@ -93,7 +93,7 @@ func TestAddRoomsExecutor_Execute(t *testing.T) {
 		roomsManager := mockports.NewMockRoomManager(mockCtrl)
 
 		schedulerStorage.EXPECT().GetScheduler(context.Background(), op.SchedulerName).Return(&scheduler, nil)
-		roomsManager.EXPECT().CreateRoomAndWaitForReadiness(gomock.Any(), gomock.Any()).Return(&gameRoom, &gameRoomInstance, nil).Times(10)
+		roomsManager.EXPECT().CreateRoomAndWaitForReadiness(gomock.Any(), gomock.Any(), gomock.Any()).Return(&gameRoom, &gameRoomInstance, nil).Times(10)
```
guilhermocc, line 96: @caiordjesus i think we can use the false value in this mock expectation since we don't expect any add room to be a validation one, right?

internal/core/operations/newschedulerversion/new_scheduler_version_executor_test.go
```diff
@@ -78,7 +78,7 @@ func TestCreateNewSchedulerVersionExecutor_Execute(t *testing.T) {
 		newSchedulerWithNewVersion.Spec.Version = "v2.0.0"
 		newSchedulerWithNewVersion.RollbackVersion = "v1.0.0"
 
-		mocksForExecutor.roomManager.EXPECT().CreateRoomAndWaitForReadiness(gomock.Any(), gomock.Any()).Return(&game_room.GameRoom{ID: "id-1"}, nil, nil)
+		mocksForExecutor.roomManager.EXPECT().CreateRoomAndWaitForReadiness(gomock.Any(), gomock.Any(), gomock.Any()).Return(&game_room.GameRoom{ID: "id-1"}, nil, nil)
```
guilhermocc, line 81: Here we can also use the real value `true` for this expectation, this way we have more confident tests, it will break if change the implementation and send a false value

internal/core/operations/switch_active_version/switch_active_version_executor_test.go
```diff
@@ -116,7 +116,7 @@ func TestSwitchActiveVersionOperation_Execute(t *testing.T) {
 				Status:      game_room.GameStatusReady,
 				LastPingAt:  time.Now(),
 			}
-			mocks.roomManager.EXPECT().CreateRoomAndWaitForReadiness(gomock.Any(), gomock.Any()).Return(gameRoom, nil, nil)
+			mocks.roomManager.EXPECT().CreateRoomAndWaitForReadiness(gomock.Any(), gomock.Any(), gomock.Any()).Return(gameRoom, nil, nil)
```
guilhermocc, line 119: same thing about expectations here

internal/core/services/events_forwarder/events_forwarder_service.go
```diff
@@ -63,12 +65,23 @@ func NewEventsForwarderService(
...
 }
 
 func (es *EventsForwarderService) ProduceEvent(ctx context.Context, event *events.Event) error {
+	gameRoom, err := es.roomStorage.GetRoom(ctx, event.SchedulerID, event.RoomID)
+	if err != nil {
+		es.logger.Error(fmt.Sprintf("cannot produce event since room \"%s\" is not registered on storage for scheduler \"%s\"", event.RoomID, event.SchedulerID))
+		return err
+	}
+
+	if gameRoom.IsValidationRoom {
```
guilhermocc, line 81: Nice!

guilhermocc (approved):
LGTM! Just need to check my comments regarding an improvement in tests 💅

e2e/suites/management/create_new_scheduler_version_test.go
```diff
@@ -181,6 +183,53 @@ func TestCreateNewSchedulerVersion(t *testing.T) {
 			require.NoError(t, err)
 			require.NotNil(t, updateResponse.OperationId, scheduler.Name)
 
+			// Wait till we have 3 pods (the third one is the validation one). Then, try to forward event.
+			// Since we don't forward event from validation room, it should return true even though we're not
+			// mocking the gRPC call.
+			require.Eventually(t, func() bool {
```
guilhermocc, line 189: @caiordjesus i just changed the index of the pod we test to 1 (second pod) and the test is still passing... With further investigation I noticed that the scheduler we are creating in this test does not have the forwarder configured, this test will always pass. We need to add the forwarder.

PR #332, opened 2022-03-10
Fix/maintain scheduler version hierarchy on switch version

guilhermocc:
## What ❓
Create methods for calculating new major and minor versions, considering that the versions hierarchy should be maintained even if we are creating new versions from previous versions and not the most recent one. 

## Why 🤔 
If we switch to a previous scheduler version and then create a minor or major version, maestro is overriding higher versions and not dealing with this scenario correctly.

We have to fix the following scenarios:

- We have three versions, v1.0.0(active) and v1.1.0 and v2.0.0. If i create a new minor version:
  - What is happening: maestro overrides v1.1.0
  - What is expected to happen: maestro creates a new version v1.2.0

- We have three versions, v1.0.0(active), v1.1.0 and v2.0.0. If i create a new major version:
  - What is happening: maestro overrides v2.0.0
  - What is expected to happen: maestro create a new major version v3.0.0

## How 👀 
- Create and unify `create_new_scheduler_version` e2e tests, including both new scenarios described above.
- Implement a method for calculating the new major version and new minor version, always considering what is the greatest minor and major version.
- Implement fix for the `IsMajorVersion` method from scheduler, it was not ignoring the spec `version` field (that will be different in some scenarios but should not be considered in the comparison), and the containers HostPort, which is a field populated by maestro and should also not be considered in the comparison. 
- Improve switch active version e2e tests to ensure that a change from minor to minor does not replace pods

### DIFF
--- a/e2e/suites/management/create_new_scheduler_version_test.go
+++ b/e2e/suites/management/create_new_scheduler_version_test.go
@@ -27,6 +27,9 @@ import (
 	"fmt"
 	"time"
 
+	_struct "github.com/golang/protobuf/ptypes/struct"
+	"google.golang.org/protobuf/types/known/structpb"
+
 	"testing"
 
 	maestroApiV1 "github.com/topfreegames/maestro/pkg/api/v1"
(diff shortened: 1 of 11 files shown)


e2e/suites/management/create_new_scheduler_version_test.go
```diff
@@ -45,11 +45,13 @@ import (
 func TestCreateNewSchedulerVersion(t *testing.T) {
 	game := "create-new-scheduler-version-game"
 	framework.WithClients(t, func(roomsApiClient *framework.APIClient, managementApiClient *framework.APIClient, kubeClient kubernetes.Interface, redisClient *redis.Client, maestro *maestro.MaestroInstance) {
-		t.Run("Should Succeed - create minor version, no pods replaces", func(t *testing.T) {
+		t.Run("Should Succeed - create minor version, no pods replaces, create major version, all pods are changed", func(t *testing.T) {
```
caiordjesus, line 48: Insane gameplay mechanics on e2e holy christ

caiordjesus (approved):
God mode PR!
WDYT about adding more logs to improve traceability of the operation? Maybe even debug logs could help us troubleshoot for the king of calculation the executor is doing (like from which version to what version and why). "Change on [x], major version change"

e2e/suites/management/create_new_scheduler_version_test.go
guilhermocc: For testing the new scenarios i had to basically create a minor and a major change to setup environment, so I thought it wouldn't make sense to have 2 tests basically testing the same thing

PR #333, opened 2022-03-10
Review debug logs

caiordjesus:
### Why?
Having some logs as "Debug" makes it harder to troubleshoot in production. Because of this, some logs we wrote as "debug" should be reviewed.


arthur29 (approved):
LGTM

PR #337, opened 2022-03-11
chore - clear cache when update scheduler

caiordjesus:
### What?
Clear scheduler cache when updating scheduler

### Why?
The **EventsForwarderService** uses a cache to have quick access to a scheduler. This is important because, in a real scenario, we could have a scheduler with thousands of rooms sending pings, and to retrieve forwarder info, we would have to access Postgres for each one of them.
However, since we don't clear cache when we create/switch to a new version, it would take a long time to refresh the cache and reflect the update to the forwarder.

### Missing:
Implement validation on e2e that the scheduler updated

### DIFF
--- a/e2e/suites/management/switch_active_version_test.go
+++ b/e2e/suites/management/switch_active_version_test.go
@@ -45,8 +45,6 @@ import (
 func TestSwitchActiveVersion(t *testing.T) {
 	game := "switch-active-version-game"
 
-
-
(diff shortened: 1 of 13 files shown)


e2e/suites/management/switch_active_version_test.go
```diff
@@ -304,7 +309,7 @@ func TestSwitchActiveVersion(t *testing.T) {
 			}
 		})
 
-		t.Run("Should fail - version does not exist", func(t *testing.T) {
+		t.Run("fail - version does not exist", func(t *testing.T) {
```
guilhermocc, line 312: Lowercase first letter or Upper?

guilhermocc (approved):

PR #339, opened 2022-03-14
Feature/adding forwarder type validation

henriqueoelze:


internal/core/validations/validations.go
```diff
@@ -23,6 +23,7 @@
 package validations
 
 import (
+	"github.com/topfreegames/maestro/internal/core/entities/forwarder"
```
guilhermocc, line 26: need to run `make goimports`

caiordjesus (approved):

internal/validations/validations.go
```diff
@@ -75,6 +75,12 @@ func RegisterValidations() error {
 	}
 	addTranslation(Validate, "kube_resource_name", "{0} must follow the public RFC 1123 about naming conventions")
 
+	err = Validate.RegisterValidation("forwarder_type", forwarderTypeValidate)
+	if err != nil {
+		return errors.New("could not register forwarderTypeValidate")
+	}
+	addTranslation(Validate, "forwarder_type", "{0} must be one of the following options: gRPC")
```
arthur29, line 82: Hey @henriqueoelze what do you think to make this message less static, creating it from all forwarder types?

guilhermocc (approved):

internal/validations/validations.go
henriqueoelze: It is already with all the forwarder types, only gRPC is available 😄

PR #340, opened 2022-03-14
fix - Unexpected behavior when trying to cancel operation on final state

caiordjesus:
### What?
Adds treatment for when management-API receives a request to cancel an operation that cannot be canceled.

### Why?
Right now, when we try to cancel an operation that is already finished, Maestro does not check the actual status of the operation before trying to execute the cancelation. Because of that, the logs produced are related to the operation not having a cancel function, and not to the status.
Also, we don't see logs at the management-API relating to what is going on.


arthur29 (approved):
LGTM

PR #343, opened 2022-03-15
Add in Operation entity ExecutionHistory and Input fields

arthur29:
### Why
The Operation entity doesn't have the information about which was its input and when and which events occurred during operation execution.

### What
Add in Operation entity the fields `ExecutionHistory` and `Input` to serve this information. And change operation_handler to reflect these changes in the entity.


internal/api/handlers/operations_handler.go
```diff
@@ -141,27 +143,55 @@ func (h *OperationsHandler) fromOperationsToResponses(entities []*operation.Oper
...
+	}
+
+	var err error
+	operation.Status, err = entity.Status.String()
 	if err != nil {
 		return nil, fmt.Errorf("failed to convert operation entity to response: %w", err)
 	}
+
+	operation.ExecutionHistory = h.fromOperationEventsToResponse(entity.ExecutionHistory)
+	if err != nil {
```
guilhermocc, line 160: @arthur29 this error check here doesn't make sense since fromOperationEventsToResponse does not return any error, this if clause will never be reached since we are already checking this error above.

guilhermocc (approved):

caiordjesus (approved):

guilhermocc (approved):

PR #344, opened 2022-03-16
Fix/main e2e tests

guilhermocc:
## What ❓ 
Fix e2e tests, ensure migration is always executed with success

## Why 🤔 
We had intermittent e2e tests, and we were not ensuring the database migration was being executed with success


arthur29 (approved):
LGTM

PR #345, opened 2022-03-16
Adds update operation execution history method to operation storage adapter

caiordjesus:
### What?
Adds new `UpdateOperationExecutionHistory` method for Operation Storage.
### Why?
We need a method to update an operation execution history, so when we fetch the operation, it will have this information stored.


internal/core/ports/operation_ports.go
```diff
@@ -88,6 +88,8 @@ type OperationStorage interface {
 	// UpdateOperationStatus only updates the operation status for the given
 	// operation ID.
 	UpdateOperationStatus(ctx context.Context, schedulerName, operationID string, status operation.Status) error
+	// UpdateOperationExecutionHistory updates the operation execution history in operation storage
```
guilhermocc, line 91: The name of the interface is OperationStorage, I don't think we need to mention that it updates in the operation storage here, right?

internal/adapters/operation/operation_storage_redis_test.go
```diff
@@ -248,6 +249,64 @@ func TestUpdateOperationStatus(t *testing.T) {
...
+
+		op := &operation.Operation{
+			ID:            "some-op-id",
+			SchedulerName: "test-scheduler",
+			Status:        operation.StatusPending,
+		}
+		client.Close()
+
+		err := storage.UpdateOperationExecutionHistory(context.Background(), op)
+		require.Error(t, err)
```
guilhermocc, line 305: Hey @caiordjesus, WDYT about asserting the error message?
```		require.EqualError(t, err, "failed to update operation execution history: redis: client is closed")
```

internal/adapters/operation/operation_storage_redis_test.go
```diff
@@ -248,6 +249,64 @@ func TestUpdateOperationStatus(t *testing.T) {
...
+
+		operationStored, err := client.HGetAll(context.Background(), storage.buildSchedulerOperationKey(op.SchedulerName, op.ID)).Result()
+		require.NoError(t, err)
+
+		execHist := []operation.OperationEvent{}
+		err = json.Unmarshal([]byte(operationStored[executionHistory]), &execHist)
+		require.NoError(t, err)
+
+		for i := range op.ExecutionHistory {
+			require.Equal(t, op.ExecutionHistory[i].Event, execHist[i].Event)
```
guilhermocc, line 286: I think we should use `assert.Equal` here

guilhermocc (approved):

PR #346, opened 2022-03-16
Feature/create metrics reporter worker

guilhermocc:
## What ❓ 
Create metrics reporter worker implementation, including unitary tests.

## Why 🤔 
We are going to have a new maestro component that will need to use a metrics reporter worker to produce periodic metrics.


internal/core/workers/metricsreporter/metrics_reporter_worker.go
```diff
@@ -0,0 +1,163 @@
...
+	pendingRooms, err := w.roomStorage.GetRoomCountByStatus(w.workerContext, w.schedulerName, game_room.GameStatusPending)
+	if err != nil {
+		w.logger.Error("Error getting pending pods", zap.Error(err))
+	} else {
+		reportGameRoomPendingNumber(w.schedulerName, pendingRooms)
+	}
+	occupiedRooms, err := w.roomStorage.GetRoomCountByStatus(w.workerContext, w.schedulerName, game_room.GameStatusOccupied)
+	if err != nil {
+		w.logger.Error("Error getting occupied pods", zap.Error(err))
+	} else {
```
caiordjesus, line 142: What is your opinion about implementing tiny functions instead of going to the if-else list approach? I think it makes it harder to read/debug a big function with very similar conditions. Also, since it happens in the else condition, my first reaction was to think that it's not the expected behavior
Like:
```
func (w *MetricsReporterWorker) reportGameRoomMetrics() {
        w.reportReadyRooms()
        w.reportPendingRooms()
        w.reportTerminatingRooms()
        ...
}

func (w *MetricsReporterWorker) reportReadyRooms () {
        readyRooms, err := w.roomStorage.GetRoomCountByStatus(w.workerContext, w.schedulerName, game_room.GameStatusReady)
	if err != nil {
		w.logger.Error("Error getting ready pods", zap.Error(err))
		return
	}
	reportGameRoomReadyNumber(w.schedulerName, readyRooms)
}
...
```

internal/core/workers/worker.go
```diff
@@ -44,10 +45,13 @@ type Worker interface {
...
-	OperationExecutors map[string]operations.Executor
-	RoomManager        ports.RoomManager
-	Runtime            ports.Runtime
+	OperationManager              ports.OperationManager
+	OperationExecutors            map[string]operations.Executor
+	RoomManager                   ports.RoomManager
+	Runtime                       ports.Runtime
+	RoomStorage                   ports.RoomStorage
+	InstanceStorage               ports.GameRoomInstanceStorage
+	MetricsReporterIntervalMillis time.Duration
```
caiordjesus, line 54: Why did we decide to use this `MetricsReporterIntervalMillis` instead of using the `config` approach we are using on managers? It feels to me that if we want to add new configuration variables it'll be harder

caiordjesus (approved):

internal/core/workers/metricsreporter/metrics_reporter_worker.go
guilhermocc: I like this approach!

PR #351, opened 2022-03-18
Feat/populate operation execution history

arthur29:
### Why
In the Operation entity, there is a field called `ExecutionHistory` that has a list of events and it isn't populated yet. This PR proposes to during operation_execution_worker populate this field.

### What
* Create operation_manager `AppendOperationEventToExecutionHistory` function
* Add events during worker execution

### DIFF
--- a/internal/adapters/operation/operation_storage_redis.go
+++ b/internal/adapters/operation/operation_storage_redis.go
@@ -43,7 +43,7 @@ const (
 	definitionNameRedisKey     = "definitionName"
 	createdAtRedisKey          = "createdAt"
 	definitionContentsRedisKey = "definitionContents"
-	executionHistory           = "executionHistory"
+	executionHistoryRedisKey   = "executionHistory"
 )
 
 var _ ports.OperationStorage = (*redisOperationStorage)(nil)
@@ -62,13 +62,19 @@ func NewRedisOperationStorage(client *redis.Client, clock ports.Clock) *redisOpe
 // CreateOperation marshal and pushes the operation to the scheduler pending
 // operations list.
 func (r *redisOperationStorage) CreateOperation(ctx context.Context, op *operation.Operation) error {
-	err := r.client.HSet(ctx, r.buildSchedulerOperationKey(op.SchedulerName, op.ID), map[string]interface{}{
+	executionHistoryJson, err := json.Marshal(op.ExecutionHistory)
+	if err != nil {
+		return errors.NewErrUnexpected("failed to create operation on redis").WithError(err)
+	}
+
+	err = r.client.HSet(ctx, r.buildSchedulerOperationKey(op.SchedulerName, op.ID), map[string]interface{}{
 		idRedisKey:                 op.ID,
 		schedulerNameRedisKey:      op.SchedulerName,
 		statusRedisKey:             strconv.Itoa(int(op.Status)),
 		definitionNameRedisKey:     op.DefinitionName,
 		createdAtRedisKey:          op.CreatedAt.Format(time.RFC3339Nano),
 		definitionContentsRedisKey: op.Input,
+		executionHistoryRedisKey:   executionHistoryJson,
 	}).Err()
 
 	if err != nil {
@@ -88,6 +94,12 @@ func (r *redisOperationStorage) GetOperation(ctx context.Context, schedulerName,
 		return nil, errors.NewErrNotFound("operation %s not found in scheduler %s", operationID, schedulerName)
 	}
 
+	var executionHistory []operation.OperationEvent
+	err = json.Unmarshal([]byte(res[executionHistoryRedisKey]), &executionHistory)
+	if err != nil {
+		return nil, errors.NewErrEncoding("failed to parse operation execution history").WithError(err)
+	}
+
 	statusInt, err := strconv.Atoi(res[statusRedisKey])
 	if err != nil {
 		return nil, errors.NewErrEncoding("failed to parse operation status").WithError(err)
@@ -99,12 +111,13 @@ func (r *redisOperationStorage) GetOperation(ctx context.Context, schedulerName,
 	}
 
 	op := &operation.Operation{
-		ID:             res[idRedisKey],
-		SchedulerName:  res[schedulerNameRedisKey],
-		DefinitionName: res[definitionNameRedisKey],
-		CreatedAt:      createdAt,
-		Status:         operation.Status(statusInt),
-		Input:          []byte(res[definitionContentsRedisKey]),
+		ID:               res[idRedisKey],
+		SchedulerName:    res[schedulerNameRedisKey],
+		DefinitionName:   res[definitionNameRedisKey],
+		CreatedAt:        createdAt,
+		Status:           operation.Status(statusInt),
+		Input:            []byte(res[definitionContentsRedisKey]),
+		ExecutionHistory: executionHistory,
 	}
 
 	return op, nil
@@ -141,7 +154,7 @@ func (r *redisOperationStorage) UpdateOperationExecutionHistory(ctx context.Cont
 		return errors.NewErrUnexpected("failed to marshal operation execution history").WithError(err)
 	}
 	err = r.client.HSet(ctx, r.buildSchedulerOperationKey(op.SchedulerName, op.ID), map[string]interface{}{
-		executionHistory: jsonExecutionHistory,
+		executionHistoryRedisKey: jsonExecutionHistory,
 	}).Err()
 
 	if err != nil {
--- a/internal/core/ports/operation_ports.go
+++ b/internal/core/ports/operation_ports.go
@@ -59,7 +59,9 @@ type OperationManager interface {
 	// RevokeLease revokes the lease of the given operation.
 	RevokeLease(ctx context.Context, operation *operation.Operation) error
 	// StartLeaseRenewGoRoutine starts an async process to keep renewing the lease of the given operation.
-	StartLeaseRenewGoRoutine(operationCtx context.Context, op *operation.Operation)
+	StartLeaseRenewGoRoutine(ctx context.Context, op *operation.Operation)
+	// AppendOperationEventToExecutionHistory add operation event to execution history
+	AppendOperationEventToExecutionHistory(ctx context.Context, op *operation.Operation, eventMessage string) error
 }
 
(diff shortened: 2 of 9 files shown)


internal/adapters/operation/operation_storage_redis.go
```diff
@@ -43,7 +43,7 @@ const (
 	definitionNameRedisKey     = "definitionName"
 	createdAtRedisKey          = "createdAt"
 	definitionContentsRedisKey = "definitionContents"
-	executionHistory           = "executionHistory"
+	executionHistoryRedisKey   = "executionHistory"
```
caiordjesus, line 46: Very nice refactor!! Didn't see we were going off the pattern with this guy!

internal/core/ports/operation_ports.go
```diff
@@ -59,7 +59,9 @@ type OperationManager interface {
 	// RevokeLease revokes the lease of the given operation.
 	RevokeLease(ctx context.Context, operation *operation.Operation) error
 	// StartLeaseRenewGoRoutine starts an async process to keep renewing the lease of the given operation.
-	StartLeaseRenewGoRoutine(operationCtx context.Context, op *operation.Operation)
+	StartLeaseRenewGoRoutine(ctx context.Context, op *operation.Operation)
+	// AppendOperationEventToExecutionHistory add operation event to execution history
```
caiordjesus, line 63: What's your opinion on adding `and updates storage`? In my mind it sound relevant to what the method does

internal/core/workers/operation_execution_worker/operation_execution_worker.go
```diff
@@ -89,14 +89,25 @@ func (w *OperationExecutionWorker) Start(ctx context.Context) error {
...
 		executor, hasExecutor := w.executorsByName[def.Name()]
 		if !hasExecutor {
 			loopLogger.Warn("operation definition has no executor")
+
 			w.evictOperation(ctx, loopLogger, op)
 			reportOperationEvicted(w.schedulerName, op.DefinitionName, LabelNoOperationExecutorFound)
+
+			err = w.operationManager.AppendOperationEventToExecutionHistory(ctx, op, "Operation definition has no executor")
+			if err != nil {
+				return fmt.Errorf("Error appending operation event to execution history: %w", err)
```
caiordjesus, line 109: In this case I think we should not return, because returning here means closing the worker. In these scenarios, I think we should log ERROR and continue.

internal/core/workers/operation_execution_worker/operation_execution_worker.go
```diff
@@ -89,14 +89,25 @@ func (w *OperationExecutionWorker) Start(ctx context.Context) error {
...
 		}
 
 		executor, hasExecutor := w.executorsByName[def.Name()]
 		if !hasExecutor {
 			loopLogger.Warn("operation definition has no executor")
+
 			w.evictOperation(ctx, loopLogger, op)
 			reportOperationEvicted(w.schedulerName, op.DefinitionName, LabelNoOperationExecutorFound)
+
+			err = w.operationManager.AppendOperationEventToExecutionHistory(ctx, op, "Operation definition has no executor")
```
caiordjesus, line 107: Instead of writing "operation has no executor", put this line to the `w.evictOperation` method, with something like "operation evicted because..."

internal/core/workers/operation_execution_worker/operation_execution_worker.go
```diff
@@ -108,6 +119,11 @@ func (w *OperationExecutionWorker) Start(ctx context.Context) error {
 
 		loopLogger.Info("Starting operation")
 
+		err = w.operationManager.AppendOperationEventToExecutionHistory(ctx, op, "Starting operation")
+		if err != nil {
+			return fmt.Errorf("Error appending operation event to execution history: %w", err)
```
caiordjesus, line 124: Continue instead of returning

internal/core/workers/operation_execution_worker/operation_execution_worker.go
```diff
@@ -121,6 +137,11 @@ func (w *OperationExecutionWorker) Start(ctx context.Context) error {
 				loopLogger.Error("failed to finish operation", zap.Error(err))
 			}
 
+			err = w.operationManager.AppendOperationEventToExecutionHistory(ctx, op, "Failed to grant lease to operation")
+			if err != nil {
+				return fmt.Errorf("Error appending operation event to execution history: %w", err)
```
caiordjesus, line 142: Continue instead of returning

internal/core/workers/operation_execution_worker/operation_execution_worker.go
```diff
@@ -136,6 +157,12 @@ func (w *OperationExecutionWorker) Start(ctx context.Context) error {
 			}
 
 			reportOperationExecutionWorkerFailed(w.schedulerName, LabelStartOperationFailed)
+
+			err = w.operationManager.AppendOperationEventToExecutionHistory(ctx, op, "Failed to finish operation")
+			if err != nil {
+				return fmt.Errorf("Error appending operation event to execution history: %w", err)
```
caiordjesus, line 163: Continue instead of returning

internal/core/workers/operation_execution_worker/operation_execution_worker.go
```diff
@@ -149,6 +176,11 @@ func (w *OperationExecutionWorker) Start(ctx context.Context) error {
 			op.Status = operation.StatusError
 			if errors.Is(executionErr, context.Canceled) {
 				op.Status = operation.StatusCanceled
+
+				err = w.operationManager.AppendOperationEventToExecutionHistory(ctx, op, "Operation cancelled")
+				if err != nil {
+					return fmt.Errorf("Error appending operation event to execution history: %w", err)
```
caiordjesus, line 182: Continue instead of returning

internal/core/workers/operation_execution_worker/operation_execution_worker.go
```diff
@@ -160,6 +192,11 @@ func (w *OperationExecutionWorker) Start(ctx context.Context) error {
 			if onErrorErr != nil {
 				loopLogger.Error("operation OnError failed", zap.Error(onErrorErr))
 			}
+
+			err = w.operationManager.AppendOperationEventToExecutionHistory(ctx, op, "Operation execution failed")
+			if err != nil {
+				return fmt.Errorf("Error appending operation event to execution history: %w", err)
```
caiordjesus, line 198: Continue instead of returning

internal/core/workers/operation_execution_worker/operation_execution_worker.go
```diff
@@ -174,6 +211,11 @@ func (w *OperationExecutionWorker) Start(ctx context.Context) error {
 		if err != nil {
 			loopLogger.Error("failed to revoke operation lease", zap.Error(err))
 		}
+
+		err = w.operationManager.AppendOperationEventToExecutionHistory(ctx, op, "Starting operation")
+		if err != nil {
+			return fmt.Errorf("Error appending operation event to execution history: %w", err)
```
caiordjesus, line 217: Continue instead of returning

internal/core/services/operation_manager/operation_manager.go
```diff
@@ -298,6 +304,16 @@ func (om *OperationManager) StartLeaseRenewGoRoutine(operationCtx context.Contex
...
 }
 
+func (om *OperationManager) AppendOperationEventToExecutionHistory(ctx context.Context, op *operation.Operation, eventMessage string) error {
+	event := operation.OperationEvent{
+		CreatedAt: time.Now().UTC(),
+		Event:     eventMessage,
+	}
+	op.ExecutionHistory = append(op.ExecutionHistory, event)
+
+	return om.Storage.UpdateOperationExecutionHistory(ctx, op)
```
caiordjesus, line 314: I think this method is not supposed to break. What's your opinion about it?
Maybe trying to update storage fails, and we log ERROR, but the append method itself does not returns an error. I don't see a reason why appending should break something. I just imagine an error being returned for logging reasons, which the method itself could be responsible for doing

caiordjesus (changes requested):
Failing on appending to execution history should not break the worker's execution, should only log

caiordjesus (approved):

internal/core/services/operation_manager/operation_manager.go
```diff
@@ -298,6 +304,21 @@ func (om *OperationManager) StartLeaseRenewGoRoutine(operationCtx context.Contex
 	}()
 }
 
+func (om *OperationManager) AppendOperationEventToExecutionHistory(ctx context.Context, op *operation.Operation, eventMessage string) {
+	managerLogger := om.Logger.With(zap.String(logs.LogFieldOperationID, op.ID), zap.String(logs.LogFieldSchedulerName, op.SchedulerName))
+	managerLogger.Info("starting operation lease renew go routine")
```
guilhermocc, line 309: @arthur29 This is log is wrong

internal/core/services/operation_manager/operation_manager_test.go
```diff
@@ -1007,6 +1007,44 @@ func TestEnqueueOperationCancellationRequest(t *testing.T) {
 	})
 }
 
+func TestAppendOperationEventToExecutionHistory(t *testing.T) {
+	t.Run("success case", func(t *testing.T) {
```
guilhermocc, line 1011: @arthur29 We are missing failure test scenario here

internal/core/workers/operation_execution_worker/operation_execution_worker_test.go
```diff
@@ -72,6 +72,7 @@ func TestSchedulerOperationsExecutionLoop(t *testing.T) {
 		workerService := NewOperationExecutionWorker(scheduler, workers.ProvideWorkerOptions(operationManager, executors, nil, nil))
 
 		operationManager.EXPECT().NextSchedulerOperation(gomock.Any(), expectedOperation.SchedulerName).Return(expectedOperation, operationDefinition, nil)
+		operationManager.EXPECT().AppendOperationEventToExecutionHistory(gomock.Any(), expectedOperation, gomock.Any()).AnyTimes()
```
guilhermocc, line 75: @arthur29 WDYT about asserting the event message and the number of calls in these tests? we are using too permissive arguments here in a sense we are not testing if we are updating the operation history correctly

guilhermocc (approved):

PR #352, opened 2022-03-18
adds timestamp info to logs

caiordjesus:
### What?
Format timestamp as readable to production logs

### Why?
Right now, the logs produced by Maestro don't have readable timestamps within (they're unix epoch). This is a problem because we end up depending on external tools to know when the logs were produced, and this information is 100% necessary for troubleshooting.

### Screen capure
image

### DIFF
--- a/internal/service/logging.go
+++ b/internal/service/logging.go
@@ -24,6 +24,7 @@ package service
 
 import (
 	"fmt"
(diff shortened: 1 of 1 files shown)


guilhermocc (approved):

arthur29 (approved):
LGTM

PR #353, opened 2022-03-18
removes version from remove rooms operation definition

caiordjesus:
### What?
Removes `version` from remove rooms definition.
### Why?
The `version` property on the remove rooms operation definition is wrong. Should not be there, and makes no sense to the operation.


arthur29 (approved):
LGTM

PR #356, opened 2022-03-21
Add metrics to scheduler storage

arthur29:
### Why
Scheduler storage doesn't have any metric related to database connections.

### What
This PR proposes to add metrics to scheduler storage with latency in database connections and errors count.

### DIFF
--- a/internal/adapters/metrics/pg.go
+++ b/internal/adapters/metrics/pg.go
@@ -20,21 +20,33 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 // SOFTWARE.
 
-package scheduler
+package metrics
 
 import (
-	"time"
-
-	"github.com/topfreegames/maestro/internal/adapters/metrics"
 	"github.com/topfreegames/maestro/internal/core/monitoring"
 )
 
-func reportSchedulerStorageFailsCounterMetric(storage, operation string, labels ...string) {
-	metrics.PostgresFailsCounterMetric.WithLabelValues(append([]string{storage, operation}, labels...)...).Inc()
-}
+var (
+	PostgresFailsCounterMetric = monitoring.CreateCounterMetric(&monitoring.MetricOpts{
+		Namespace: monitoring.Namespace,
+		Subsystem: monitoring.SubsystemWorker,
+		Name:      "postgres_fails_counter_metric",
+		Help:      "Postgres fails counter metric",
+		Labels: []string{
+			monitoring.LabelStorage,
+			monitoring.LabelOperation,
+			monitoring.LabelScheduler,
+		},
+	})
 
-func reportSchedulerStorageLatency(start time.Time, storage, operation string) {
-	monitoring.ReportLatencyMetricInMillis(
-		metrics.PostgresLatencyMetric, start, storage, operation,
-	)
-}
+	PostgresLatencyMetric = monitoring.CreateLatencyMetric(&monitoring.MetricOpts{
+		Namespace: monitoring.Namespace,
+		Subsystem: monitoring.SubsystemWorker,
+		Name:      "postgres_latency_metric",
+		Help:      "Postgres latency metric",
+		Labels: []string{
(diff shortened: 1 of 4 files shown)


internal/adapters/scheduler/scheduler_storage_pg.go
```diff
@@ -100,11 +101,14 @@ func (s schedulerStorage) GetScheduler(ctx context.Context, name string) (*entit
 	var dbScheduler Scheduler
 
 	queryString := queryGetActiveScheduler
+	start := time.Now()
```
guilhermocc, line 104: @arthur29 What's your opinion on having a wrapper function to be used in all of these methods for reporting latency metrics instead of calling it directly, something like:
```
func queryReportingLatency(storageOperation string, queryFunction func() error) error {
	start := time.Now()
	functionResult := queryFunction()
	reportSchedulerStorageLatency(start, storageOperation)
	return functionResult
}
```

And them when executing some query:
```
	err := queryReportingLatency("GetScheduler", func() error {
		_, err := client.QueryOne(&dbScheduler, queryString, name)
		return err
	})
```

internal/adapters/metrics/pg.go
```diff
@@ -0,0 +1,52 @@
...
+)
+
+var (
+	PostgresFailsCounterMetric = monitoring.CreateCounterMetric(&monitoring.MetricOpts{
+		Namespace: monitoring.Namespace,
+		Subsystem: monitoring.SubsystemWorker,
+		Name:      "postgres_fails_counter_metric",
+		Help:      "Postgres fails counter metric",
+		Labels: []string{
+			monitoring.LabelStorage,
```
guilhermocc, line 36: @arthur29 What about having a `Reason` label here so we can index the metrics by error message?

internal/adapters/scheduler/scheduler_storage_pg.go
```diff
@@ -114,45 +118,51 @@ func (s schedulerStorage) GetScheduler(ctx context.Context, name string) (*entit
...
 	_, err := client.QueryOne(&dbScheduler, queryString, queryArr...)
+	reportSchedulerStorageLatency(start, "GetSchedulerWithFilter")
 	if err == pg.ErrNoRows {
-		return nil, errors.NewErrNotFound("scheduler %s not found", SchedulerFilter.Name)
+		return nil, errors.NewErrNotFound("scheduler %s not found", schedulerFilter.Name)
 	}
 	if err != nil {
-		return nil, errors.NewErrUnexpected("error getting scheduler %s", SchedulerFilter.Name).WithError(err)
+		reportSchedulerStorageFailsCounterMetric(schedulerFilter.Name)
```
guilhermocc, line 145: Shouldn't we use the operation name here? (GetSchedulerWithFilter)

internal/adapters/scheduler/scheduler_storage_pg.go
guilhermocc: Also, we could even handle fail metrics in this wrapper: 
```
func queryReportingLatency(storageOperation string, queryFunction func() error) error {
	start := time.Now()
	err := queryFunction()
         if err != nil {
            reportSchedulerStorageFailsCounterMetric(storageOperation)
         }

	reportSchedulerStorageLatency(start, storageOperation)
	return err
}
```

guilhermocc (changes requested):
Pass storage operation as argument at line 145 in scheduler_storage_pg

internal/adapters/metrics/pg.go
guilhermocc: This could led to a scenario where we would have a high number of labels

internal/adapters/scheduler/scheduler_storage_pg.go
arthur29: Yeah

guilhermocc (approved):

internal/adapters/scheduler/scheduler_storage_pg.go
guilhermocc: Awesome!

PR #358, opened 2022-03-21
refactor - switch active version definition

caiordjesus:
### What?
Make `switch_active_version` operation definition have only `"newVersion"`, instead of `"newScheduler"` and `"replacePods"`.

### Why?
Since now we have the "input" property in the Operation entity, and "input" is the representation of what the client gave to Maestro, we want the definition to be exactly this.
That said, the definition of "switch active version" should have only the version being switched to, and not the whole entity (since the user is not giving it to Maestro, is something that Maestro fetches from the database) and "replacePods" (which is a business rule of the operation itself)

### DIFF
--- a/internal/api/handlers/schedulers_handler_test.go
+++ b/internal/api/handlers/schedulers_handler_test.go
@@ -1084,7 +1084,6 @@ func TestSwitchActiveVersion(t *testing.T) {
 		schedulerManager := scheduler_manager.NewSchedulerManager(schedulerStorage, schedulerCache, operationManager, roomStorage)
 
 		schedulerStorage.EXPECT().GetSchedulerWithFilter(gomock.Any(), gomock.Any()).Return(newValidScheduler(), nil)
-		schedulerStorage.EXPECT().GetScheduler(gomock.Any(), gomock.Any()).Return(newValidScheduler(), nil)
 		operationManager.EXPECT().CreateOperation(gomock.Any(), "scheduler-name-1", gomock.Any()).Return(&operation.Operation{ID: "id-1"}, nil)
 
 		mux := runtime.NewServeMux()
@@ -1125,10 +1124,10 @@ func TestSwitchActiveVersion(t *testing.T) {
 		mux.ServeHTTP(rr, req)
 
 		require.Equal(t, 404, rr.Code)
-		require.Contains(t, rr.Body.String(), "no scheduler versions found to switch")
+		require.Contains(t, rr.Body.String(), "not found")
 	})
 
-	t.Run("fails when operation enqueue fails", func(t *testing.T) {
+	t.Run("fails when operation enqueue fails since version does not exist", func(t *testing.T) {
 		mockCtrl := gomock.NewController(t)
 		defer mockCtrl.Finish()
 		schedulerStorage := mockports.NewMockSchedulerStorage(mockCtrl)
@@ -1138,7 +1137,6 @@ func TestSwitchActiveVersion(t *testing.T) {
 		schedulerManager := scheduler_manager.NewSchedulerManager(schedulerStorage, schedulerCache, operationManager, roomStorage)
 
 		schedulerStorage.EXPECT().GetSchedulerWithFilter(gomock.Any(), gomock.Any()).Return(newValidScheduler(), nil)
-		schedulerStorage.EXPECT().GetScheduler(gomock.Any(), gomock.Any()).Return(newValidScheduler(), nil)
 		operationManager.EXPECT().CreateOperation(gomock.Any(), "scheduler-name-1", gomock.Any()).Return(nil, errors.NewErrUnexpected("internal error"))
 
 		mux := runtime.NewServeMux()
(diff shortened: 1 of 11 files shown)


internal/api/handlers/schedulers_handler_test.go
```diff
@@ -1152,7 +1150,7 @@ func TestSwitchActiveVersion(t *testing.T) {
 		mux.ServeHTTP(rr, req)
 
 		require.Equal(t, 500, rr.Code)
-		require.Contains(t, rr.Body.String(), "failed to schedule operation: failed to schedule switch_active_version operation")
+		require.Contains(t, rr.Body.String(), "internal error")
```
guilhermocc, line 1153: Shouldn't we return status code 404 here?

internal/core/operations/switch_active_version/switch_active_version_executor.go
```diff
@@ -79,8 +79,18 @@ func (ex *SwitchActiveVersionExecutor) Execute(ctx context.Context, op *operatio
...
+
+	scheduler, err := ex.schedulerManager.GetSchedulerByVersion(ctx, op.SchedulerName, updateDefinition.NewActiveVersion)
+	if err != nil {
+		logger.Error("error fetching scheduler to be switched to", zap.Error(err))
+		return err
+	}
+
+	replacePods, err := ex.shouldReplacePods(ctx, scheduler)
+	if err != nil {
+		logger.Error("error calculating if should replace pods", zap.Error(err))
```
guilhermocc, line 91:
```suggestion
		logger.Error("error deciding if should replace pods", zap.Error(err))
```

internal/core/operations/switch_active_version/switch_active_version_executor.go
```diff
@@ -79,8 +79,18 @@ func (ex *SwitchActiveVersionExecutor) Execute(ctx context.Context, op *operatio
 	if !ok {
 		return fmt.Errorf("the definition is invalid. Should be type SwitchActiveVersionDefinition")
 	}
-	scheduler := &updateDefinition.NewActiveScheduler
-	replacePods := updateDefinition.ReplacePods
+
+	scheduler, err := ex.schedulerManager.GetSchedulerByVersion(ctx, op.SchedulerName, updateDefinition.NewActiveVersion)
```
guilhermocc, line 83: We need to cover both of these new error scenarios with tests

internal/api/handlers/schedulers_handler_test.go
guilhermocc: BTW, this change should not to be implemented in this PR

internal/core/services/scheduler_manager/scheduler_manager.go
```diff
@@ -67,6 +72,18 @@ func (s *SchedulerManager) GetActiveScheduler(ctx context.Context, schedulerName
 	return activeScheduler, nil
 }
 
+func (s *SchedulerManager) GetSchedulerByVersion(ctx context.Context, schedulerName, schedulerVersion string) (*entities.Scheduler, error) {
```
guilhermocc, line 75: Missing test for this new guy

guilhermocc (changes requested):
Need to implement missing tests

internal/core/services/scheduler_manager/scheduler_manager.go
```diff
@@ -49,6 +52,8 @@ type SchedulerManager struct {
 	logger           *zap.Logger
 }
 
+var _ ports.SchedulerManager = (*SchedulerManager)(nil)
```
arthur29, line 55: Top irmão

internal/api/handlers/schedulers_handler_test.go
caiordjesus: No, there's another test that returns 404 when the scheduler version is not found. This one simply reproduces an internal server error. See line 1126

guilhermocc (approved):

internal/core/operations/switch_active_version/switch_active_version_executor.go
```diff
@@ -227,3 +237,11 @@ func (ex *SwitchActiveVersionExecutor) appendToNewCreatedRooms(schedulerName str
 func (ex *SwitchActiveVersionExecutor) clearNewCreatedRooms(schedulerName string) {
 	delete(ex.newCreatedRooms, schedulerName)
 }
+
+func (ex *SwitchActiveVersionExecutor) shouldReplacePods(ctx context.Context, newScheduler *entities.Scheduler) (bool, error) {
+	actualActiveScheduler, err := ex.schedulerManager.GetActiveScheduler(ctx, newScheduler.Name)
+	if err != nil {
+		return false, err
+	}
+	return actualActiveScheduler.IsMajorVersion(newScheduler), nil
```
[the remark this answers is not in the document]
caiordjesus: It's a great idea! I just think it should be contained on another PR in the future... We'll come back to see this asap

PR #359, opened 2022-03-21
Fix success metric label value 

guilhermocc:
## What ❓ 
Fix `success` label value on operation execution metric reporting

## Why 🤔 
The `success` label value was wrong, it should be success true when no error occurs (err == nil) and false otherwise.

### DIFF
--- a/internal/core/workers/operation_execution_worker/operation_execution_worker.go
+++ b/internal/core/workers/operation_execution_worker/operation_execution_worker.go
@@ -150,9 +150,9 @@ func (w *OperationExecutionWorker) Start(ctx context.Context) error {
 		}
 		w.operationManager.StartLeaseRenewGoRoutine(operationContext, op)
 
-		executeStartTime := time.Now()
-		executionErr := executor.Execute(operationContext, op, def)
-		reportOperationExecutionLatency(executeStartTime, w.schedulerName, op.DefinitionName, executionErr == nil)
+		executionErr := executeCollectingLatencyMetrics(w.schedulerName, op.DefinitionName, func() error {
+			return executor.Execute(operationContext, op, def)
+		})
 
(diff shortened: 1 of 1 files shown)


caiordjesus (approved):

internal/core/workers/operation_execution_worker/operation_execution_worker.go
```diff
@@ -152,7 +152,7 @@ func (w *OperationExecutionWorker) Start(ctx context.Context) error {
 
 		executeStartTime := time.Now()
 		executionErr := executor.Execute(operationContext, op, def)
-		reportOperationExecutionLatency(executeStartTime, w.schedulerName, op.DefinitionName, executionErr != nil)
+		reportOperationExecutionLatency(executeStartTime, w.schedulerName, op.DefinitionName, executionErr == nil)
```
arthur29, line 155: Hey @guilhermocc what do you think to use the same approach from schedulers' adapters? On that, it would receive a function in the `reportOperationExecutionLatency` instead of the start time.

arthur29, line 155: I think that is not a good approach in Golang to test if something is equal to nil ([link](https://yourbasic.org/golang/gotcha-why-nil-error-not-equal-nil/)). What do you think to remove this label `isSuccess` in favor of an exclusive metric `errorCount`?

guilhermocc: Actually, the link you shared warns about the possibility of returning a nil value of a specific error type that can cause problems on checking both `err == nil` and `err != nil`... I think this PR is not the ideal context to make this change and create an abstraction of operation execution error.

arthur29 (approved):
LGTM

PR #361, opened 2022-03-22
fix cancel operation logs

caiordjesus:
### What?
Fix cancel operation logs
### Why?
Logs were duplicated and not well put


arthur29 (approved):
LGTM

PR #362, opened 2022-03-23
Include events forwarding latency metrics

guilhermocc:
## What ❓ 
Create and produce metrics for events forwarding grpc calls. Rename postgres metrics.

## Why 🤔 
To improve events forwarding observability


internal/adapters/metrics/pg.go
```diff
@@ -30,7 +30,7 @@ var (
 	PostgresFailsCounterMetric = monitoring.CreateCounterMetric(&monitoring.MetricOpts{
 		Namespace: monitoring.Namespace,
 		Subsystem: monitoring.SubsystemWorker,
-		Name:      "postgres_fails_counter_metric",
```
guilhermocc, line 33: Just a refactor here

PR #366, opened 2022-03-24
Improvement/add game label to metrics

guilhermocc:
## What ❓ 
Add `game` label to operations, event forwarding, rooms, and instances metrics 

## Why 🤔 
The `game` information is a good value to have as a label so we can index those metrics by the game once each maestro deployment can have an arbitrary number of games/schedulers

### DIFF
--- a/internal/core/services/events_forwarder/metrics.go
+++ b/internal/core/services/events_forwarder/metrics.go
@@ -33,6 +33,7 @@ var (
 		Name:      "success_room_event_forwarding",
 		Help:      "Current number of room events forwarding with success",
 		Labels: []string{
+			monitoring.LabelGame,
 			monitoring.LabelScheduler,
 		},
 	})
@@ -45,6 +46,7 @@ var (
 		Name:      "success_player_event_forwarding",
 		Help:      "Current number of player events forwarding with success",
 		Labels: []string{
+			monitoring.LabelGame,
 			monitoring.LabelScheduler,
 		},
 	})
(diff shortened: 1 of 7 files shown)


internal/core/services/events_forwarder/metrics.go
```diff
@@ -74,18 +74,18 @@ var (
 	})
 )
 
-func reportRoomEventForwardingSuccess(schedulerName string) {
+func reportRoomEventForwardingSuccess(game, schedulerName string) {
 	successRoomEventForwardingMetric.WithLabelValues(schedulerName).Inc()
```
caiordjesus, line 78: Did you forgot to put game as a label? xD
image

internal/core/workers/metricsreporter/metrics_reporter_worker.go
```diff
@@ -54,6 +55,7 @@ type MetricsReporterWorker struct {
 func NewMetricsReporterWorker(scheduler *entities.Scheduler, opts *workers.WorkerOptions) workers.Worker {
 	return &MetricsReporterWorker{
 		schedulerName:   scheduler.Name,
+		game:            scheduler.Game,
```
caiordjesus, line 58: Is there a specific reason to not give the entire scheduler here? Feels like for every new scheduler property we will change this struct and give scheduler.Something 🤔

internal/core/services/events_forwarder/metrics.go
guilhermocc: YES!

internal/core/workers/metricsreporter/metrics_reporter_worker.go
guilhermocc: Maybe we were not including because we didn't need any other scheduler field... maybe it makes more sense to have the entire scheduler now

caiordjesus (approved):

PR #369, opened 2022-03-28
Feature/accept value from env var

guilhermocc:
## What ❓ 
Update ContainerEnvVar struct to define field ref and secret key ref env vars. Update unitary, integration and e2e tests.

## Why 🤔 
Until now we weren't supporting field ref and secret key ref env vars on schedulers, and this is a product requirement to provide more flexibility for building schedulers with more flexibility.


e2e/suites/management/utils.go
```diff
@@ -121,6 +136,16 @@ func createSchedulerAndWaitForIt(
 
 		return len(svcAccs.Items) > 0
 	}, 5*time.Second, time.Second)
+
+	// creating secret used by the pods
+	kubeClient.CoreV1().Secrets(schedulerName).Create(context.Background(), &v1.Secret{
```
guilhermocc, line 141: If we don't create the secret in the namespace, the add rooms operations will fail because there will be errors while creating the instance on kubernetes

guilhermocc: > In this one which tool we will offer to the game team to create this kind of secret without touching in the scheduler namespace?

Currently, in this first release, the maestro does not provide means for creating those secrets. They must be created manually. We are adding it here for supporting maestro v9 schedulers, but we will need to better explore this feature (like supporting other tools like vault).

arthur29 (approved):

e2e/suites/management/utils.go
```diff
@@ -204,6 +249,11 @@ func createSchedulerWithForwardersAndWaitForIt(
 
 		return len(svcAccs.Items) > 0
 	}, 5*time.Second, time.Second)
+
+	// creating secret used by the pods
+	// creating secret used by the pods
```
caiordjesus, line 254: Is this duplicated? Or it a bug from github? xD

caiordjesus (approved):

PR #372, opened 2022-03-30
Improvement/normalize operation execution errors

guilhermocc:
## What ❓ 
Create custom errors for operation execution and services. Use a formatted error message when populating operation execution history with possible errors.

## Why 🤔 
Currently, we are not formatting error messages that are appended in the operation execution history, this behavior makes the user experience worse for understanding and dealing with the problem.

### DIFF
--- a/internal/core/operations/remove_rooms/remove_rooms_executor.go
+++ b/internal/core/operations/remove_rooms/remove_rooms_executor.go
@@ -72,9 +72,8 @@ func (e *RemoveRoomsExecutor) Execute(ctx context.Context, op *operation.Operati
 
 			if errors.Is(err, serviceerrors.ErrGameRoomStatusWaitingTimeout) {
 				return operations.NewErrTerminatingPingTimeout(deleteErr)
-			} else {
-				return operations.NewErrUnexpected(deleteErr)
 			}
+			return operations.NewErrUnexpected(deleteErr)
 		}
 	}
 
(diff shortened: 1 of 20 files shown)


internal/core/operations/remove_rooms/remove_rooms_executor.go
```diff
@@ -63,15 +68,21 @@ func (e *RemoveRoomsExecutor) Execute(ctx context.Context, op *operation.Operati
 		if err != nil {
 			reportDeletionFailedTotal(op.SchedulerName, op.ID)
 			logger.Warn("failed to remove rooms", zap.Error(err))
-			return fmt.Errorf("failed to remove room: %w", err)
+			deleteErr := fmt.Errorf("failed to remove room: %w", err)
+
+			if errors.Is(err, serviceerrors.ErrGameRoomStatusWaitingTimeout) {
+				return operations.NewErrTerminatingPingTimeout(deleteErr)
+			} else {
```
caiordjesus, line 75: I think this else is unnecessary, no? Like I imagine we could write more else ifs in case we map more kinds of error, but they would always fallback to "return unexpected"

caiordjesus (approved):
Very clean! Approved! Just so we don't lose habit: did you remembered to run the e2e? xD

internal/core/operations/remove_rooms/remove_rooms_executor.go
guilhermocc: [7d6d6d9](https://github.com/topfreegames/maestro/pull/372/commits/7d6d6d9e568f1e7317e7cba0007b773a5e989935)

PR #379, opened 2022-04-01
Fix create room method rollback

guilhermocc:
## What ❓ 
Use a timed context while delivering create room in an error scenario on CreateRoomAndWaitForReadiness method.

## Why 🤔 
We were using the same context for deleting created game rooms in an error scenario on CreateRoomAndWaitForReadiness method. This was causing some rooms to remain created when canceling add_rooms or switch_version operations.


caiordjesus (approved):

guilhermocc:
Fix create room method rollback

PR #380, opened 2022-04-01
Fix/port allocation logic

guilhermocc:
## What ❓ 
Fix port allocation logic by updating port_allocator adapter and creating a function to make a deep copy of the spec struct.

## Why 🤔 
Currently, our port allocation logic is generating port conflicts while creating pods for two reasons:

- We are using a non-safe seed for generating random numbers in RandomPortAllocator. According to the random lib we are using, we should use the lib exported functions for concurrency safety (e.g rand.Int31n(currentRange.Total()), and not the methods from the struct we are building.
- In room_manager **CreateRoomAndWaitForReadiness** method, we are assigning the generated HostPort to the same address (Spec container slice element), which is causing the ports to conflict when we add more than 1 room.


internal/core/entities/game_room/spec.go
```diff
@@ -48,3 +48,15 @@ func NewSpec(version string, terminationGracePeriod time.Duration, containers []
 		Toleration:             toleration,
 		Affinity:               affinity}
 }
+
+func (s *Spec) DeepCopy() *Spec {
```
caiordjesus, line 52: Kinda implementing prototype here, right? 🎉

caiordjesus (approved):

PR #382, opened 2022-04-01
Remove error log when cancelation was called

arthur29:
### Why
When one operation was canceled it add an error to the operation execution history.

### What
This PR proposes:
- Change the cancel message
- Remove the error when an operation was canceled
- Add an rollback start message

### DIFF
--- a/internal/core/workers/operation_execution_worker/operation_execution_worker.go
+++ b/internal/core/workers/operation_execution_worker/operation_execution_worker.go
@@ -156,15 +156,17 @@ func (w *OperationExecutionWorker) Start(ctx context.Context) error {
 
 		op.Status = operation.StatusFinished
 		if executionErr != nil {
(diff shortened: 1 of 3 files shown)


internal/core/workers/operation_execution_worker/operation_execution_worker.go
```diff
@@ -156,16 +156,19 @@ func (w *OperationExecutionWorker) Start(ctx context.Context) error {
 
 		op.Status = operation.StatusFinished
 		if executionErr != nil {
-			op.Status = operation.StatusError
-			if errors.Is(executionErr.Error(), context.Canceled) {
+			if executionErr.IsContextCanceled() {
```
guilhermocc, line 159: Sweet!

guilhermocc (approved):

PR #383, opened 2022-04-01
Make e2e tests run in parallel

guilhermocc:
## What ❓ 
Add **t.Parallel()** call in e2e tests functions 

## Why 🤔 
With the fix in the port allocation logic, we can now assure that the e2e tests intermittence will drop drastically, enabling us to run e2e tests in parallel


caiordjesus (approved):
SEXY

PR #384, opened 2022-04-04
Feature/get operation by

caiordjesus:
### What?
Adds GET Operation method.
### Why?
To make it easier to understand what is happening with an Operation, we need an API endpoint GET /operations/:operation_id


internal/api/handlers/operations_handler.go
```diff
@@ -130,6 +130,27 @@ func (h *OperationsHandler) CancelOperation(ctx context.Context, request *api.Ca
 	return &api.CancelOperationResponse{}, nil
 }
 
+func (h *OperationsHandler) GetOperation(ctx context.Context, request *api.GetOperationRequest) (*api.GetOperationResponse, error) {
+	handlerLogger := h.logger.With(zap.String(logs.LogFieldSchedulerName, request.GetSchedulerName()), zap.String(logs.LogFieldOperationID, request.GetOperationId()))
+	handlerLogger.Info("received request to get operation by id")
+	op, _, err := h.operationManager.GetOperation(ctx, request.GetSchedulerName(), request.GetOperationId())
+	if err != nil {
+		if errors.Is(err, portsErrors.ErrNotFound) {
+			handlerLogger.Warn("operation not found", zap.String(logs.LogFieldSchedulerName, request.GetSchedulerName()), zap.String(logs.LogFieldOperationID, request.GetOperationId()), zap.Error(err))
```
guilhermocc, line 139:
```suggestion
			handlerLogger.Warn("operation not found", zap.String(logs.LogFieldSchedulerName, request.GetSchedulerName()), zap.Error(err))
```
We already defined the op ID label at line 134

internal/api/handlers/operations_handler.go
```diff
@@ -130,6 +130,27 @@ func (h *OperationsHandler) CancelOperation(ctx context.Context, request *api.Ca
...
+			return nil, status.Error(codes.NotFound, err.Error())
+		}
+		handlerLogger.Error("error fetching operation", zap.String(logs.LogFieldSchedulerName, request.GetSchedulerName()), zap.String(logs.LogFieldOperationID, request.GetOperationId()), zap.Error(err))
+		return nil, status.Error(codes.Unknown, err.Error())
+	}
+
+	convertedOp, err := h.fromOperationToResponse(op)
+	if err != nil {
+		handlerLogger.Error("invalid operation object. Fail to convert", zap.String(logs.LogFieldSchedulerName, request.GetSchedulerName()), zap.String(logs.LogFieldOperationID, request.GetOperationId()), zap.Error(err))
```
guilhermocc, line 148:
```suggestion
		handlerLogger.Error("invalid operation object. Fail to convert", zap.String(logs.LogFieldSchedulerName, request.GetSchedulerName()), zap.Error(err))
```

internal/api/handlers/operations_handler_test.go
```diff
@@ -454,6 +454,124 @@ func TestCancelOperation(t *testing.T) {
...
 
+func TestGetOperation(t *testing.T) {
+	dates := []time.Time{
+		time.Time{}.AddDate(2020, 0, 0),
+		time.Time{}.AddDate(2020, 1, 0),
+		time.Time{}.AddDate(2020, 2, 0),
+	}
+
+	type tableTestInput struct {
+		Operation *operation.Operation
```
guilhermocc, line 465: This guy over here is mock preparation, right? it seems to me that this Operation field is not actually input for the test but rather some kind of mock preparation. WDYT?

internal/api/handlers/operations_handler_test.go
```diff
@@ -454,6 +454,124 @@ func TestCancelOperation(t *testing.T) {
...
+		Err       error
+	}
+	type tableTests struct {
+		Description string
+		Input       tableTestInput
+		Output      tableTestOutput
+	}
+
+	test1 := tableTests{
+		Description: "succeed - returns operation",
```
guilhermocc, line 479:
```suggestion
		Description: "returns operation when no error occur",
``` 
as we have discussed in the last weeks, we were trying to adopt some test name standards and we had agreed on something like `return <something> when <something happens>` WDYT?

guilhermocc (changes requested):
@caiordjesus we need to update management-api cmd metrics middleware to include the new route

guilhermocc (approved):

PR #391, opened 2022-04-06
Implement PatchSchedulerAndSwitchActiveVersionOperation service

arthur29:
### Why
Maestro needs a new route to patch a scheduler. And to it, a service that executes this function should be done.

### What
- Implement a `PatchSchedulerAndSwitchActiveVersionOperation`
- Create a `patch_scheduler` package

### DIFF
--- a/internal/core/services/scheduler_manager/scheduler_manager.go
+++ b/internal/core/services/scheduler_manager/scheduler_manager.go
@@ -39,6 +39,7 @@ import (
 	"github.com/topfreegames/maestro/internal/core/operations/create_scheduler"
 	"github.com/topfreegames/maestro/internal/core/operations/remove_rooms"
 	"github.com/topfreegames/maestro/internal/core/ports"
+	portsErrors "github.com/topfreegames/maestro/internal/core/ports/errors"
 	"go.uber.org/zap"
 )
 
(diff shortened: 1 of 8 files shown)


internal/core/services/scheduler_manager/scheduler_manager.go
```diff
@@ -140,6 +141,27 @@ func (s *SchedulerManager) CreateNewSchedulerVersionAndEnqueueSwitchVersion(ctx
 	return nil
 }
 
+func (s *SchedulerManager) PatchSchedulerAndSwitchActiveVersionOperation(ctx context.Context, schedulerName string, patchMap map[string]interface{}) (*operation.Operation, error) {
```
caiordjesus, line 144: This method name is kind of strange because it enqueues `CreateNewSchedulerVersion`, and not the `SwitchActiveVersion`... What do you think about renaming to `PatchSchedulerAndCreateNewSchedulerVersion`?

internal/core/services/scheduler_manager/scheduler_manager.go
```diff
@@ -140,6 +141,27 @@ func (s *SchedulerManager) CreateNewSchedulerVersionAndEnqueueSwitchVersion(ctx
...
+	scheduler, err := s.schedulerStorage.GetScheduler(ctx, schedulerName)
+	if err != nil {
+		return nil, fmt.Errorf("no scheduler found, can not create new version for inexistent scheduler: %w", err)
```
caiordjesus, line 147: Maybe here would be nice to return errors.NewErrNotFound() instead of a simple error, so in the handler it is easier to check the type of the error

internal/core/services/scheduler_manager/scheduler_manager.go
arthur29: Looks great!

caiordjesus (approved):

PR #397, opened 2022-04-11
Feature/create patch schedulers handler

arthur29:
### Why
Maestro needs a patch scheduler route that updates only certain fields.
### What
- Create `PATCH /schedulers/:scheduler_id` route


internal/api/handlers/adapters/schedulers.go
```diff
@@ -0,0 +1,169 @@
+package adapters
```
guilhermocc, line 1: @arthur29 What do you think about renaming this package to `request_adapters` or something like this? We already have an `adapters` package for out ports adapters structure.

guilhermocc (approved):

internal/api/handlers/adapters/schedulers.go
arthur29: Done

guilhermocc (approved):

PR #400, opened 2022-04-13
Improve migration script - Set min to 0, set replica to 0 before deleting

caiordjesus:
### Why?
Improve migration script - Set min to 0, set replica to 0 before deleting;
### What?
v9 is inconsistent, so delete usually times out. Setting replica to 0 usually times out. Because of that, the script is improved with retries and new steps


guilhermocc (approved):

arthur29 (approved):
LGTM

PR #402, opened 2022-04-14
Implement ForwarderClient SendRoomStatus method

guilhermocc:
## What ❓ 
Implement SendRoomStatus method in ForwarderClient

## Why 🤔 
We need to add a new method SendRoomStatus in the forwarder client so we can forward room status events successfully.


arthur29 (approved):
LGTM

PR #404, opened 2022-04-18
improves migration script compatibility

caiordjesus:
### What?
Improves migration script compatibility.
### Why?
Certain schedulers created on v9 have different formats that the actual script does not support.


arthur29 (approved):
LGTM

PR #408, opened 2022-04-20
include Getting Started documentation to run maestro-next locally

caiordjesus:
### What?
Include 'Getting Started' documentation .md

### Why?
To create a better user experience using Maestro-Next, we understand that running the application locally is very important to simulate its usage and help users understand how Maestro can be useful for them

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -140,3 +140,15 @@ deps/down: ## Delete containers dependencies.
 	@echo "Deleting dependencies "
 	@docker-compose --project-name maestro down
 	@echo "Dependencies deleted successfully."
+
+
+#-------------------------------------------------------------------------------
+#  Easily start Maestro
+#-------------------------------------------------------------------------------
+.PHONY: maestro/start
+maestro/start: build-linux-x86_64 ## Start Maestro with all of its dependencies
+	@echo "Starting maestro..."
+	@cd ./e2e/framework/maestro; docker-compose --project-name maestro up -d
+	@go run main.go migrate;
+	@cd ./e2e/framework/maestro; docker-compose --project-name maestro start
+	@echo "Maestro is up and running!"
--- a/Makefile
+++ b/Makefile
@@ -146,9 +146,9 @@ deps/down: ## Delete containers dependencies.
 #  Easily start Maestro
 #-------------------------------------------------------------------------------
 .PHONY: maestro/start
-maestro/start: build-linux-x86_64 ## Start Maestro with all of its dependencies
+maestro/start: build-linux-x86_64 ## Start Maestro with all of its dependencies.
 	@echo "Starting maestro..."
 	@cd ./e2e/framework/maestro; docker-compose --project-name maestro up -d
 	@go run main.go migrate;
-	@cd ./e2e/framework/maestro; docker-compose --project-name maestro start
+	@cd ./e2e/framework/maestro; docker-compose --project-name maestro start worker runtime-watcher #Worker and watcher do not work before migration, so we start them after it.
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -140,3 +140,15 @@ deps/down: ## Delete containers dependencies.
...
+
+#-------------------------------------------------------------------------------
+#  Easily start Maestro
+#-------------------------------------------------------------------------------
+.PHONY: maestro/start
+maestro/start: build-linux-x86_64 ## Start Maestro with all of its dependencies
+	@echo "Starting maestro..."
+	@cd ./e2e/framework/maestro; docker-compose --project-name maestro up -d
+	@go run main.go migrate;
+	@cd ./e2e/framework/maestro; docker-compose --project-name maestro start
```
arthur29, line 153: Why do you run docker-compose `up` and `start`

docs/GettingStarted.md
```diff
@@ -0,0 +1,95 @@
...
+This will build and start all containers needed by Maestro, such as databases and maestro-modules. 
+Because of that, be aware that it might take some time to finish.
+
+## Find rooms-api address
+To simulate a game room, it's important to find the address of running **rooms-api** on the local network.
+
+To do that, with Maestro containers running, simply use:
+
+```shell
+docker inspect -f '{{range.NetworkSettings.Networks}}{{.Gateway}}{{end}}' maestro_rooms-api_1
```
arthur29, line 29: I think that is good to point out here that the name of the container could change. It shouldn't  always be `maestro_rooms-api_1`

docs/GettingStarted.md
```diff
@@ -0,0 +1,95 @@
...
+```
+
+This command should give you an IP address. 
+This IP is important because the game rooms will use it to communicate their status.
+
+An example of the answer could be: **172.30.0.1**. This is the **ROOMS_API_ADDRESS**.
+
```
arthur29, line 36:
```suggestion
```

arthur29: I think you could remove this paragraph. It does not add any new information, than the basic one related to basic computer concepts.

docs/GettingStarted.md
```diff
@@ -0,0 +1,95 @@
...
+## Create a scheduler
+If everything is working as expected till now and each Maestro-module is up and running, use the command below to create a new scheduler:
```
arthur29, line 38:
```suggestion
If everything is working as expected now each Maestro module is up and running, use the command below to create a new scheduler:
```

docs/GettingStarted.md
```diff
@@ -0,0 +1,95 @@
...
+		],
+		"toleration": "",
+		"affinity": ""
+	},
+	"forwarders": []
+}'
+```
+
+## Congratulations
+If you followed the steps above (should not be hard) you have Maestro running in your local machine, and with a [scheduler](Scheduler.md) to try different [operations](Operations.md) on it.
```
arthur29, line 92:
```suggestion
If you followed the steps above you have Maestro running in your local machine, and with a [scheduler](Scheduler.md) to try different [operations](Operations.md) on it.
```

I don't like value judgment in documentation, because it is basic for us that is inside the project, but to someone more newbie this could be hard to complete. What do you think?

Makefile
caiordjesus: Because after you use `up`, the worker does not work since the migration hasn't been executed. Because of that, we use up, run migrate, then start everything.

caiordjesus: I improved the readability of it, can you look it again?

arthur29 (approved):

PR #409, opened 2022-04-26
Move scheduler_handler adapters to request_adapters package

caiordjesus:
### What?
Move scheduler_handler adapters to request_adapters package
### Why?
The scheduler_handler had a massive amount of methods that were not handler methods, but adapters to the relation API <> Maestro

### Format Refactors
Some methods on the adapter were simply static methods used by the public methods. Because of that, these methods were made private, and the tests were adapted to test the public methods. The unit tests made for those static methods were absorbed by the new tests for the public ones

### DIFF
--- a/internal/api/handlers/request_adapters/schedulers.go
+++ b/internal/api/handlers/request_adapters/schedulers.go
@@ -23,7 +23,11 @@
 package request_adapters
 
 import (
+	"fmt"
 	"time"
+
+	_struct "google.golang.org/protobuf/types/known/structpb"
+	"google.golang.org/protobuf/types/known/timestamppb"
 
 	"github.com/topfreegames/maestro/internal/core/entities"
 	"github.com/topfreegames/maestro/internal/core/entities/forwarder"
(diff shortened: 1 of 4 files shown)


internal/api/handlers/request_adapters/schedulers.go
```diff
@@ -179,17 +329,111 @@ func FromApiContainerEnvironments(apiEnvironments []*api.ContainerEnvironment) [
...
-	for _, apiPort := range apiPorts {
-		port := game_room.ContainerPort{
-			Name:     apiPort.GetName(),
-			Port:     int(apiPort.GetPort()),
-			Protocol: apiPort.GetProtocol(),
-			HostPort: int(apiPort.GetHostPort()),
+func fromApiForwarders(apiForwarders []*api.Forwarder) []*forwarder.Forwarder {
+	var forwarders []*forwarder.Forwarder
+	for _, apiForwarder := range apiForwarders {
+		_forwarder := forwarder.Forwarder{
```
arthur29, line 335: Hey @caiordjesus commonly variables with the first character as `_` are created to not be used. What do you think to rename it to something like `entityForwarder`?

arthur29 (approved):

PR #413, opened 2022-04-29
Use goroutines for processing runtime watcher events

guilhermocc:
Use goroutines for processing runtime watcher events, we have identified that runtime watcher is having some problems while processing a large number of runtime watcher events.


arthur29 (approved):
LGTM

caiordjesus (approved):

PR #414, opened 2022-05-02
Fix/runtime watcher duplicated events

caiordjesus:
### What?
Improves k8s usage caching instances to reduce load on kube API.
### Why?
Maestro behavior is bad when it need to manage too many rooms, since they produce too many events and all those events are fetching kube api.


arthur29 (approved):
LGTM

PR #416, opened 2022-05-03
Add rooms replicas in scheduler

arthur29:
### Why
To create the self-healing feature Scheduler should have the desired number of Game Rooms replicas to be used by the worker.

### What
This PR proposes to add a new field in the `Scheduler` entity the `RoomsReplicas` to be used by the new feature.


internal/core/entities/scheduler.go
```diff
@@ -47,19 +47,32 @@ const (
 	StateOnError = "on-error"
 )
 
+// Scheduler represent one of the basic maestro struct. It holds GameRooms definitions, as
+//   Spectations, PortRange and Forwarders to correct GameRooms work.
```
guilhermocc, line 51: `to correct GameRooms work` this part seems a bit odd to me. What about 
```
Scheduler represents one of the basic maestro structs. It holds GameRooms creation specifications, as well as optional events forwarders.
```

guilhermocc (approved):

PR #418, opened 2022-05-04
Refactor/operation worker execution flow

guilhermocc:
## What ❓ 
Refactor operation execution worker code, isolating the core execution logic into separate functions.

## Why 🤔 
Currently, our core execution logic resides inside a loop and is not isolated into a function, making it hard to expand operation execution worker logic in the future.

### DIFF
--- a/internal/core/workers/operation_execution_worker/operation_execution_worker.go
+++ b/internal/core/workers/operation_execution_worker/operation_execution_worker.go
@@ -83,14 +83,14 @@ func (w *OperationExecutionWorker) Start(ctx context.Context) error {
 			return fmt.Errorf("failed to get next operation: %w", err)
 		}
 
-		err = w.executeOperation(op, def)
+		err = w.executeOperationFlow(op, def)
 		if err != nil && shouldFinishWorker(err) {
 			return err
 		}
 	}
 }
 
-func (w *OperationExecutionWorker) executeOperation(op *operation.Operation, def operations.Definition) error {
+func (w *OperationExecutionWorker) executeOperationFlow(op *operation.Operation, def operations.Definition) error {
 	loopLogger := w.logger.With(
 		zap.String(logs.LogFieldOperationID, op.ID),
 		zap.String(logs.LogFieldOperationDefinition, def.Name()),
@@ -107,10 +107,12 @@ func (w *OperationExecutionWorker) executeOperation(op *operation.Operation, def
 		return nil
 	}
 
-	err, executionErr := w.startAndExecuteOperationWithLease(op, def, loopLogger, executor)
+	operationContext, err := w.prepareExecutionAndLease(op, loopLogger)
 	if err != nil {
 		return err
 	}
+
+	executionErr := w.executeOperationWithLease(operationContext, op, def, executor)
 
 	if executionErr != nil {
 		w.handleExecutionError(op, executionErr, loopLogger)
@@ -168,7 +170,13 @@ func (w *OperationExecutionWorker) finishOperationAndLease(op *operation.Operati
 	w.operationManager.AppendOperationEventToExecutionHistory(w.workerContext, op, "Operation finished")
 }
 
-func (w *OperationExecutionWorker) startAndExecuteOperationWithLease(op *operation.Operation, def operations.Definition, loopLogger *zap.Logger, executor operations.Executor) (err error, executionErr operations.ExecutionError) {
+func (w OperationExecutionWorker) executeOperationWithLease(operationContext context.Context, op *operation.Operation, def operations.Definition, executor operations.Executor) operations.ExecutionError {
+	return w.executeCollectingLatencyMetrics(op.DefinitionName, func() operations.ExecutionError {
+		return executor.Execute(operationContext, op, def)
+	})
+}
+
+func (w *OperationExecutionWorker) prepareExecutionAndLease(op *operation.Operation, loopLogger *zap.Logger) (operationContext context.Context, err error) {
 	loopLogger.Info("Starting operation")
 
 	w.operationManager.AppendOperationEventToExecutionHistory(w.workerContext, op, "Starting operation")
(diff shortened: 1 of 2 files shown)


internal/core/workers/operation_execution_worker/operation_execution_worker.go
```diff
@@ -76,125 +79,158 @@ func (w *OperationExecutionWorker) Start(ctx context.Context) error {
...
-			loopLogger.Warn("operation definition has no executor")
-
-			w.evictOperation(ctx, loopLogger, op)
-			reportOperationEvicted(w.scheduler.Game, w.scheduler.Name, op.DefinitionName, LabelNoOperationExecutorFound)
-
-			continue
+		err = w.executeOperation(op, def)
+		if err != nil {
+			switch {
+			case errors.Is(err, workererrors.ErrStartOperationFailed) ||
```
guilhermocc, line 89: Now it is easier to decide if we should stop the worker in those scenarios. Initially, I made just a refactor, didn't change the logic at all... Now it is a chance to change this behavior, @arthur29 @caiordjesus what do you guys think? should we continue to stop the worker and let the worker manage to recreate it in those scenarios? Do you guys think it is ok to change this behavior is this PR?

internal/core/workers/operation_execution_worker/operation_execution_worker.go
```diff
@@ -76,125 +79,158 @@ func (w *OperationExecutionWorker) Start(ctx context.Context) error {
...
+				errors.Is(err, workererrors.ErrGrantLeaseFailed) ||
+				errors.Is(err, workererrors.ErrOperationWithInvalidStatus):
+				return err
+			}
```
arthur29, line 93: What do you think to refactor this code to be more simple like:
```
		if err != nil && shouldFinishWorker(err)  {
			return err
		}
	}
}

func shouldFinishWorker(err error) bool {
	return errors.Is(err, workererrors.ErrStartOperationFailed) ||
	errors.Is(err, workererrors.ErrGrantLeaseFailed) ||
	errors.Is(err, workererrors.ErrOperationWithInvalidStatus)
}
```

internal/core/workers/operation_execution_worker/operation_execution_worker.go
```diff
@@ -76,125 +79,158 @@ func (w *OperationExecutionWorker) Start(ctx context.Context) error {
...
+		zap.String(logs.LogFieldOperationID, op.ID),
+		zap.String(logs.LogFieldOperationDefinition, def.Name()),
+	)
 
-		loopLogger.Info("Starting operation")
+	if op.Status != operation.StatusPending {
+		loopLogger.Warn("operation is at an invalid status to proceed")
 
-		w.operationManager.AppendOperationEventToExecutionHistory(ctx, op, "Starting operation")
+		return workererrors.NewErrOperationWithInvalidStatus("operation is at and invalid status to proceed")
```
arthur29, line 107:
```suggestion
		return workererrors.NewErrOperationWithInvalidStatus("operation is at an invalid status to proceed")
```

internal/core/workers/operation_execution_worker/operation_execution_worker.go
```diff
@@ -76,125 +79,158 @@ func (w *OperationExecutionWorker) Start(ctx context.Context) error {
...
+
+		return workererrors.NewErrStartOperationFailed("failed to start operation \"%s\" for the scheduler \"%s\"", op.ID, op.SchedulerName), nil
+	}
+
+	w.operationManager.StartLeaseRenewGoRoutine(operationContext, op)
+
+	executionErr = w.executeCollectingLatencyMetrics(op.DefinitionName, func() operations.ExecutionError {
+		return executor.Execute(operationContext, op, def)
+	})
+	return err, executionErr
```
arthur29, line 215: Hey @guilhermocc why had you done this one? It is kinda of weird.

internal/core/workers/operation_execution_worker/operation_execution_worker.go
```diff
@@ -76,125 +79,158 @@ func (w *OperationExecutionWorker) Start(ctx context.Context) error {
...
-			reportOperationExecutionWorkerFailed(w.scheduler.Game, w.scheduler.Name, LabelStartOperationFailed)
-			operationCancellationFunction()
+	executor, hasExecutor := w.executorsByName[def.Name()]
+	if w.shouldEvictOperation(op, def, hasExecutor, loopLogger) {
+		return nil
+	}
 
-			op.Status = operation.StatusError
-			w.operationManager.AppendOperationEventToExecutionHistory(ctx, op, fmt.Sprintf("Failed to grant lease to operation, reason: %s", err.Error()))
+	err, executionErr := w.startAndExecuteOperationWithLease(op, def, loopLogger, executor)
```
arthur29, line 115: What do you think to unify these errors in only one?

internal/core/workers/operation_execution_worker/operation_execution_worker.go
```diff
@@ -76,125 +79,158 @@ func (w *OperationExecutionWorker) Start(ctx context.Context) error {
...
+	if err != nil {
+		return err
+	}
 
-			err = w.operationManager.FinishOperation(ctx, op)
-			if err != nil {
-				loopLogger.Error("failed to finish operation", zap.Error(err))
-			}
+	w.treatExecutionError(op, def, executionErr, loopLogger, executor)
```
arthur29, line 120: This function is receiving too many arguments, what do you think to split them into two functions, the first one to handle the error and the second one to execute the rollback, in this way I think that you could decrease the number of arguments to these functions. Last but not least comment is to move these functions into the `if err != nil` so you could remove that first if in the `treatExecutionError`.

arthur29 (changes requested):

internal/core/workers/operation_execution_worker/operation_execution_worker.go
guilhermocc: err and executionErr are different errors, not sure how we would unify them

arthur29 (approved):
LGTM

PR #419, opened 2022-05-04
Fix/runtime watcher duplicated events

caiordjesus:
### What?
Implements client-go workqueue
### Why?
Maestro runtime watcher behavior is bad when there are many rooms being managed. We expect that with workqueue maestro have improved k8s usage

### DIFF
--- a/internal/adapters/runtime/kubernetes/game_room_watcher.go
+++ b/internal/adapters/runtime/kubernetes/game_room_watcher.go
@@ -141,13 +141,23 @@ func (kw *kubernetesWatcher) processEvent(eventType game_room.InstanceEventType,
 			return
 		}
 		// Caching before ready consists in lack of information about the instance.
-		if instance.Status.Type == game_room.InstanceReady {
+		if !(instance.Status.Type == game_room.InstanceUnknown || instance.Status.Type == game_room.InstancePending) {
 			kw.instanceCacheMap[pod.Name] = instance
 		}
 	} else {
 		if instance.ResourceVersion == pod.ResourceVersion {
 			return
 		}
+		tempInstance := &game_room.Instance{
+			ID:              instance.ID,
+			SchedulerID:     instance.SchedulerID,
+			Status:          convertPodStatus(pod),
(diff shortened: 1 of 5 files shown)


internal/adapters/runtime/kubernetes/game_room_watcher.go
```diff
@@ -200,5 +237,35 @@ func (k *kubernetes) WatchGameRoomInstances(ctx context.Context, scheduler *enti
 	})
 
 	go podsInformer.Run(watcher.stopChan)
+
+	// Same amount of routines reading from the channel
+	for i := 0; i < 200; i++ {
```
guilhermocc, line 242: nice

internal/adapters/runtime/kubernetes/game_room_watcher_test.go
```diff
@@ -25,231 +25,219 @@
...
-			select {
-			case _, ok := <-watcher.ResultChan():
-				return !ok
-			default:
-				return false
-			}
-		}, time.Second, 100*time.Millisecond)
-	})
-}
+//func TestGameRoomsWatch(t *testing.T) {
```
arthur29, line 28: Do you remember these comments here?

arthur29 (changes requested):

internal/adapters/runtime/kubernetes/game_room_watcher.go
```diff
@@ -200,5 +235,35 @@ func (k *kubernetes) WatchGameRoomInstances(ctx context.Context, scheduler *enti
...
```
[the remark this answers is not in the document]
caiordjesus: In the future, yes, of course
This number does not have a silver bullet, and in my opinion it should not be an environment variable, for example, since it's very hard to infer which number is okay
My approach would be to scale it based on the len of the workqueue. It the workqueue becomes too big, a new worker would be created
This thought, however, are not meant to be implemented now
There are more important features to look for (like self-healing and autoscaling). We should catch up in the near feature

PR #421, opened 2022-05-05
implements InsertPriorityOperationID on operation flow redis adapter

caiordjesus:
### What?
- Implement `InsertPriorityOperationID` method into operation flow redis adapter.
### Why?
- So we are able to add high-priority operations to the pending operations list (inserted at the top of the list).
### Refactor contained
- Running `make generate` updates swagger docs from a previous commit where this command hasn't been executed again after some upgrades.

### DIFF
--- a/internal/adapters/operation/operation_flow_redis_test.go
+++ b/internal/adapters/operation/operation_flow_redis_test.go
@@ -66,6 +66,39 @@ func TestInsertOperationID(t *testing.T) {
 	})
 }
 
+func TestInsertPriorityOperationID(t *testing.T) {
+	t.Run("with success", func(t *testing.T) {
+		client := test.GetRedisConnection(t, redisAddress)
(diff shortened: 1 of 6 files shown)


internal/adapters/operation/operation_flow_redis_test.go
```diff
@@ -66,6 +66,39 @@ func TestInsertOperationID(t *testing.T) {
...
+		client := test.GetRedisConnection(t, redisAddress)
+		flow := NewRedisOperationFlow(client)
+
+		// "drop" redis connection
+		client.Close()
+
+		err := flow.InsertPriorityOperationID(context.Background(), "", "")
+		require.ErrorIs(t, errors.ErrUnexpected, err)
+	})
+}
```
guilhermocc, line 100: Hey @caiordjesus aren't we using table-driven tests>

PR #422, opened 2022-05-05
Feature/create priority operation operation manager

caiordjesus:
### What?
- Implements CreatePriorityOperation Method in operationManager
### Why?
- To be able to add high-priority operations in the queue (should be executed before all others)

### Refactors included
Since `CreateOperation` and `CreatePriorityOperation` both creates operation, a constructor was added to the entity, so it won't repeat itself. Also, added simple creation test for this little guy (to prevent further alteration that might change the way it's supposed to work

### DIFF
--- a/internal/core/ports/operation_ports.go
+++ b/internal/core/ports/operation_ports.go
@@ -36,6 +36,9 @@ import (
 type OperationManager interface {
 	// CreateOperation creates a new operation for the given scheduler and includes it in the execution process.
 	CreateOperation(ctx context.Context, schedulerName string, definition operations.Definition) (*operation.Operation, error)
+	// CreatePriorityOperation creates a new priority operation (it'll run before other enqueued operation)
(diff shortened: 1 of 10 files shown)


internal/core/ports/operation_ports.go
```diff
@@ -36,6 +36,9 @@ import (
 type OperationManager interface {
 	// CreateOperation creates a new operation for the given scheduler and includes it in the execution process.
 	CreateOperation(ctx context.Context, schedulerName string, definition operations.Definition) (*operation.Operation, error)
+	// CreatePriorityOperation creates a new priority operation (it'll run before other enqueued operation)
+	//for the given scheduler and includes it in the execution process.
```
guilhermocc, line 40:
```suggestion
	// for the given scheduler and includes it in the execution process.
```

guilhermocc (approved):

PR #424, opened 2022-05-05
Improvement/implement reliable queue for operations

guilhermocc:
## What ❓
Implement a reliable queue pattern for operations. Every pending operation taken from the queue to be processed by the worker will now reside in an auxiliary queue before being taken from the channel so we can prevent operations loss if the worker goes down at that moment.

## Why 🤔  
Currently, we always pop an operation from the queue before actually executing it. If the worker component goes down in the meantime, these operations will no longer be in the queue and will remain in pending operation forever without lease data.


internal/adapters/operation/operation_flow_redis.go
```diff
@@ -68,36 +68,67 @@ func (r *redisOperationFlow) InsertOperationID(ctx context.Context, schedulerNam
...
+	metrics.RunWithMetrics(operationFlowStorageMetricLabel, func() error {
+		operationsIDsFromAux, err = r.client.LRange(ctx, r.buildSchedulerAuxiliaryPendingOperationsKey(schedulerName), 0, -1).Result()
+		return err
+	})
+	if err != nil && err != redis.Nil {
+		return nil, errors.NewErrUnexpected("failed to list pending operations for \"%s\"", schedulerName).WithError(err)
+	}
 
-	return operationsIDs, nil
+	return append(operationsIDs, operationsIDsFromAux...), nil
```
caiordjesus, line 131: Shouldn't operationIDsFromAux be in the front of operationIDs? This return put operations from the auxiliary list at the end of the line... Is this what we want here?

guilhermocc: It doesn't really matter the order of the IDs, since the list operation endpoint sorts operations by the createdAt.
But it makes sense to include ids from the aux in the beginning here

internal/adapters/operation/operation_flow_redis.go
```diff
@@ -80,37 +80,54 @@ func (r *redisOperationFlow) InsertPriorityOperationID(ctx context.Context, sche
...
+	if err != nil {
+		return "", errors.NewErrUnexpected("failed to fetch next operation ID from main queue").WithError(err)
+	}
+
+	return opID, nil
 }
 
-func (r *redisOperationFlow) ListSchedulerPendingOperationIDs(ctx context.Context, schedulerName string) (operationsIDs []string, err error) {
+// RemoveNextOperation removes the next operation from the operation flow.
+func (r *redisOperationFlow) RemoveNextOperation(ctx context.Context, schedulerName string) (err error) {
```
arthur29, line 104: What do you think to rename this function to something like: `FinishOperation` since it is removing the operation from the `auxiliary` queue and in the practice, it is finishing the operation to the Flow struct.

internal/adapters/operation/operation_flow_redis_test.go
```diff
@@ -270,3 +274,159 @@ func TestWatchOperationCancellationRequests(t *testing.T) {
 		}, 5*time.Second, 100*time.Millisecond)
 	})
 }
+
+func TestListSchedulerPendingOperationIDs(t *testing.T) {
+	type args struct {
+		ctx           context.Context
```
arthur29, line 280: The `ctx` as a less important argument that doesn't matter much to our test is commonly created inside the test. What do you think to remove this variable from the table?

internal/adapters/operation/operation_flow_redis_test.go
```diff
@@ -270,3 +274,159 @@ func TestWatchOperationCancellationRequests(t *testing.T) {
...
+		args             args
+		environmentSetup environmentSetup
+	}{
+		{"return no error and pops the operation from de auxiliary operations queue if there is some", args{
+			ctx:           context.Background(),
+			schedulerName: "test-scheduler",
+		},
+			environmentSetup{numberOfOps: 1, forceClientError: false},
+		},
+		{"return no error and does nothing if the auxiliary operations queue is empty", args{
```
arthur29, line 385: Hey in this case we should not send an error saying that the user is trying to remove an inexistent operation? It seems like an error.

arthur29 (changes requested):

internal/adapters/operation/operation_flow_redis.go
guilhermocc: We already have a FinishOperation method in OperationsManager, besides that, this method does not "finishes" the operation, it removes the operation from what we call "flow". The operation is finished when it finishes getting executed.

caiordjesus (approved):

internal/adapters/operation/operation_flow_redis_test.go
```diff
@@ -270,3 +274,166 @@ func TestWatchOperationCancellationRequests(t *testing.T) {
...
+		forceClientError bool
+	}
+	tests := []struct {
+		name             string
+		args             args
+		environmentSetup environmentSetup
+		errorExpected    error
+	}{
+		{
+			name: "return no error and pops the operation from de auxiliary operations queue if there is some",
```
arthur29, line 375:
```suggestion
			name: "return no error and pops the operation from the auxiliary operations queue if there is some",
```

arthur29 (approved):

PR #425, opened 2022-05-06
Feature/health controller operation

caiordjesus:
### What?
- Creates `health_controller` operation, that checks the health of a scheduler and takes actions to fix it.
### Why?
- We need a routine to keep looking to a Maestro Scheduler state and keep track of it, trying constantly to fix out-of-state problems


internal/core/operations/healthcontroller/health_controller_executor_test.go
```diff
@@ -0,0 +1,518 @@
...
+
+func TestSchedulerHealthController_Execute(t *testing.T) {
+	type ExecutionPlan struct {
+		PlanMocks func(
+			roomStorage *mockports.MockRoomStorage,
+			instanceStorage *ismock.MockGameRoomInstanceStorage,
+			schedulerStorage *mockports.MockSchedulerStorage,
+			operationManager *mockports.MockOperationManager,
+		)
+		ShouldFail bool
```
guilhermocc, line 54:
```suggestion
		planMocks func(
			roomStorage *mockports.MockRoomStorage,
			instanceStorage *ismock.MockGameRoomInstanceStorage,
			schedulerStorage *mockports.MockSchedulerStorage,
			operationManager *mockports.MockOperationManager,
		)
		shouldFail bool
```

guilhermocc: @caiordjesus Any reason for using exported fields here and below?

internal/core/operations/healthcontroller/health_controller_executor.go
```diff
@@ -0,0 +1,246 @@
...
+}
+
+func (ex *SchedulerHealthControllerExecutor) findAvailableAndExpiredRooms(ctx context.Context, op *operation.Operation, gameRoomsIDs []string) (availableRoomsIDs, expiredRoomsIDs []string) {
+	for _, gameRoomID := range gameRoomsIDs {
+		room, err := ex.roomStorage.GetRoom(ctx, op.SchedulerName, gameRoomID)
+		if err != nil {
+			continue
+		}
+
+		if ex.isRoomExpired(room) || ex.isRoomStatusError(room) {
```
guilhermocc, line 197: We are creating a new concept of an expired room in Next here. It is removing both rooms with expired ping and rooms with `error` status which seems to be a bit confusing to me. In maestro v9 we don't have the `error` which led us to the question? is a gru in `error` state an expired room?

IMO they are different things, a game room gets the `error` status when the instance reports the status error. Can the instance recover from this status? if not, the gru will stop sending ping and we will identify it as expired later.

internal/core/operations/healthcontroller/health_controller_executor.go
```diff
@@ -0,0 +1,246 @@
...
+			expiredRoomsIDs = append(expiredRoomsIDs, room.ID)
+			continue
+		}
+
+		if room.Status != game_room.GameStatusTerminating {
+			availableRoomsIDs = append(availableRoomsIDs, gameRoomID)
+		}
+	}
+
+	return availableRoomsIDs, expiredRoomsIDs
```
guilhermocc, line 207: here we are considering as available the room that doesn't have expired ping age, and is in one of the following status:
- GameStatusPending
- GameStatusUnready
- GameStatusOccupied
- GameStatusReady

Is everyone OK with considering these statuses as `available` rooms? i couldn't think in a better name, I'm just asking it here to see if anyone has more ideas. @caiordjesus @arthur29

internal/core/operations/healthcontroller/health_controller_executor_test.go
caiordjesus: Not really, very well observed

internal/core/operations/healthcontroller/health_controller_executor.go
caiordjesus: After thinking a lot about this, I could not think of a better name, honestly
And thinking a lot about those status, I think yes they should be considered available

internal/core/operations/healthcontroller/health_controller_executor.go
caiordjesus: You are totally right, just changed

internal/core/operations/healthcontroller/health_controller_executor_test.go
```diff
@@ -0,0 +1,559 @@
...
+	"github.com/topfreegames/maestro/internal/core/entities/game_room"
+	"github.com/topfreegames/maestro/internal/core/entities/operation"
+	"github.com/topfreegames/maestro/internal/core/operations/add_rooms"
+	"github.com/topfreegames/maestro/internal/core/operations/healthcontroller"
+	"github.com/topfreegames/maestro/internal/core/operations/remove_rooms"
+	mockports "github.com/topfreegames/maestro/internal/core/ports/mock"
+)
+
+func TestSchedulerHealthController_Execute(t *testing.T) {
+	type ExecutionPlan struct {
```
guilhermocc, line 47:
```suggestion
	type executionPlan struct {
```

guilhermocc (approved):

internal/core/operations/healthcontroller/health_controller_executor.go
```diff
@@ -0,0 +1,250 @@
...
+		room, err := ex.roomStorage.GetRoom(ctx, op.SchedulerName, gameRoomID)
+		if err != nil {
+			continue
+		}
+
+		if ex.isRoomStatusError(room) {
+			continue
+		}
+
+		if ex.isRoomExpired(room) {
```
arthur29, line 201: I think that here there is a possibility of the status of the room being `terminating`, so it should not be considered an expired one. WDYT?

arthur29: To fix this one I think that you should first ask whether the room is `terminating`, so not asks if it is expired.

arthur29, line 201: Another thread is that `pending` rooms should be considered available, but it should be asked for pings (since it is starting).

internal/core/operations/healthcontroller/health_controller_executor.go
```diff
@@ -0,0 +1,250 @@
...
+			Amount: removeAmount,
+		})
+		if err != nil {
+			return err
+		}
+		msgToAppend = fmt.Sprintf("created operation (id: %s) to remove %v rooms.", removeOperation.ID, removeAmount)
+	} else {
+		addAmount := desiredAmount - actualAmount
+		addOperation, err := ex.operationManager.CreatePriorityOperation(ctx, op.SchedulerName, &add_rooms.AddRoomsDefinition{
+			Amount: int32(addAmount),
```
arthur29, line 177: Are you joking that AddRooms and RemoveRooms use a different variable type to `Amount`? Let's create a technical debt for this?

arthur29 (changes requested):

internal/core/operations/healthcontroller/health_controller_executor.go
```diff
@@ -0,0 +1,250 @@
...
+		return operations.NewErrUnexpected(err)
+	}
+
+	nonexistentGameRoomIDs := ex.checkNonexistentGameRoomsIDs(gameRoomIDs, instances)
+	if len(nonexistentGameRoomIDs) > 0 {
+		logger.Error("found registered rooms that no longer exists")
+		ex.tryEnsureCorrectRoomsOnStorage(ctx, op, logger, nonexistentGameRoomIDs)
+	}
+
+	existentGameRoomIDs := difference(gameRoomIDs, nonexistentGameRoomIDs)
```
arthur29, line 84: What do you think to rename this function to something more semantic like `filterExistentGameRooms`

guilhermocc (approved):

arthur29 (approved):

PR #426, opened 2022-05-06
Remove Input and Execution History from ListOperations route

arthur29:
### Why
With the increase in the number of operations, ListOperations could become too long.

### What
- Remove `Input` and `ExecutionHistory` from LIstOperations route.


proto/api/v1/messages.proto
```diff
@@ -196,6 +196,23 @@ message SchedulerWithoutSpec {
...
+  // Unique identifier for the operation.
+  string id = 1;
+  // Current status of the operation.
+  // TODO(gabrielcorado): change to enum.
+  string status = 2;
+  // Name of the operation being executed.
+  string definition_name = 3;
+  // Operation lease. This is an optional field since not all operations have a lease.
+  optional Lease lease = 4;
+  // Scheduler identifier that the operation is from.
```
guilhermocc, line 210:
```suggestion
  // Operation scheduler identifier.
```

guilhermocc (approved):

caiordjesus (approved):

PR #429, opened 2022-05-11
Feature/forward terminating event when removing expired rooms

guilhermocc:
## What ❓ 
Forward terminating events when removing rooms in remove rooms operations.

## Why 🤔 
Remove rooms operation may be used to remove rooms that have expired ping, this means that these rooms will not be able to use maestro to forward terminating ping so that maestro will ensure this forwarding.

### DIFF
--- a/internal/core/operations/remove_rooms/remove_rooms_executor.go
+++ b/internal/core/operations/remove_rooms/remove_rooms_executor.go
@@ -26,6 +26,9 @@ import (
 	"context"
 	"errors"
 	"fmt"
+
+	"github.com/topfreegames/maestro/internal/core/entities/events"
+
 	"sync"
 
 	serviceerrors "github.com/topfreegames/maestro/internal/core/services/errors"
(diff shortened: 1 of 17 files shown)


internal/core/operations/remove_rooms/remove_rooms_executor.go
```diff
@@ -111,35 +116,36 @@ func (e *RemoveRoomsExecutor) removeRoomsByIDs(ctx context.Context, schedulerNam
...
 	}
 
 	return nil
 }
 
-func (e *RemoveRoomsExecutor) deleteRooms(ctx context.Context, rooms []*game_room.GameRoom) error {
+func (e *RemoveRoomsExecutor) deleteRooms(ctx context.Context, logger *zap.Logger, rooms []*game_room.GameRoom) error {
 	var err error
 	for _, room := range rooms {
 		err = e.roomManager.DeleteRoomAndWaitForRoomTerminated(ctx, room)
```
caiordjesus, line 144: Shouldn't we implement the `forwardingStatusTerminatingEvent` inside the `DeleteRoomAndWaitForRoomTerminated`? I think it'll be better since we should re-implement it on the `switch active version` operation too, creating 2 dependencies (events service) and code duplication

internal/core/services/room_manager/room_manager.go
```diff
@@ -163,6 +163,10 @@ func (m *RoomManager) DeleteRoomAndWaitForRoomTerminated(ctx context.Context, ga
 	timeoutContext, cancelFunc := context.WithTimeout(ctx, duration)
 	defer cancelFunc()
 	err = m.WaitRoomStatus(timeoutContext, gameRoom, game_room.GameStatusTerminating)
+	if err == nil {
```
caiordjesus, line 166: About the best practices, wdyt about:
```
if err != nil {
    return err
}
m.forwardStatusTerminatingEvent(ctx, gameRoom)
return nil
```

internal/core/operations/remove_rooms/remove_rooms_executor.go
```diff
@@ -26,6 +26,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+
```
caiordjesus:
```suggestion
```

caiordjesus (approved):

PR #430, opened 2022-05-11
Integrate health_controller with execution worker

arthur29:
### Why
Currently `health_controller` isn't being triggered to execute.

### What
- Add a ticker to control when `health_controller` executes
- Integrate `health_controller` in the `operation_execution_worker`

### DIFF
--- a/internal/core/workers/operation_execution_worker/operation_execution_worker.go
+++ b/internal/core/workers/operation_execution_worker/operation_execution_worker.go
@@ -81,6 +81,7 @@ func (w *OperationExecutionWorker) Start(ctx context.Context) error {
 	pendingOpsChan := w.operationManager.PendingOperationsChan(w.workerContext, w.scheduler.Name)
 
 	healthControllerTicker := time.NewTicker(w.config.GetDuration(healthControllerExecutionIntervalConfigPath))
+	defer healthControllerTicker.Stop()
 
 	for {
 		select {
--- a/internal/core/workers/operation_execution_worker/operation_execution_worker.go
+++ b/internal/core/workers/operation_execution_worker/operation_execution_worker.go
@@ -28,7 +28,6 @@ import (
 	"fmt"
 	"time"
 
-	"github.com/topfreegames/maestro/internal/config"
 	"github.com/topfreegames/maestro/internal/core/logs"
 	"github.com/topfreegames/maestro/internal/core/operations/healthcontroller"
 	workererrors "github.com/topfreegames/maestro/internal/core/workers/errors"
@@ -44,14 +43,12 @@ import (
 
 var _ workers.Worker = (*OperationExecutionWorker)(nil)
 
-const healthControllerExecutionIntervalConfigPath = "workers.operationExecution.healthControllerInterval"
-
 // OperationExecutionWorker is the service responsible for implementing the worker
 // responsibilities.
 type OperationExecutionWorker struct {
-	scheduler        *entities.Scheduler
-	config           config.Config
-	operationManager ports.OperationManager
+	scheduler                         *entities.Scheduler
+	healthControllerExecutionInterval time.Duration
+	operationManager                  ports.OperationManager
 	// TODO(gabrielcorado): check if we this is the right place to have all
 	// executors.
 	executorsByName     map[string]operations.Executor
@@ -61,13 +58,14 @@ type OperationExecutionWorker struct {
 	logger *zap.Logger
 }
 
+// NewOperationExecutionWorker instantiate a new OperationExecutionWorker to a specified scheduler.
 func NewOperationExecutionWorker(scheduler *entities.Scheduler, opts *workers.WorkerOptions) workers.Worker {
 	return &OperationExecutionWorker{
-		config:           opts.Config,
-		operationManager: opts.OperationManager,
-		executorsByName:  opts.OperationExecutors,
-		scheduler:        scheduler,
-		logger:           zap.L().With(zap.String(logs.LogFieldServiceName, "worker"), zap.String(logs.LogFieldSchedulerName, scheduler.Name)),
+		healthControllerExecutionInterval: opts.Configuration.HealthControllerExecutionInterval,
+		operationManager:                  opts.OperationManager,
+		executorsByName:                   opts.OperationExecutors,
+		scheduler:                         scheduler,
+		logger:                            zap.L().With(zap.String(logs.LogFieldServiceName, "worker"), zap.String(logs.LogFieldSchedulerName, scheduler.Name)),
 	}
 }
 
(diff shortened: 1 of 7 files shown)


internal/core/workers/operation_execution_worker/operation_execution_worker.go
```diff
@@ -42,10 +44,13 @@ import (
 
 var _ workers.Worker = (*OperationExecutionWorker)(nil)
 
+const healthControllerExecutionIntervalConfigPath = "workers.operationExecution.healthControllerInterval"
+
 // OperationExecutionWorker is the service responsible for implementing the worker
 // responsibilities.
 type OperationExecutionWorker struct {
 	scheduler        *entities.Scheduler
+	config           config.Config
```
guilhermocc, line 53: @arthur29 usually we dont use the config interface directly in our domain structs, instead, we create a type for the config, receive it in the constructor, and use an external provider that knows how to build it. This will make the domain struct independent from the viper lib. you can see some examples here internal/service/config.go.

internal/core/workers/operation_execution_worker/operation_execution_worker.go
```diff
@@ -74,6 +80,8 @@ func (w *OperationExecutionWorker) Start(ctx context.Context) error {
 	w.workerContext, w.cancelWorkerContext = context.WithCancel(ctx)
 	pendingOpsChan := w.operationManager.PendingOperationsChan(w.workerContext, w.scheduler.Name)
 
+	healthControllerTicker := time.NewTicker(w.config.GetDuration(healthControllerExecutionIntervalConfigPath))
+
```
guilhermocc, line 84:
```suggestion
	defer healthControllerTicker.Stop()

```

internal/core/workers/operation_execution_worker/operation_execution_worker_test.go
```diff
@@ -490,4 +522,95 @@ func TestSchedulerOperationsExecutionLoop(t *testing.T) {
...
+
+		scheduler := &entities.Scheduler{Name: "random-scheduler"}
+
+		executors := map[string]operations.Executor{}
+		executors[operationName] = operationExecutor
+		workerService := NewOperationExecutionWorker(scheduler, workers.ProvideWorkerOptions(config, operationManager, executors, nil, nil))
+		pendingOpsChan := make(chan string)
+
+		operationManager.EXPECT().PendingOperationsChan(gomock.Any(), gomock.Any()).Return(pendingOpsChan)
+		operationManager.EXPECT().CreateOperation(gomock.Any(), gomock.Any(), gomock.Any()).Do(func(_ context.Context, schedulerName string, op operations.Definition) {
```
guilhermocc, line 552:
```suggestion
		operationManager.EXPECT().CreateOperation(gomock.Any(), scheduler.Name, &healthcontroller.SchedulerHealthControllerDefinition{}).Return(&operation.Operation{}, nil).AnyTimes()
```
no need to use Do here, right?

internal/core/workers/operation_execution_worker/operation_execution_worker_test.go
```diff
@@ -490,4 +522,95 @@ func TestSchedulerOperationsExecutionLoop(t *testing.T) {
...
+			require.Equal(t, scheduler.Name, schedulerName)
+			require.IsType(t, &healthcontroller.SchedulerHealthControllerDefinition{}, op)
+		}).Return(&operation.Operation{}, nil).AnyTimes()
```
guilhermocc, line 555: No need to use Do() here, right?

```suggestion
operationManager.EXPECT().CreateOperation(gomock.Any(), scheduler.Name, &healthcontroller.SchedulerHealthControllerDefinition{}).Return(&operation.Operation{}, nil).AnyTimes()

```

guilhermocc: Besides that, what about using **MaxTimes(5)** instead of **AnyTimes()**? we wait 5ms for the test to run and the period is 1ms, so it should never trigger the healthcontroller more than 5 times.

internal/core/workers/operation_execution_worker/operation_execution_worker_test.go
```diff
@@ -490,4 +522,95 @@ func TestSchedulerOperationsExecutionLoop(t *testing.T) {
...
+		scheduler := &entities.Scheduler{Name: "random-scheduler"}
+
+		executors := map[string]operations.Executor{}
+		executors[operationName] = operationExecutor
+		workerService := NewOperationExecutionWorker(scheduler, workers.ProvideWorkerOptions(config, operationManager, executors, nil, nil))
+		pendingOpsChan := make(chan string)
+
+		operationManager.EXPECT().PendingOperationsChan(gomock.Any(), gomock.Any()).Return(pendingOpsChan)
+		executionCount := 0
+		operationManager.EXPECT().CreateOperation(gomock.Any(), gomock.Any(), gomock.Any()).DoAndReturn(func(_ context.Context, schedulerName string, op operations.Definition) (*operation.Operation, error) {
```
guilhermocc, line 595: Again, no need to use Do() here, we can assert the types directly replacing de gomock.Anys() and just return an error in every call, the **executionCount** var is not being used here.

guilhermocc (approved):

internal/core/workers/operation_execution_worker/operation_execution_worker.go
```diff
@@ -88,6 +95,11 @@ func (w *OperationExecutionWorker) Start(ctx context.Context) error {
 			if err != nil && shouldFinishWorker(err) {
 				return err
 			}
+		case <-healthControllerTicker.C:
+			err := w.createHealthControllerOperation(w.workerContext)
```
caiordjesus, line 99: Shouldn't we check if there is another health_controller operation on the queue? Looks to me that simply creating a new operation when the ticker rings could lead to a case where we have like 2 health operations enqueued

internal/service/config.go
```diff
@@ -46,28 +50,40 @@ func NewRoomManagerConfig(c config.Config) (room_manager.RoomManagerConfig, erro
 	return roomManagerConfig, nil
 }
 
+// NewWorkersConfig instantiate a new workers Config stucture to be used by the workers to customize them from the config package.
+func NewWorkersConfig(c config.Config) (workers.Configuration, error) {
+	healthControllerExecutionInterval := c.GetDuration(healthControllerExecutionIntervalConfigPath)
```
caiordjesus, line 55: Wdyt about having a default value for this little guy? It could simplify maestro deploy for some users

internal/core/workers/operation_execution_worker/operation_execution_worker.go
arthur29: There is no problem if this happen, because in this case the first one could generate a change in the scheduler, otherwise it will only check and finish (because the cluster will be in the desired state)

internal/service/config.go
arthur29: Yeah, I added in the `config/worker.local.yaml`

caiordjesus (approved):

PR #435, opened 2022-05-16
Fix list scheduler active operations to not return error when an oper…

arthur29:
### WHY
`ListSchedulerActiveOperations` in the Redis adapter returned an error when an operation was expired. This PR proposes to avoid errors when it is `redis.Nil`.
### WHAT
- Change the Redis operation storage adapter to avoid returning an error when it is `redis.Nil`
- Use the Redis pipeline to avoid multiple calls to the Redis and improve latency.

### DIFF
--- a/internal/adapters/operation/operation_storage_redis.go
+++ b/internal/adapters/operation/operation_storage_redis.go
@@ -190,7 +190,7 @@ func (r *redisOperationStorage) ListSchedulerActiveOperations(ctx context.Contex
 	})
 
 	if err != nil {
-		return nil, errors.NewErrUnexpected("failed execute pipe for retrieving schedulers").WithError(err)
+		return nil, errors.NewErrUnexpected("failed execute pipe for retrieving operations").WithError(err)
 	}
 	operations = make([]*operation.Operation, 0, len(operationsIDs))
 	for _, cmder := range cmders {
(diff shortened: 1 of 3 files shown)


caiordjesus (approved):

internal/adapters/operation/operation_storage_redis.go
```diff
@@ -177,14 +177,37 @@ func (r *redisOperationStorage) ListSchedulerActiveOperations(ctx context.Contex
...
+	}
+
+	var cmders []redis.Cmder
+	metrics.RunWithMetrics(operationStorageMetricLabel, func() error {
+		cmders, err = pipe.Exec(ctx)
+		return err
+	})
+
+	if err != nil {
+		return nil, errors.NewErrUnexpected("failed execute pipe for retrieving schedulers").WithError(err)
```
guilhermocc, line 193:
```suggestion
		return nil, errors.NewErrUnexpected("failed execute pipe for retrieving operations").WithError(err)
```

guilhermocc (approved):

internal/adapters/operation/operation_storage_redis.go
caiordjesus: Actually, this is a blocking change :0
Did not see it!

PR #436, opened 2022-05-20
Limit finished operations query for last day only

guilhermocc:
## What ❓ 
Limit get finish operations to only fetch operations since last day.

## Why 🤔 
Currently, without pagination and filtering on get operations endpoint, our queries can cause an overload on management-api


arthur29 (approved):
LGTM

caiordjesus (approved):

PR #441, opened 2022-05-23
Update add rooms operation to not wait for rooms readiness

guilhermocc:
## What ❓ 
Update add rooms operation to not wait for readiness and remove rollback logic

## Why 🤔 
The add rooms operation uses the CreateRoomAndWaitForReadiness method for creating rooms, this method, as the name suggests, creates the game room and blocks the execution, waiting for the room to reach the ready state (when the room sends the ping). This behavior can make add rooms operations take a lot of time if this waiting period reaches the roomInitializationTimout period. For a scheduler with a health checker or autoscaling enabled, having add rooms operation as a blocking operation can create scenarios in which the maestro will take too long to reach the desired state.

Besides all that, the current rollback logic of adding rooms consists in removing rooms that were created, this behavior also needs to be updated for a health-checker/autoscaling scenario. The rollback logic of add rooms should be empty, in a way that Maestro will keep rooms created successfully if some error occurs in the middle of add rooms process.

### DIFF
--- a/internal/core/operations/add_rooms/add_rooms_executor.go
+++ b/internal/core/operations/add_rooms/add_rooms_executor.go
@@ -26,14 +26,12 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"sync"
 
 	"github.com/topfreegames/maestro/internal/core/logs"
 
 	"github.com/topfreegames/maestro/internal/core/ports"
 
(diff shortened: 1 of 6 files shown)


caiordjesus (approved):
It's bizarre how we are simplifying the whole operation now by a lot 😆

PR #442, opened 2022-05-23
Fix/expired rooms event forwarding

guilhermocc:
## What ❓ 
Forward **RoomStatus** event instead of **RoomEvent** when deleting rooms.

## Why 🤔 
In maestro v9, the event type that maestro forward when deleting expired rooms is **RoomStatus** with **terminated** state

### DIFF
--- a/internal/core/entities/events/player_event.go
+++ b/internal/core/entities/events/player_event.go
@@ -40,9 +40,9 @@ const (
 
 func ConvertToPlayerEventType(value string) (PlayerEventType, error) {
 	switch value {
-	case "playerLeft":
+	case string(PlayerLeft):
 		return PlayerLeft, nil
-	case "playerJoin":
+	case string(PlayerJoin):
 		return PlayerJoin, nil
 	default:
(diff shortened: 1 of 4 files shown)


internal/core/entities/events/player_event.go
```diff
@@ -40,9 +40,9 @@ const (
 
 func ConvertToPlayerEventType(value string) (PlayerEventType, error) {
 	switch value {
-	case "playerLeft":
+	case string(PlayerLeft):
 		return PlayerLeft, nil
-	case "playerJoin":
+	case string(PlayerJoin):
```
guilhermocc, line 45: just refactor

internal/core/entities/events/room_event.go
```diff
@@ -53,11 +53,11 @@ const (
 
 func ConvertToRoomEventType(value string) (RoomEventType, error) {
 	switch value {
-	case "resync":
+	case string(Ping):
```
guilhermocc, line 56: just refactor

caiordjesus (approved):

PR #443, opened 2022-05-23
Fix e2e tests after health controller

arthur29:
### Why

The health controller changes how maestro works, time spaced it creates a new operation and many test scenarios have changed.

### What

This PR proposes to fix these end-to-end scenarios and tests to reflect this new way that the maestro works.


e2e/suites/management/create_new_scheduler_version_test.go
```diff
@@ -91,7 +91,7 @@ func TestCreateNewSchedulerVersion(t *testing.T) {
 
 			err = managementApiClient.Do("GET", fmt.Sprintf("/schedulers/%s", scheduler.Name), getSchedulerRequest, getSchedulerResponse)
 			require.NoError(t, err)
-			require.Equal(t, "v1.0.0", getSchedulerResponse.Scheduler.Spec.Version)
+			require.Equal(t, "v1.1.0", getSchedulerResponse.Scheduler.Spec.Version)
```
caiordjesus, line 94: Why is this version  v1.1.0? Since we just created the scheduler and the create new version failed, it should keep on version v1.0.0, no?

e2e/suites/management/create_new_scheduler_version_test.go
```diff
@@ -201,7 +211,7 @@ func TestCreateNewSchedulerVersion(t *testing.T) {
 			require.NoError(t, err)
 
 			// No version was created
-			require.Len(t, getVersionsResponse.Versions, 1)
+			require.Len(t, getVersionsResponse.Versions, 2)
```
caiordjesus, line 214: Since we have 2 now, a version was created
Maybe this test is showing a failure on maestro?

e2e/suites/management/utils.go
```diff
@@ -291,15 +319,48 @@ func createSchedulerWithRoomsAndWaitForIt(t *testing.T, maestro *maestro.Maestro
...
 
-	// Add rooms to the created scheduler
-	// TODO(guilhermecarvalho): when autoscaling is implemented, this part of the test can be deleted
-	addRoomsRequest := &maestroApiV1.AddRoomsRequest{SchedulerName: scheduler.Name, Amount: 2}
-	addRoomsResponse := &maestroApiV1.AddRoomsResponse{}
-	err = managementApiClient.Do("POST", fmt.Sprintf("/schedulers/%s/add-rooms", scheduler.Name), addRoomsRequest, addRoomsResponse)
+	roomsReplicas := int32(2)
+	patchSchedulerRequest := &maestroApiV1.PatchSchedulerRequest{RoomsReplicas: &roomsReplicas}
+	patchSchedulerResponse := &maestroApiV1.PatchSchedulerResponse{}
+	err = managementApiClient.Do("PATCH", fmt.Sprintf("/schedulers/%s", scheduler.Name), patchSchedulerRequest, patchSchedulerResponse)
```
caiordjesus, line 329: Now I understand why the schedulers are starting at v1.1.0 🌝

caiordjesus (approved):

e2e/suites/management/create_new_scheduler_version_test.go
caiordjesus: Self-answered in my last comment
I would recommend making it more clear that the scheduler returned has the version v1.1.0, but the way it is, works!

e2e/suites/management/utils.go
guilhermocc: hehe

guilhermocc: they are still staring at v1.0.0 👀

guilhermocc (approved):

PR #447, opened 2022-05-26
Implement roomoccupancy policy to autoscaler

arthur29:
### Why

To autoscaling works well it needs policies to execute and calculate the desired number of rooms that `healthcontroller` should create/delete. Initially, only `roomoccupancy` policy will be implemented, and that consists of getting the number of occupied rooms and calculating the desired total number of game rooms. The formula for this will be:
```
totalNumberOfRoom = ROUND UP [occupiedRooms / (1 - readyTarget)]
```

Note: `readyTarget` must be between 0 and 1.

### What

To support multiple policies an `autoscaler` `Policy` interface was created. It consists of two functions `CurrentStateBuilder` which fetches current data of the `Scheduler` and put them together to be used by the autoscaler and `CalculateDesiredNumberOfRooms` which executes the formula and returns how many rooms that cluster should have.
It implements tests for these functions, as well.


internal/core/services/autoscaler/policies/roomoccupancy/policy.go
```diff
@@ -0,0 +1,88 @@
...
+func (p *Policy) CurrentStateBuilder(ctx context.Context, scheduler *entities.Scheduler) (policies.CurrentState, error) {
+	occupiedRoomsAmount, err := p.roomStorage.GetRoomCountByStatus(ctx, scheduler.Name, game_room.GameStatusOccupied)
+	if err != nil {
+		return nil, fmt.Errorf("error fetching occupied game rooms amount: %w", err)
+	}
+
+	readyTarget := scheduler.Autoscaling.Policy.Parameters.RoomOccupancy.ReadyTarget
+
+	currentState := policies.CurrentState{
+		ReadyTargetKey:   readyTarget,
```
guilhermocc, line 70: Ready target isn't really part of the current state, right? it is an autoscaling policy parameter. Why configure it here since we already have access to it in `Autoscaling` struct configuration? 

What about updating the CalculateDesiredNumberOfRooms method to receive the current state and the **policy parameters** since all policies will have to use its parameter to calculate the desired number of rooms:

`CalculateDesiredNumberOfRooms(currentState policies.CurrentState, policyParameters PolicyParameters) (int, error)`

internal/core/services/autoscaler/policies/roomoccupancy/policy_test.go
```diff
@@ -0,0 +1,190 @@
...
+				Parameters: autoscaling.PolicyParameters{
+					RoomOccupancy: &autoscaling.RoomOccupancyParams{
+						ReadyTarget: readyTarget,
+					},
+				},
+			},
+		},
+	}
+
+	t.Run("Success cases - when all gonna ok", func(t *testing.T) {
```
guilhermocc, line 64:
```suggestion
	t.Run("Success cases - when no error occurs it builds the state with occupied rooms amount", func(t *testing.T) {
```

internal/core/services/autoscaler/policies/roomoccupancy/policy_test.go
```diff
@@ -0,0 +1,190 @@
...
+		policy := roomoccupancy.NewPolicy(roomStorageMock)
+
+		currentState, err := policy.CurrentStateBuilder(context.Background(), scheduler)
+		assert.NoError(t, err)
+
+		assert.Equal(t, readyTarget, currentState[roomoccupancy.ReadyTargetKey])
+		assert.Equal(t, occupiedRoomsAmount, currentState[roomoccupancy.OccupiedRoomsKey])
+	})
+
+	t.Run("Error case - When GetRoomCountByStatus return in error returns nil error", func(t *testing.T) {
```
guilhermocc, line 78:
```suggestion
	t.Run("Error case - When some error occurs in GetRoomCountByStatus it returns error", func(t *testing.T) {
```

internal/core/services/autoscaler/policies/state.go
```diff
@@ -0,0 +1,26 @@
...
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+package policies
+
+// CurrentState is a map that will be used by policies to calculate the desired number of rooms to a scheduler.
```
guilhermocc, line 25:
```suggestion
// CurrentState is a map that represents the current state of a scheduler and can be populated with arbitrary information (rooms count, instances status, etc)
// it will be used by policies to calculate the desired number of rooms to a scheduler.
```

internal/core/services/autoscaler/policies/roomoccupancy/policy.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+// CurrentStateBuilder fill the fields that should be considered during the autoscaling policy.
+func (p *Policy) CurrentStateBuilder(ctx context.Context, scheduler *entities.Scheduler) (policies.CurrentState, error) {
+	occupiedRoomsAmount, err := p.roomStorage.GetRoomCountByStatus(ctx, scheduler.Name, game_room.GameStatusOccupied)
+	if err != nil {
+		return nil, fmt.Errorf("error fetching occupied game rooms amount: %w", err)
+	}
+
+	readyTarget := scheduler.Autoscaling.Policy.Parameters.RoomOccupancy.ReadyTarget
```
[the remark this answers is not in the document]
arthur29: With the previous change to move parameters outside the CurrentState, this comment does not apply anymore.

guilhermocc (approved):

PR #448, opened 2022-05-26
Chore/create autoscaler service

arthur29:
### Why

To autoscaling execute it needs a service that is responsible for injecting its dependency into the `healthcontroller` and executing its logic.

### What

Create a package called `autoscaler` with `Autoscaler` struct that executes the logic to calculate the desired number of rooms in the scheduler.

### DIFF
--- a/internal/core/entities/autoscaling/autoscaling.go
+++ b/internal/core/entities/autoscaling/autoscaling.go
@@ -29,7 +29,7 @@ type PolicyType string
 
 const (
 	// RoomOccupancy is an implemented policy in maestro autoscaler,
-	// it uses the number of occupied rooms to calculate the desired number of rooms in a scheduler.
+	// it uses the number of occupied rooms and a ready rooms target percentage to calculate the desired number of rooms in a scheduler.
 	RoomOccupancy PolicyType = "roomOccupancy"
 )
 
--- a/internal/core/entities/autoscaling/autoscaling.go
+++ b/internal/core/entities/autoscaling/autoscaling.go
@@ -47,7 +47,7 @@ type Autoscaling struct {
 	Policy Policy
 }
 
-// Validate check if an Autoscaling struct is well formated.
+// Validate check if an Autoscaling struct is well formatted and contains valid values.
 func (a *Autoscaling) Validate() error {
 	return validations.Validate.Struct(a)
 }
--- a/internal/core/entities/autoscaling/autoscaling.go
+++ b/internal/core/entities/autoscaling/autoscaling.go
@@ -24,6 +24,15 @@ package autoscaling
 
 import "github.com/topfreegames/maestro/internal/validations"
 
+// PolicyType represents an enum of possible policy types/strategies a scheduler can have.
+type PolicyType string
+
+const (
+	// RoomOccupancy is an implemented policy in maestro autoscaler,
+	// it uses the number of occupied rooms to calculate the desired number of rooms in a scheduler.
+	RoomOccupancy PolicyType = "roomOccupancy"
+)
+
 // Autoscaling represents the autoscaling configuration for a scheduler.
 type Autoscaling struct {
 	// Enabled indicates if autoscaling is enabled.
@@ -38,10 +47,12 @@ type Autoscaling struct {
 	Policy Policy
 }
 
+// Validate check if an Autoscaling struct is well formated.
 func (a *Autoscaling) Validate() error {
 	return validations.Validate.Struct(a)
 }
 
+// NewAutoscaling instantiates a new autoscaling struct based on its parameters.
 func NewAutoscaling(enabled bool, min, max int, policy Policy) (*Autoscaling, error) {
 	autoscaling := &Autoscaling{
 		Enabled: enabled,
@@ -68,14 +79,8 @@ type PolicyParameters struct {
 	RoomOccupancy *RoomOccupancyParams `validate:"required_for_room_occupancy=Type"`
 }
 
+// RoomOccupancyParams represents the parameters accepted by rooms occupancy autoscaling properties.
 type RoomOccupancyParams struct {
 	// ReadyTarget indicates the target percentage of ready rooms a scheduler should maintain.
 	ReadyTarget float64 `validate:"gt=0,lt=1"`
 }
-
-// PolicyType represents an enum of possible policy types/strategies a scheduler can have.
-type PolicyType string
-
-const (
-	RoomOccupancy PolicyType = "roomOccupancy"
-)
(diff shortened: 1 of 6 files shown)


internal/core/entities/autoscaling/autoscaling.go
```diff
@@ -24,6 +24,15 @@ package autoscaling
 
 import "github.com/topfreegames/maestro/internal/validations"
 
+// PolicyType represents an enum of possible policy types/strategies a scheduler can have.
+type PolicyType string
+
+const (
+	// RoomOccupancy is an implemented policy in maestro autoscaler,
+	// it uses the number of occupied rooms to calculate the desired number of rooms in a scheduler.
```
guilhermocc, line 32:
```suggestion
	// it uses the number of occupied rooms and a ready rooms target percentage to calculate the desired number of rooms in a scheduler.
```

internal/core/entities/autoscaling/autoscaling.go
```diff
@@ -38,10 +47,12 @@ type Autoscaling struct {
 	Policy Policy
 }
 
+// Validate check if an Autoscaling struct is well formated.
```
guilhermocc, line 50:
```suggestion
// Validate check if an Autoscaling struct is well formatted and contains valid values.
```

internal/core/services/autoscaler/autoscaler.go
```diff
@@ -0,0 +1,88 @@
...
+		roomStorage: roomStorage,
+	}
+
+	autoscaler.PolicyFactory = autoscaler.DefaultPolicyFactory
+
+	return autoscaler
+}
+
+// CalculateDesiredNumberOfRooms return in success case the number of rooms that a Scheduler should have based on its policy,
+// in error cases it returns nil and the error.
```
guilhermocc, line 59:
```suggestion
// CalculateDesiredNumberOfRooms return the number of rooms that a Scheduler should have based on its policy or error if it can calculate.
```

internal/core/services/autoscaler/autoscaler.go
```diff
@@ -0,0 +1,88 @@
...
+	if err != nil {
+		return -1, fmt.Errorf("Error fetching current state to scheduler %s: %w", scheduler.Name, err)
+	}
+
+	desiredNumberOfRooms, err := policy.CalculateDesiredNumberOfRooms(currentState)
+	if err != nil {
+		return -1, fmt.Errorf("Error calculating the desired number of rooms to scheduler %s: %w", scheduler.Name, err)
+	}
+
+	return desiredNumberOfRooms, nil
```
guilhermocc, line 76: We must enforce the calculated desiredNumberOfRooms value is between autoscaling configured min/max values

internal/core/services/autoscaler/autoscaler_test.go
```diff
@@ -0,0 +1,145 @@
...
+				PolicyFactory: func(autoscaling.PolicyType) (autoscalerPorts.Policy, error) {
+					return nil, errors.New("Error getting policy to policyType")
+				},
+			}
+
+			_, err := autoscaler.CalculateDesiredNumberOfRooms(context.Background(), scheduler)
+			assert.ErrorContains(t, err, "Error building policy to scheduler")
+		})
+
+		t.Run("When CurrentStateBuilder returns in error", func(t *testing.T) {
```
guilhermocc, line 88:
```suggestion
		t.Run("When CurrentStateBuilder returns error", func(t *testing.T) {
```

internal/core/services/autoscaler/autoscaler_test.go
```diff
@@ -0,0 +1,145 @@
...
+				PolicyFactory: func(autoscaling.PolicyType) (autoscalerPorts.Policy, error) {
+					return mockPolicy, nil
+				},
+			}
+
+			_, err := autoscaler.CalculateDesiredNumberOfRooms(context.Background(), scheduler)
+			assert.ErrorContains(t, err, "Error fetching current state to scheduler")
+		})
+
+		t.Run("When CalculateDesiredNumberOfRooms returns in error", func(t *testing.T) {
```
guilhermocc, line 103:
```suggestion
		t.Run("When CalculateDesiredNumberOfRooms returns error", func(t *testing.T) {
```

guilhermocc (changes requested):
We must enforce the calculated desiredNumberOfRooms value is between autoscaling configured min/max values

internal/core/services/autoscaler/autoscaler.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+// DefaultPolicyFactory is a function that is used by the the autoscaler to return which policy should be used
```
guilhermocc, line 79:
```suggestion
// DefaultPolicyFactory is a function that is used by the autoscaler to return which policy should be used
```

internal/core/services/autoscaler/autoscaler.go
arthur29: Done

internal/core/services/autoscaler/autoscaler.go
```diff
@@ -0,0 +1,87 @@
...
+	"errors"
+	"fmt"
+
+	"github.com/topfreegames/maestro/internal/core/entities"
+	"github.com/topfreegames/maestro/internal/core/entities/autoscaling"
+	autoscalerPorts "github.com/topfreegames/maestro/internal/core/ports/autoscaler"
+)
+
+// PolicyFactory is a type that corelates a policy type with an autoscaling policy.
+type PolicyFactory map[autoscaling.PolicyType]autoscalerPorts.Policy
```
guilhermocc, line 36: @arthur29 this is not actually a factory, right? what about renaming it to
```suggestion
type PolicyMap map[autoscaling.PolicyType]autoscalerPorts.Policy
```

guilhermocc (approved):

internal/core/services/autoscaler/autoscaler.go
```diff
@@ -0,0 +1,87 @@
...
+	policy := a.policyFactory[scheduler.Autoscaling.Policy.Type]
+
+	currentState, err := policy.CurrentStateBuilder(ctx, scheduler)
+	if err != nil {
+		return -1, fmt.Errorf("Error fetching current state to scheduler %s: %w", scheduler.Name, err)
+	}
+
+	desiredNumberOfRooms, err := policy.CalculateDesiredNumberOfRooms(scheduler.Autoscaling.Policy.Parameters, currentState)
+	if err != nil {
+		return -1, fmt.Errorf("Error calculating the desired number of rooms to scheduler %s: %w", scheduler.Name, err)
```
caiordjesus, line 71: Very tiny detail: the pattern Maestro has now consists of non-capitalized errors (like all of these errors start with a Capitalized letter)

internal/core/services/autoscaler/autoscaler.go
caiordjesus: @guilhermocc your PR open right now will change it, right?

caiordjesus (approved):

PR #449, opened 2022-05-27
Fix/expired pending rooms

guilhermocc:
## What ❓ 
Add CreatedAt field in GameRoom struct, update room storage db adapter to populate and retrieve this new field.

Update health controller operation logic to check for pending rooms with age bigger than **roomInitializationTimeout** to consider them as expired.

## Why 🤔 
Since add rooms operation doesn't wait for rooms to reach ready state anymore, we can have scenarios of game rooms stuck in `pending` state. We have to update our health controller to be able to identify and remove those rooms.

### DIFF
--- a/internal/core/operations/healthcontroller/health_controller_executor.go
+++ b/internal/core/operations/healthcontroller/health_controller_executor.go
@@ -200,9 +200,7 @@ func (ex *SchedulerHealthControllerExecutor) findAvailableAndExpiredRooms(ctx co
 		}
 
 		switch {
-		case ex.isRoomStatus(room, game_room.GameStatusUnready):
-			availableRoomsIDs = append(availableRoomsIDs, gameRoomID)
-		case ex.isPendingRoomExpired(room):
+		case ex.isInitializingRoomExpired(room):
 			expiredRoomsIDs = append(expiredRoomsIDs, gameRoomID)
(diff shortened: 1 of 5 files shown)


internal/core/operations/healthcontroller/health_controller_executor.go
```diff
@@ -200,12 +200,12 @@ func (ex *SchedulerHealthControllerExecutor) findAvailableAndExpiredRooms(ctx co
 		}
 
 		switch {
-		case ex.isRoomStatus(room, game_room.GameStatusPending):
-			availableRoomsIDs = append(availableRoomsIDs, gameRoomID)
 		case ex.isRoomStatus(room, game_room.GameStatusUnready):
 			availableRoomsIDs = append(availableRoomsIDs, gameRoomID)
```
caiordjesus, line 204: Since the logic seems to test if the room is expired somehow, and the non-expired are considered available, why do we need to check if the status is unready? Seems to me that we could take off the unready condition 🤔

caiordjesus (approved):

internal/core/operations/healthcontroller/health_controller_executor.go
guilhermocc: We have to consider unready rooms in the initialization timeout check as well!

PR #450, opened 2022-05-30
Feature/integrate health controller with autoscaler

guilhermocc:
## What ❓ 
Integrate health controller with autoscaler service by creating a new dependency, and using the autoscaler service for calculating the desired number of rooms if the autoscaling is configured and enabled for the given scheduler.

## Why 🤔 
health_controller operation will now use configured autoscaling policies for calculating the desired number of rooms. This will basically enable schedulers to perform automatic scaling based on user-configured policies.


internal/service/adapters.go
```diff
@@ -211,6 +208,15 @@ func NewOperationFlowRedis(c config.Config) (ports.OperationFlow, error) {
...
 
+func createRedisClient(url string) (*redis.Client, error) {
+	opts, err := redis.ParseURL(url)
+	if err != nil {
+		return nil, fmt.Errorf("invalid redis URL: %w", err)
+	}
+
+	return redis.NewClient(opts), nil
+}
+
```
guilhermocc, line 219: Just moved the function, no changes

caiordjesus (approved):

PR #466, opened 2022-06-10
Chore/adds validation for forwarder options nil

caiordjesus:
### What?
Changes how forwarder is initialized with default options

### Why?
Although options are optional, maestro needs a default value for some options when forwarding pings


internal/core/entities/forwarder/forwarder.go
```diff
@@ -35,3 +35,16 @@ type Forwarder struct {
...
+	if options == nil {
+		options = NewDefaultForwarderOptions()
+	}
+	return &Forwarder{
+		Name:        name,
+		Enabled:     enabled,
+		ForwardType: forwardType,
+		Address:     address,
+		Options:     options,
+	}
```
arthur29, line 49: About the happy path be always on the right. I suggest to:
```suggestion
        forwarder := &Forwarder{
		Name:        name,
		Enabled:     enabled,
		ForwardType: forwardType,
		Address:     address,
		Options:     NewDefaultForwarderOptions(),
	}
	
	if options != nil {
		forwarder.Options = options
	}
	return 
```

PR #468, opened 2022-06-13
Ensure worker stops when resultChan is closed

guilhermocc:
## What ❓ 
Ensure runtime watcher stops when resultChan is closed

## Why 🤔 
Currently, if resultChan is closed, the runtime watcher worker hangs forever until someone cancels the context.

### DIFF
--- a/internal/core/workers/runtime_watcher_worker/runtime_watcher_worker.go
+++ b/internal/core/workers/runtime_watcher_worker/runtime_watcher_worker.go
@@ -25,6 +25,7 @@ package runtime_watcher_worker
 import (
 	"context"
 	"fmt"
+	"sync"
 
 	"github.com/topfreegames/maestro/internal/core/logs"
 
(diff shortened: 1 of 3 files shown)


internal/core/workers/runtime_watcher_worker/runtime_watcher_worker.go
```diff
@@ -71,20 +72,27 @@ func (w *runtimeWatcherWorker) Start(ctx context.Context) error {
...
 						return
 					}
-					err := w.processEvent(w.ctx, event)
+					err = w.processEvent(w.ctx, event)
 					if err != nil {
 						w.logger.Warn("failed to process event", zap.Error(err))
+						reportEventProcessingStatus(event, false)
+					} else {
+						reportEventProcessingStatus(event, true)
 					}
```
arthur29, line 96:
```suggestion
					}
					
					reportEventProcessingStatus(event, true)
```

arthur29 (approved):
LGTM

internal/core/workers/runtime_watcher_worker/runtime_watcher_worker.go
guilhermocc: we can not do this because the function is not returning I fail branch

PR #472, opened 2022-06-17
Chore/enhance switch active version experience at large scale

caiordjesus:
### What?
Enhanced experience of `switch_active_version` operation with retry and progress

### Why?
On a large scale, this operation takes a lot of time and sometimes fails under circumstances it should not (like listing rooms with deletion priority, which could be retried), or trying to delete an already deleted room. Also, waiting an hour without knowing what is happening inside represents a very poor experience, which can simply be enhanced with operation progress.

### Screenshot
image

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,6 +5,7 @@ go 1.17
 require (
 	github.com/Masterminds/semver v1.5.0
 	github.com/Masterminds/semver/v3 v3.1.1
+	github.com/avast/retry-go/v4 v4.1.0
 	github.com/bufbuild/buf v0.47.0
 	github.com/ghodss/yaml v1.0.0
 	github.com/go-pg/pg v6.13.2+incompatible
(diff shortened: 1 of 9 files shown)


internal/core/operations/switch_active_version/switch_active_version_executor.go
```diff
@@ -155,11 +160,21 @@ func (ex *SwitchActiveVersionExecutor) deleteNewCreatedRooms(ctx context.Context
...
 
-func (ex *SwitchActiveVersionExecutor) startReplaceRoomsLoop(ctx context.Context, logger *zap.Logger, maxSurgeNum int, scheduler entities.Scheduler) error {
+func (ex *SwitchActiveVersionExecutor) startReplaceRoomsLoop(ctx context.Context, logger *zap.Logger, maxSurgeNum int, scheduler entities.Scheduler, op *operation.Operation) error {
 	logger.Info("replacing rooms loop - start")
 	roomsChan := make(chan *game_room.GameRoom)
 	errs, ctx := errgroup.WithContext(ctx)
 
+	var totalRoomsAmount int
+	var err error
+	err = retry.Do(func() error {
```
guilhermocc, line 170: This code will retry just one more time?

go.mod
```diff
@@ -5,6 +5,7 @@ go 1.17
 require (
```
guilhermocc, line 5: @caiordjesus need to run go mod tidy on e2e folder!

internal/core/operations/switch_active_version/switch_active_version_executor.go
caiordjesus: Nope! This method is using default options from the library, which is 10 retries (with increasing delay)

guilhermocc (approved):

arthur29 (approved):
LGTM

PR #474, opened 2022-06-17
Cache node address to optimize kubernetes api usage

guilhermocc:
## What ❓ 
Cache node adresses to optimize kubernetes api usage. Also, use a filter when creating pods informer to ensure maestro only watches for pods that it created.

## Why 🤔 
Currently, for every new pod that maestro starts watching, it makes a **get node** request for fetching the node address, and since this request can delay the event processing in a throttling scenario, we should optimize kube API usage using a cache.

### DIFF
--- a/internal/adapters/runtime/kubernetes/game_room_watcher.go
+++ b/internal/adapters/runtime/kubernetes/game_room_watcher.go
@@ -24,8 +24,11 @@ package kubernetes
 
 import (
 	"context"
+	"fmt"
 	"sync"
 	"time"
+
+	gocache "github.com/patrickmn/go-cache"
 
 	"github.com/topfreegames/maestro/internal/core/entities"
 	"github.com/topfreegames/maestro/internal/core/ports"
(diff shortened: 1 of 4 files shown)


internal/adapters/runtime/kubernetes/game_room_watcher.go
```diff
@@ -224,6 +239,9 @@ func (k *kubernetes) WatchGameRoomInstances(ctx context.Context, scheduler *enti
 		k.clientSet,
 		defaultResyncTime,
 		informers.WithNamespace(scheduler.Name),
+		informers.WithTweakListOptions(func(options *metav1.ListOptions) {
+			options.LabelSelector = fmt.Sprintf("%s=%s", maestroLabelKey, maestroLabelValue)
+		}),
```
guilhermocc, line 244: Without this, maestro will receive events from every pod in the namespace, event if those that are not game rooms.

caiordjesus (approved):

arthur29 (approved):
LGTM

PR #478, opened 2022-06-22
Chore/unify env vars

arthur29:
### What
There are many config files in the maestro with a redundant amount of configs. This PR proposes to unify them and simplify them to avoid this huge number of repetitive configuration entries.

### Why
- Unify configs in `config/config.yaml` file
- Remove old configuration files
- Apply changes on `e2e-tests`


config/config.yaml
```diff
@@ -0,0 +1,44 @@
...
+  healthcheck:
+    enabled: true
+api:
+  port: 8080
+  gracefulShutdownTimeout: 10s
+
+postgres:
+  url: "postgres://maestro:maestro@localhost:5432/maestro?sslmode=disable"
+redis:
+  url: "redis://localhost:6379/0"
```
caiordjesus, line 15: If we were to start the project, I would totally go this way, having only one redis-url and postgres-url. However, since we already have one redis for each adapter, I think we could keep it, since it gives us the possibility to have one redis for each adapter. What do you think about keeping the adapters format we had before but in this unified config file?

docs/test.go
```diff
@@ -0,0 +1,29 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+package main
+
+import "fmt"
+
+func main() {
+	fmt.Println(500 / 35.0)
```
caiordjesus, line 28: I think this file shouldn't be here 😆

Makefile
```diff
@@ -78,24 +78,24 @@ build-linux-x86_64: ## Build the project and generates a binary for x86_64 archi
...
-run/management-api: build ## Runs maestro management-api.
-	@go run main.go start management-api -l development
+.PHONY: run/runtime-watcher
+run/runtime-watcher: build ## Runs maestro runtime-watcher.
+	@MAESTRO_INTERNALAPI_PORT=8061 go run main.go start runtime-watcher -l development
 
 .PHONY: run/rooms-api
 run/rooms-api: build ## Runs maestro rooms-api.
-	@go run main.go start rooms-api -l development
+	@MAESTRO_INTERNALAPI_PORT=8071 MAESTRO_API_PORT=8070 go run main.go start rooms-api -l development
```
caiordjesus, line 89: Do we really need those guys? If yes we'll have to change our deploy, which I think is not the best approach

config/config.yaml
guilhermocc: Totally agree with @caiordjesus, i think that the greatest pain that we currently have is to need to define redis URLs for each maestro component (and this PR already fixes it), and it doesn't make sense to have different maestro components pointing to different redis, but it does make sense to be able to have different redis for different ports, (e.g one redis for game rooms and another just for operations)

guilhermocc: PS: Not a blocking change, consider it just if doesn't take too much effort

PR #481, opened 2022-06-28
Feature/create clean operation history method

guilhermocc:
### What ❓
Create **CleanOperationsHistory** method in operation storage port and adapter, including unitary tests.

### Why 🤔 
This will be one of the steps that our delete scheduler logic flow will take: clean all the operations history, including the history sorted set and each operation hash key.

### DIFF
--- a/internal/adapters/operation/operation_storage_redis.go
+++ b/internal/adapters/operation/operation_storage_redis.go
@@ -271,11 +271,11 @@ func (r *redisOperationStorage) CleanOperationsHistory(ctx context.Context, sche
 		return err
 	}
 
-	operationIDsKeys := make([]string, len(operationsIDs))
-	for i, operationID := range operationsIDs {
-		operationIDsKeys[i] = r.buildSchedulerOperationKey(schedulerName, operationID)
-	}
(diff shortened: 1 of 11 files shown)


internal/adapters/operation/operation_storage_redis.go
```diff
@@ -263,6 +265,34 @@ func (r *redisOperationStorage) ListSchedulerFinishedOperations(ctx context.Cont
...
+	operationsIDs, err := r.getFinishedOperationsFromHistory(ctx, schedulerName, time.Time{}, time.Now())
+	if err != nil {
+		return err
+	}
+
+	var operationIDsKeys []string
+	for _, operationID := range operationsIDs {
+		operationIDsKeys = append(operationIDsKeys, r.buildSchedulerOperationKey(schedulerName, operationID))
+	}
+	if len(operationIDsKeys) > 0 {
```
caiordjesus, line 278: Since you already know the len of "operationsIDs", why don't you test it before trying to populate an array and then testing it? Also, an optimization coming from this would be using make to create an array with the size of len(operationIDs)

caiordjesus (approved):

PR #482, opened 2022-06-30
Feature/implement delete scheduler op executor

guilhermocc:
### What ❓ 
Implement delete scheduler operation executor logic, including unitary tests and a little fix on scheduler storage DeleteScheduler method.

### Why 🤔 
To enable users to delete a scheduler by enqueueing a operation that has the core logic implemented.

### DIFF
--- a/internal/core/operations/deletescheduler/delete_scheduler_executor.go
+++ b/internal/core/operations/deletescheduler/delete_scheduler_executor.go
@@ -24,6 +24,10 @@ package deletescheduler
 
 import (
 	"context"
+	"time"
+
+	"github.com/topfreegames/maestro/internal/core/entities"
+	"github.com/topfreegames/maestro/internal/core/ports"
 
 	"github.com/topfreegames/maestro/internal/core/entities/operation"
 	"github.com/topfreegames/maestro/internal/core/logs"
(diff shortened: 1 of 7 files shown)


internal/core/services/operation_manager/operation_manager.go
```diff
@@ -125,6 +125,7 @@ func (om *OperationManager) PendingOperationsChan(ctx context.Context, scheduler
 		for {
 			operationID, err := om.Flow.NextOperationID(ctx, schedulerName)
 			if err != nil {
+				om.Logger.Error("failed to get next operation ID", zap.Error(err), zap.String(logs.LogFieldSchedulerName, schedulerName))
```
guilhermocc, line 128: Just a little log improvement (non related to the task itself)

internal/core/operations/deletescheduler/delete_scheduler_executor.go
```diff
@@ -61,3 +131,53 @@ func (e *DeleteSchedulerExecutor) Rollback(ctx context.Context, op *operation.Op
...
 	return OperationName
 }
+
+func (e *DeleteSchedulerExecutor) waitForAllInstancesToBeDeleted(ctx context.Context, op *operation.Operation, scheduler *entities.Scheduler, logger *zap.Logger) error {
+	instancesCount, err := e.instanceStorage.GetInstanceCount(ctx, scheduler.Name)
+	if err != nil {
+		logger.Error("failed to get instance count", zap.Error(err))
+		return err
+	}
+
```
caiordjesus, line 141:
```suggestion
	if instancesCount == 0 {
		return nil
	}
```

caiordjesus (approved):

PR #483, opened 2022-07-01
fix logic that makes it harder for maestro to scale up on a high load

caiordjesus:
## What?
Fix logic that makes health controller enqueue remove rooms for rooms that kubernetes is still trying to process

## Why?
Although this logic does not make Maestro fail to scale, it is very counterproductive since, on a high load, Maestro asks kube to delete instances that are still pending

### DIFF
--- a/internal/core/operations/healthcontroller/health_controller_executor.go
+++ b/internal/core/operations/healthcontroller/health_controller_executor.go
@@ -86,15 +86,13 @@ func (ex *SchedulerHealthControllerExecutor) Execute(ctx context.Context, op *op
 		return operations.NewErrUnexpected(err)
 	}
 
-	nonexistentGameRoomIDs := ex.checkNonexistentGameRoomsIDs(gameRoomIDs, instances)
-	if len(nonexistentGameRoomIDs) > 0 {
+	nonexistentGameRoomsIDs, existentGameRoomsInstancesMap := ex.mapExistentAndNonExistentGameRooms(gameRoomIDs, instances)
+	if len(nonexistentGameRoomsIDs) > 0 {
 		logger.Error("found registered rooms that no longer exists")
-		ex.tryEnsureCorrectRoomsOnStorage(ctx, op, logger, nonexistentGameRoomIDs)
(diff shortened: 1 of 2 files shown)


internal/core/operations/healthcontroller/health_controller_executor.go
```diff
@@ -232,6 +237,15 @@ func (ex *SchedulerHealthControllerExecutor) findAvailableAndExpiredRooms(ctx co
 	return availableRoomsIDs, expiredRoomsIDs
 }
 
+func (ex *SchedulerHealthControllerExecutor) isGameRoomInstanceReady(gameRoomId string, instances []*game_room.Instance) bool {
```
guilhermocc, line 240: @caiordjesus this method will be called 1 time for each game room, what about building a **map[gameRoomId]*game_room.Instance** before `findAvailableAndExpiredRooms` function call and pass it as an argument?

internal/core/operations/healthcontroller/health_controller_executor.go
```diff
@@ -208,8 +208,13 @@ func (ex *SchedulerHealthControllerExecutor) ensureDesiredAmountOfInstances(ctx
 	return nil
 }
 
-func (ex *SchedulerHealthControllerExecutor) findAvailableAndExpiredRooms(ctx context.Context, op *operation.Operation, gameRoomsIDs []string) (availableRoomsIDs, expiredRoomsIDs []string) {
+func (ex *SchedulerHealthControllerExecutor) findAvailableAndExpiredRooms(ctx context.Context, op *operation.Operation, gameRoomsIDs []string, instances []*game_room.Instance) (availableRoomsIDs, expiredRoomsIDs []string) {
 	for _, gameRoomID := range gameRoomsIDs {
+		if !ex.isGameRoomInstanceReady(gameRoomID, instances) {
```
guilhermocc, line 213: If an instance is in CrashLoop, its status will be `error`, and it will never be deleted by maestro since we are considering it as available

guilhermocc (changes requested):
If an instance is in CrashLoop, its status will be error, and it will never be deleted by maestro since we are considering it as available before making any other check.

guilhermocc (approved):

PR #484, opened 2022-07-01
Implements EnqueueDeleteSchedulerOperation 

guilhermocc:
### What ❓ 
Implements scheduler manager `EnqueueDeleteSchedulerOperation` method, including unitary tests.

### Why 🤔 
This method will be used for the delete scheduler endpoint to enqueue a delete scheduler operation.


internal/core/services/scheduler_manager/scheduler_manager.go
```diff
@@ -304,6 +324,15 @@ func (s *SchedulerManager) DeleteScheduler(ctx context.Context, schedulerName st
 	return nil
 }
 
+func (s *SchedulerManager) getScheduler(ctx context.Context, schedulerManager string) (*entities.Scheduler, error) {
```
caiordjesus, line 327: Why don't you add the scheduler cache logic to the "GetActiveScheduler" instead of creating a new method? I have the impression we have A LOT of "getScheduler" methods for the scheduler manager

caiordjesus: Also, why did you put the name of the parameter "schedulerManager" instead of "schedulerName"? It is a typo?

guilhermocc: @caiordjesus  By doing this, we would create a lateral dependency between the methods and we already had trouble with it in the past... the right thing would be to change the GetActiveScheduler to use this new method I've created, what do you think?

guilhermocc: [fa898a4](https://github.com/topfreegames/maestro/pull/484/commits/fa898a4d4601a9280bd9b7e4554bdec248b5e3db)

caiordjesus (approved):

PR #486, opened 2022-07-04
Add component metric label to workers manager metrics

guilhermocc:
### What ❓ 
Add `component` label to workers' manager metrics.

### Why 🤔 
Currently, we aren't able to differentiate workers' manager metrics from the maestro component that is running it, and we have three components that use the worker's structure: `runtime watcher`, `operation executor`, and `metrics reporter`.

### DIFF
--- a/internal/core/services/workers_manager/workers_manager.go
+++ b/internal/core/services/workers_manager/workers_manager.go
@@ -26,8 +26,13 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"reflect"
 	"sync"
 	"time"
+
+	"github.com/topfreegames/maestro/internal/core/workers/metricsreporter"
+	"github.com/topfreegames/maestro/internal/core/workers/operation_execution_worker"
+	"github.com/topfreegames/maestro/internal/core/workers/runtime_watcher_worker"
 
 	"github.com/topfreegames/maestro/internal/config"
 	"github.com/topfreegames/maestro/internal/core/entities"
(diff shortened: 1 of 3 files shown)


caiordjesus (approved):

internal/core/services/workers_manager/workers_manager.go
```diff
@@ -196,3 +203,16 @@ func (w *WorkersManager) getDispensableWorkers(schedulers []*entities.Scheduler)
...
+func getRunningComponent(builder workers.WorkerBuilder) string {
+	switch reflect.ValueOf(builder).Pointer() {
+	case reflect.ValueOf(operation_execution_worker.NewOperationExecutionWorker).Pointer():
+		return "operation_execution_worker"
+	case reflect.ValueOf(runtime_watcher_worker.NewRuntimeWatcherWorker).Pointer():
+		return "runtime_watcher_worker"
+	case reflect.ValueOf(metricsreporter.NewMetricsReporterWorker).Pointer():
+		return "metrics_reporter_worker"
+	default:
+		return "unknown"
```
arthur29, line 216: Hey @guilhermocc reflection is not a good practice to use on Golang (including [decreasing the performance](https://betterprogramming.pub/understand-reflect-in-go-24a68fcf1011)), what do you think instead of defining this `getRunningComponent` creating a method like what we do on `operations` called `Name()` or something like that and use this to infer which worker we are getting metrics?

arthur29 (changes requested):

PR #487, opened 2022-07-04
Parametrize redis pool size configuration

guilhermocc:
### What ❓ 
Parametrize redis pool size configuration

### Why 🤔 
To facilitate the parameter configuration and remove the current hardcoded value

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -13,6 +13,8 @@ internalApi:
     enabled: true
 
 adapters:
+  redis:
+    poolSize: 500
   schedulerStorage:
     postgres:
       url: "postgres://maestro:maestro@localhost:5432/maestro?sslmode=disable"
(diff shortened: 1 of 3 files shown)


config/config.yaml
```diff
@@ -13,6 +13,8 @@ internalApi:
     enabled: true
 
 adapters:
+  redis:
+    poolSize: 500
```
caiordjesus, line 17: What makes me think about this is a VERY SPECIFIC way to say the max number of schedulers is 500 (and non-straightforward). Is there a way we could improve what this config var says in practical terms?

guilhermocc: Not sure if I agree with you, technically this is not the max number of schedulers, you could have 500 schedulers with a lower number of connections pool, since theoretically, a scheduler will always be taking and releasing a connection to perform its actions. Besides that, this configuration is EXACTLY the `redis client connections pool size` and I think we should be clear about what it is to prevent misconfiguration, not sure how we could be more precise with the naming here, do you have any suggestions?

caiordjesus (approved):

PR #488, opened 2022-07-04
Feature/e2e tests for delete scheduler

caiordjesus:
## What?
Includes e2e tests for `delete scheduler` feature
## Why?
It's important to have a more complete test so we can guarantee a feature is working as expected from a more holistic view.
## Refactors
The logic for getting from cache was not correct since it was testing errors only. However, not founding on cache is not an error.


internal/core/operations/deletescheduler/delete_scheduler_executor.go
```diff
@@ -176,7 +176,7 @@ func (e *DeleteSchedulerExecutor) waitForAllInstancesToBeDeleted(ctx context.Con
 
 func (e *DeleteSchedulerExecutor) getScheduler(ctx context.Context, schedulerName string) (*entities.Scheduler, error) {
 	scheduler, err := e.schedulerCache.GetScheduler(ctx, schedulerName)
-	if err != nil {
+	if err != nil || scheduler == nil {
```
caiordjesus, line 179: Little fix for when scheduler cache doesn't have value for scheduler yet.

internal/core/services/scheduler_manager/scheduler_manager.go
```diff
@@ -326,7 +326,7 @@ func (s *SchedulerManager) DeleteScheduler(ctx context.Context, schedulerName st
 
 func (s *SchedulerManager) getScheduler(ctx context.Context, schedulerName string) (*entities.Scheduler, error) {
 	scheduler, err := s.schedulerCache.GetScheduler(ctx, schedulerName)
-	if err != nil {
+	if err != nil || scheduler == nil {
```
caiordjesus, line 329: [2] Little fix for when scheduler cache doesn't have value for scheduler yet.

internal/core/operations/switch_active_version/switch_active_version_executor_test.go
```diff
@@ -245,7 +245,7 @@ func TestSwitchActiveVersionOperation_Execute(t *testing.T) {
 			})
 		}
 
-		mocks.operationManager.EXPECT().AppendOperationEventToExecutionHistory(gomock.Any(), gomock.Any(), gomock.Any()).MaxTimes(1)
```
caiordjesus, line 248: This was making this test intermittent

e2e/suites/management/delete_scheduler_test.go
```diff
@@ -0,0 +1,125 @@
...
+				getOperationsResponse := &maestroApiV1.ListOperationsResponse{}
+				err = managementApiClient.Do("GET", fmt.Sprintf("/scheduler/%s/operations", scheduler.Name), getOperationsRequest, getOperationsResponse)
+
+				if len(getOperationsResponse.GetFinishedOperations()) == 0 &&
+					len(getOperationsResponse.GetActiveOperations()) == 0 &&
+					len(getOperationsResponse.GetPendingOperations()) == 0 {
+					return true
+				}
+				return false
+			}, time.Second*60, time.Second*5)
```
guilhermocc, line 109: Hey, @caiordjesus, do you think it is worth trying to create the same scheduler after the deletion to ensure the full cleanup was succeeded?

PR #490, opened 2022-07-08
Improve metrics name convention

arthur29:
### Why
Maestro is using a non-conventional name for its metrics.
### What
- Change metrics finishing in `_counter` to `_total`.
- Change metrics finishing in  `_gauge` to `_unit_total`.

### DIFF
--- a/internal/core/monitoring/metrics.go
+++ b/internal/core/monitoring/metrics.go
@@ -41,6 +41,9 @@ type MetricOpts struct {
 	Name      string
 	Help      string
 	Labels    []string
+
+	// MetricUnit will only be used by Gauge metric to fill which unit the metric is related.
+	MetricUnit string
 }
 
 func CreateCounterMetric(options *MetricOpts) *prometheus.CounterVec {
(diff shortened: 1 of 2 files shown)


caiordjesus (approved):

internal/core/monitoring/metrics.go
```diff
@@ -73,7 +76,7 @@ func CreateGaugeMetric(options *MetricOpts) *prometheus.GaugeVec {
 		prometheus.GaugeOpts{
 			Namespace: options.Namespace,
 			Subsystem: options.Subsystem,
-			Name:      options.Name + "_gauge",
+			Name:      options.Name + "_" + options.MetricUnit + "_total",
```
guilhermocc, line 79: Hey @arthur29, you are following [prometheus naming conventions](https://prometheus.io/docs/practices/naming/) right? if so, it states that "an accumulating count has total as a suffix, in addition to the unit if applicable". Since gauge metric is not cumulative, this piece of code "unit + total" should be in counter metric only.

arthur29: Yeah I think that I understand bad the idea, and the unit should be on metrics that change its number (aka gauge)

guilhermocc (approved):

PR #492, opened 2022-07-14
Removes unusable routes

mflilian:
### What ❓ 
The /remove-rooms proto function and /add-rooms proto function were deleted. As well as the RemoveRooms and AddRooms handler were also deleted from the project. It was ensured that other functions in the flow keep being used by other maestro components.

### Why 🤔 
After the self-healing feature AKA health-controller, the /remove-rooms and /add-rooms route lost their functionality. This could cause confusion for maestro game devs that try to use them and nothing works.


internal/api/handlers/schedulers_handler_test.go
```diff
@@ -627,202 +627,6 @@ func TestCreateScheduler(t *testing.T) {
...
-
-		require.Equal(t, http.StatusConflict, rr.Code)
-		bodyString := rr.Body.String()
-		var body map[string]interface{}
-		err = json.Unmarshal([]byte(bodyString), &body)
-
-		require.NoError(t, err)
-		require.Equal(t, "error creating scheduler scheduler: name already exists", body["message"])
-	})
-}
```
guilhermocc, line 661: Hey @mflilian , I think this test was accidentally removed, right? 😛

mflilian: Yesss!!! hahaah Just putted it back

guilhermocc (approved):

PR #494, opened 2022-07-15
AddRooms req/res and RemoveRooms req/res should be removed

mflilian:
### What ❓ 

- Delete AddRoomsRequest/Response proto operation
- Delete RemoveRoomsRequest/Response proto operation
- Delete AddRoomsRequest/Response struct
- Delete RemoveRoomsRequest/Response struct
- Ensure other functions in the flow keep being used by other maestro components

### Why 🤔 

After the self-healing feature AKA health-controller, the AddRoomsRequest/Response and 
RemoveRoomsRequest/Response struct lost its functionality. This could cause confusion for maestro game devs that try to use it. We removed these routes, now we have to remove their Request/Response struct.


guilhermocc (approved):

arthur29 (commented):
LGTM

PR #497, opened 2022-07-20
Update readme with Wildlife's template

guilhermocc:
## What ❓ 
Update readme file with Wildlife products template. 

image

## Why 🤔 
Since Maestro and other products are managed by Wildlife studios (@topfreegames), we are updating all of our product documentation to follow a standard.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@ manages a fleet of game rooms in isolated namespaces.
 
 Usually a scheduler will define which docker image will be used, its commands, resources that need to be allocated (cpu and memory), and other parameters for fleet management.
 
-Every action that maestro does on kubernetes to manage its resources is encapsulated in what we call an **Operation**. An Operation can be triggered by the user
+Every action that Maestro does on runtime to manage its resources is encapsulated in what we call an **Operation**. An Operation can be triggered by the user
 or by Maestro itself. Maestro uses a queue to control operations flow and disposes of a worker that keeps processing operations.
 
 Maestro provides two APIs for external and internal communication: 
--- a/README.md
+++ b/README.md
@@ -4,133 +4,64 @@
 *WARNING*: The [version v9.x](https://github.com/topfreegames/maestro/tree/v9) of Maestro is under deprecation, complete guide of the new version v10.x can be found [here](https://github.com/topfreegames/maestro/issues/283).
 
 ---
-Maestro: Kubernetes Game Room Scheduler 
-=======================================
+# Maestro 
 
 [![Build Status](https://github.com/topfreegames/maestro/actions/workflows/test.yaml/badge.svg?branch=next)](https://github.com/topfreegames/maestro/actions/workflows/test.yaml)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -4,133 +4,64 @@
...
+# Maestro 
 
 [![Build Status](https://github.com/topfreegames/maestro/actions/workflows/test.yaml/badge.svg?branch=next)](https://github.com/topfreegames/maestro/actions/workflows/test.yaml)
 [![Codecov Status](https://codecov.io/gh/topfreegames/maestro/branch/next/graph/badge.svg?token=KCN2SZDRJF)](https://codecov.io/gh/topfreegames/maestro)
 
-## Docs
-All documentation regarding this version (v10.x, AKA NEXT) can be accessed at https://topfreegames.github.io/maestro/.
+[//]: # (One or two sentences description of the module/service in non technical terms and information about the owner. Text in quotes are examples.)
+
+Game room management service, designed to manage multiple game room fleets in Kubernetes clusters.
```
arthur29, line 14: When I came to the maestro team, we had a discussion about Kubernetes being connected to the maestro. AFAIK, the Kubernetes is currently only an implementation of the runtime, which could be changed to another in the future. What do you think about removing the last part?

README.md
```diff
@@ -4,133 +4,64 @@
...
-
-1. Run `make run/unit-tests` to run all unit tests
-2. Run `make run/integration-tests` to run all integration tests
-3. Run `make run/e2e-tests` to run all E2E tests. NOTE: Currently it is not
-   possible to run it with the development environment set. This command will
-   stop the dev dependencies before running.
-4. Run `make lint` to run all registered linters
+[//]: # (List all the modules or other dependencies that module has, if possible with links to their repositories.)
+
+Maestro does not have any dependencies, but it provides an events forwarding feature that can be used to integrate with external matchmaking services.
```
arthur29, line 40: On the `v9` documentation I wrote that Maestro does not depend on any other system, only the internal databases and the runtime system where it resides

README.md
```diff
@@ -4,133 +4,64 @@
...
+
+## How is the problem solved?
+
+[//]: # (A **more technical description** how the module solved the problem description above. It should not get into too much detail, but provide enough information for a technical leader to understand the implications of using this module.)
+
+With a scheduler, the user can define how a game room can be built and deployed in kubernetes. Each scheduler
```
arthur29, line 46:
```suggestion
With a scheduler, the user can define how a game room can be built and deployed on the runtime. Each scheduler
```

README.md
```diff
@@ -4,133 +4,64 @@
...
+manages a fleet of game rooms in isolated namespaces.
+
+Usually a scheduler will define which docker image will be used, its commands, resources that need to be allocated (cpu and memory), and other parameters for fleet management.
+
+Every action that maestro does on kubernetes to manage its resources is encapsulated in what we call an **Operation**. An Operation can be triggered by the user
```
arthur29, line 51:
```suggestion
Every action that Maestro does on runtime to manage its resources is encapsulated in what we call an **Operation**. An Operation can be triggered by the user
```

arthur29 (changes requested):
The only thing that I suggest to change is talking about Kubernetes as the only current runtime supported but nothing more than this.

arthur29 (approved):
LGTM

PR #502, opened 2022-07-26
Chore/update migration script to consider autoscaling

guilhermocc:
### What ❓ 
Update migration script to convert autoscaling policies.

### Why 🤔 
Our initial version of the migration script didn't had the support for converting v9 - Next autoscaling policies


arthur29 (approved):
LGTM

PR #503, opened 2022-07-26
List operation

mflilian:
### What ❓
Added status and pagination query parameter to the endpoint; 

### Why 🤔
Currently our requests for operations are taking longer than they should. That's why we are implementing pagination on it, starting with the addition of the new queries. 

#### Postscript 📮 
This PR is failing on Integration Tests. These tests are going to be updated on another PR, that so, they should be merged together.


arthur29 (changes requested):

proto/api/v1/operations.proto
```diff
@@ -41,6 +41,12 @@ message ListOperationsRequest {
   // Default value is `created_at desc`
   // NOTE: On http protocol, this operates as a query parameter.
   string order_by = 2;
+  // Parameter to check the operations status.
+  string status = 3;
+  // Parameter for pagination.
+  string page = 4;
+  // Optional parameter for pagination.
+  string per_page = 5;
```
arthur29, line 49: I forgot to comment here that page and per_page should be `optional` parameters since they could not be sent on the request.

proto/api/v1/operations.proto
```diff
@@ -41,6 +41,12 @@ message ListOperationsRequest {
   // Default value is `created_at desc`
   // NOTE: On http protocol, this operates as a query parameter.
   string order_by = 2;
+  // Parameter to check the operations status.
+  optional string status = 3;
```
guilhermocc, line 45: @mflilian we must rename it to `stage` now, right? maybe this PR should be rebased with https://github.com/topfreegames/maestro/pull/507

proto/api/v1/operations.proto
```diff
@@ -52,6 +58,10 @@ message ListOperationsResponse {
   repeated ListOperationItem active_operations = 2;
   // List of the scheduler finished (with error or not) operations.
   repeated ListOperationItem finished_operations = 3;
```
arthur29, line 60: Hey @mflilian what do you think to return only operations without saying which state they are in?

PR #507, opened 2022-07-28
Operations handler - add status validation

mflilian:
### What :question: 
It was implemented on on ListOperations method, a request validation for the status parameter passed on the query. 

### Why 🤔 
Given the need to filter by operation's status, we have to make additional updates on Operations handler.


internal/api/handlers/operations_handler.go
```diff
@@ -64,6 +64,12 @@ func (h *OperationsHandler) ListOperations(ctx context.Context, request *api.Lis
 		return nil, status.Error(codes.InvalidArgument, err.Error())
 	}
 
+	operationStatus := request.Status
+	if err != nil {
+		handlerLogger.Error(fmt.Sprintf("error parsing sorting parameters, orderBy: %+v", request.Status), zap.Error(err))
+		return nil, status.Error(codes.InvalidArgument, err.Error())
+	}
```
arthur29, line 71: Why did this error validation is being done after an assignment to `operation.Status`? What do you think to move it close to which it was assigned?

internal/api/handlers/operations_handler.go
```diff
@@ -103,11 +109,21 @@ func (h *OperationsHandler) ListOperations(ctx context.Context, request *api.Lis
...
+	case "finished":
+		return &api.ListOperationsResponse{FinishedOperations: finishedOperationResponse}, nil
+
+	default:
+		return &api.ListOperationsResponse{
+			PendingOperations:  pendingOperationResponse,
+			ActiveOperations:   activeOperationResponses,
+			FinishedOperations: finishedOperationResponse,
+		}, nil
+	}
```
arthur29, line 126: How are you planning to handle the operations on error status?

proto/api/v1/operations.proto
```diff
@@ -41,6 +41,8 @@ message ListOperationsRequest {
   // Default value is `created_at desc`
   // NOTE: On http protocol, this operates as a query parameter.
   string order_by = 2;
+  // Optional parameter for enabling filter by operation status
+  string status = 3;
```
arthur29, line 45: AFAIK the status is not a required parameter as `page` and `per_page`, what do you think to add the `optional` parameter? This link could help you https://developers.google.com/protocol-buffers/docs/proto#optional

arthur29 (changes requested):

proto/api/v1/operations.proto
```diff
@@ -41,6 +41,8 @@ message ListOperationsRequest {
   // Default value is `created_at desc`
   // NOTE: On http protocol, this operates as a query parameter.
   string order_by = 2;
+  // Optional parameter for enabling filter by operation status
```
guilhermocc, line 44:
```suggestion
  // Optional parameter for enabling filter by operation execution stage, can be one of [pending, active, history]
```

PR #510, opened 2022-08-01
Chore/make operations responsible for generating history

guilhermocc:
### What ❓ 
Update operation executor interface and implementations to use go built-in error interface instead of our custom one.

### Why 🤔 
Currently, the operation executor interface works with a custom interface type that represents an error, and the responsibility of knowing which type of error holds which message to append to the execution history resides in the operation execution worker instead of the implementations of the operations itself.


internal/core/operations/add_rooms/add_rooms_executor.go
```diff
@@ -81,17 +79,14 @@ func (ex *AddRoomsExecutor) Execute(ctx context.Context, op *operation.Operation
 
 	if executionErr := errGroup.Wait(); executionErr != nil {
 		executionLogger.Error("Error creating rooms", zap.Error(executionErr))
-		if errors.Is(executionErr, serviceerrors.ErrGameRoomStatusWaitingTimeout) {
-			return operations.NewErrReadyPingTimeout(executionErr)
-		}
```
guilhermocc, line 86: This line is useless since CreateRoom does not return ping timeout error anymore

internal/core/operations/newschedulerversion/errors.go
```diff
@@ -28,17 +28,74 @@ import (
...
+func (e *ValidationUnexpectedError) Is(other error) bool {
+	if _, ok := other.(*ValidationUnexpectedError); ok {
+		return true
+	}
+	return false
+}
+
+func (e ValidationUnexpectedError) Error() string {
+	return fmt.Sprintf("error validating game room, unexpected error: %s", e.Err.Error())
 }
```
guilhermocc, line 101: @mflilian @arthur29 I had to create these tree types of validation error here, but im not 100% comfortable with this final solution. I thought about creating an executionError struct type to define the common functions for error types (`Error()`,`Unwrap()`, `Is()`) and use `composition` for not having to re-declare these for all new custom error types, but that ended up with me having to use **_reflection_** in the `Is()` function... Any ideas?

internal/core/operations/add_rooms/add_rooms_executor.go
```diff
@@ -81,17 +79,14 @@ func (ex *AddRoomsExecutor) Execute(ctx context.Context, op *operation.Operation
 
 	if executionErr := errGroup.Wait(); executionErr != nil {
 		executionLogger.Error("Error creating rooms", zap.Error(executionErr))
-		if errors.Is(executionErr, serviceerrors.ErrGameRoomStatusWaitingTimeout) {
```
guilhermocc, line 84: This line is useless since CreateRoom method don't return ErrGameRoomStatusWaitingTimeout error anymore

internal/core/operations/remove_rooms/remove_rooms_executor.go
```diff
@@ -71,13 +69,8 @@ func (e *RemoveRoomsExecutor) Execute(ctx context.Context, op *operation.Operati
 		if err != nil {
 			reportDeletionFailedTotal(op.SchedulerName, op.ID)
 			logger.Warn("failed to remove rooms", zap.Error(err))
-			deleteErr := fmt.Errorf("failed to remove room: %w", err)
 
-			if errors.Is(err, serviceerrors.ErrGameRoomStatusWaitingTimeout) {
-				return operations.NewErrTerminatingPingTimeout(deleteErr)
-			}
-
-			return operations.NewErrUnexpected(fmt.Errorf("failed to remove room by ids: %w", err))
```
guilhermocc, line 80: This line is useless since DeleteRoom method doesn't return ErrGameRoomStatusWaitingTimeout error anymore

internal/core/operations/remove_rooms/remove_rooms_executor.go
```diff
@@ -87,13 +80,8 @@ func (e *RemoveRoomsExecutor) Execute(ctx context.Context, op *operation.Operati
 		if err != nil {
 			reportDeletionFailedTotal(op.SchedulerName, op.ID)
 			logger.Warn("failed to remove rooms", zap.Error(err))
-			deleteErr := fmt.Errorf("failed to remove room: %w", err)
-
-			if errors.Is(err, serviceerrors.ErrGameRoomStatusWaitingTimeout) {
-				return operations.NewErrTerminatingPingTimeout(deleteErr)
-			}
 
-			return operations.NewErrUnexpected(fmt.Errorf("failed to remove room by amount: %w", err))
```
guilhermocc, line 96: This line is useless since DeleteRoom method doesn't return ErrGameRoomStatusWaitingTimeout error anymore

internal/core/operations/deletescheduler/delete_scheduler_executor_test.go
```diff
@@ -265,7 +264,7 @@ func TestDeleteSchedulerExecutor_Execute(t *testing.T) {
 
 			err := executor.Execute(ctx, op, definition)
 
-			require.Equal(t, operations.NewErrUnexpected(errors.New("some error on storage")), err)
+			require.Equal(t, errors.New("some error on storage"), err)
```
arthur29, line 267: What do you think to use here `require.EqualError` ([link](https://pkg.go.dev/github.com/stretchr/testify/require#EqualError)) and instead of creating a new error, you will only pass the error message as in the link.

internal/core/operations/newschedulerversion/messages.go
```diff
@@ -0,0 +1,33 @@
...
+package newschedulerversion
+
+const (
+	validationTimeoutMessage = `The GRU could not be validated. Maestro got timeout waiting for the GRU with ID: %s to be ready. You can check if
+		the GRU image is stable on its logs. If you could not spot any issues, contact the Maestro's responsible team for helping.`
+
+	validationPodInErrorMessage = `The GRU could not be validated. The room created for validation with ID %s is entering in error state. You can check if
+		the GRU image is stable on its logs using the provided room id. Last event in the game room: %s.`
+
+	validationUnexpectedErrorMessage = `The GRU could not be validated. Unexpected Error: %s - Contact the Maestro's responsible team for helping.`
```
arthur29, line 32: Hey @guilhermocc as these constants represent an error format, what do you think about renaming them and adding something that represents this thing? For example: `validationTimeoutMessageMold` or something like that

proto/apidocs.swagger.json
```diff
@@ -103,7 +103,23 @@
             "in": "body",
             "required": true,
             "schema": {
-              "$ref": "#/definitions/v1UpdateRoomWithPingRequest"
```
arthur29, line 106: Should swagger changes here?

guilhermocc: No!

arthur29 (approved):

PR #512, opened 2022-08-03
Add retry mechanism to room validation logic

guilhermocc:
### What ❓ 
Update new scheduler version executor to add the room validation retry mechanism. Create new configuration parameter for setting the number of validation retry attempts.

### Why 🤔 
Currently, when validating a new scheduler version, Maestro tries at least once to get a game room up and running and then waits for the room to be validated.

Maestro should have a gain in resiliency by using a retry mechanism in the validation since several intermittencies (network, infrastructure) can occur during the execution.


mflilian (approved):
LGTM

arthur29 (changes requested):

arthur29 (approved):
LGTM

PR #514, opened 2022-08-12
Create CleanExpiredOperations method

mflilian:
### What ❓

On this PR you can see those implementations listed bellow: 
- Get all operation IDs: ZRANGEBYSCORE <operations history key> -inf +inf
- For each operation ID:
- if not EXISTS opID
- ZREM operations:history opID{}

### Why 🤔
Given the need to filter by operation's status, this task proposes to create a method on the operation storage that remove expired operations.

### DIFF
--- a/internal/adapters/operation/operation_storage_redis.go
+++ b/internal/adapters/operation/operation_storage_redis.go
@@ -373,7 +373,7 @@ func (r *redisOperationStorage) CleanExpiredOperations(ctx context.Context, sche
 	}).Result()
 
 	if err != nil {
-		return errors.NewErrUnexpected("failed to list operations for \"%s\"", schedulerName).WithError(err)
+		return errors.NewErrUnexpected("failed to list operations for \"%s\" to perform clean up", schedulerName).WithError(err)
 	}
 
 	if len(operationsIDs) > 0 {
--- a/internal/adapters/operation/operation_storage_redis.go
+++ b/internal/adapters/operation/operation_storage_redis.go
@@ -373,14 +373,14 @@ func (r *redisOperationStorage) CleanExpiredOperations(ctx context.Context, sche
 	}).Result()
 
 	if err != nil {
-		return errors.NewErrUnexpected("failed to list operations for \"%s\" to perform clean up", schedulerName).WithError(err)
+		return errors.NewErrUnexpected("failed to list operations for \"%s\" when trying to clean expired operations", schedulerName).WithError(err)
 	}
 
 	if len(operationsIDs) > 0 {
 
 		pipe := r.client.Pipeline()
 		for _, operationID := range operationsIDs {
-			operationExists := r.client.Exists(ctx, r.buildSchedulerOperationKey(schedulerName, operationID)).Val()
+			operationExists, _ := r.client.Exists(ctx, r.buildSchedulerOperationKey(schedulerName, operationID)).Result()
 			if operationExists == 0 {
 				pipe.ZRem(ctx, r.buildSchedulerHistoryOperationsKey(schedulerName), operationID)
 			}
@@ -392,7 +392,7 @@ func (r *redisOperationStorage) CleanExpiredOperations(ctx context.Context, sche
 		})
 
 		if err != nil {
-			return fmt.Errorf("failed to delete expired operations: %w", err)
+			return fmt.Errorf("failed to clean expired operations: %w", err)
 		}
 	}
 	return nil
--- a/internal/adapters/operation/operation_storage_redis.go
+++ b/internal/adapters/operation/operation_storage_redis.go
@@ -365,28 +365,3 @@ func buildOperationFromMap(opMap map[string]string) (*operation.Operation, error
 		ExecutionHistory: executionHistory,
 	}, nil
 }
-
-func (r *redisOperationStorage) CleanExpiredOperations(ctx context.Context, schedulerName string, from, to time.Time) error {
-	operationsIDs, err := r.client.ZRangeByScore(ctx, r.buildSchedulerHistoryOperationsKey(schedulerName), &redis.ZRangeBy{
-		Min: fmt.Sprint(from.Unix()),
-		Max: fmt.Sprint(to.Unix()),
-	}).Result()
-
-	pipe := r.client.Pipeline()
-
-	if len(operationsIDs) == 0 {
-		for _, operationID := range operationsIDs {
-			pipe.ZRem(ctx, r.buildSchedulerHistoryOperationsKey(schedulerName), operationID)
-		}
-
-		metrics.RunWithMetrics(operationStorageMetricLabel, func() error {
-			_, err = pipe.Exec(ctx)
-			return err
-		})
-
-		if err != nil {
-			return fmt.Errorf("failed to delete expired operations: %w", err)
-		}
-	}
-	return nil
-}
--- a/internal/adapters/operation/operation_storage_redis.go
+++ b/internal/adapters/operation/operation_storage_redis.go
@@ -372,11 +372,15 @@ func (r *redisOperationStorage) CleanExpiredOperations(ctx context.Context, sche
 		Max: "+inf",
 	}).Result()
 
+	if err != nil {
+		return errors.NewErrUnexpected("failed to list operations for \"%s\"", schedulerName).WithError(err)
+	}
+
 	if len(operationsIDs) > 0 {
 
 		pipe := r.client.Pipeline()
 		for _, operationID := range operationsIDs {
-			operationExists := r.client.Exists(ctx, r.buildSchedulerOperationKey(schedulerName)).Val()
+			operationExists := r.client.Exists(ctx, r.buildSchedulerOperationKey(schedulerName, operationID)).Val()
 			if operationExists == 0 {
 				pipe.ZRem(ctx, r.buildSchedulerHistoryOperationsKey(schedulerName), operationID)
 			}
--- a/internal/adapters/operation/operation_storage_redis.go
+++ b/internal/adapters/operation/operation_storage_redis.go
@@ -365,3 +365,28 @@ func buildOperationFromMap(opMap map[string]string) (*operation.Operation, error
 		ExecutionHistory: executionHistory,
 	}, nil
 }
+
+func (r *redisOperationStorage) CleanExpiredOperations(ctx context.Context, schedulerName string, from, to time.Time) error {
+	operationsIDs, err := r.client.ZRangeByScore(ctx, r.buildSchedulerHistoryOperationsKey(schedulerName), &redis.ZRangeBy{
+		Min: fmt.Sprint(from.Unix()),
+		Max: fmt.Sprint(to.Unix()),
+	}).Result()
+
+	pipe := r.client.Pipeline()
+
+	if len(operationsIDs) == 0 {
+		for _, operationID := range operationsIDs {
+			pipe.ZRem(ctx, r.buildSchedulerHistoryOperationsKey(schedulerName), operationID)
+		}
+
+		metrics.RunWithMetrics(operationStorageMetricLabel, func() error {
+			_, err = pipe.Exec(ctx)
+			return err
+		})
+
+		if err != nil {
+			return fmt.Errorf("failed to delete expired operations: %w", err)
+		}
+	}
+	return nil
+}
--- a/internal/adapters/operation/operation_storage_redis.go
+++ b/internal/adapters/operation/operation_storage_redis.go
@@ -366,10 +366,10 @@ func buildOperationFromMap(opMap map[string]string) (*operation.Operation, error
 	}, nil
 }
 
-func (r *redisOperationStorage) CleanExpiredOperations(ctx context.Context, schedulerName string, from, to time.Time) error {
+func (r *redisOperationStorage) CleanExpiredOperations(ctx context.Context, schedulerName string) error {
 	operationsIDs, err := r.client.ZRangeByScore(ctx, r.buildSchedulerHistoryOperationsKey(schedulerName), &redis.ZRangeBy{
-		Min: fmt.Sprint(from.Unix()),
-		Max: fmt.Sprint(to.Unix()),
+		Min: "-inf",
+		Max: "+inf",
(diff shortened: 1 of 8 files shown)


internal/adapters/operation/operation_storage_redis.go
```diff
@@ -365,3 +365,28 @@ func buildOperationFromMap(opMap map[string]string) (*operation.Operation, error
...
+
+func (r *redisOperationStorage) CleanExpiredOperations(ctx context.Context, schedulerName string) error {
+	operationsIDs, err := r.client.ZRangeByScore(ctx, r.buildSchedulerHistoryOperationsKey(schedulerName), &redis.ZRangeBy{
+		Min: "-inf",
+		Max: "+inf",
+	}).Result()
+
+	pipe := r.client.Pipeline()
+
+	if len(operationsIDs) == 0 {
```
guilhermocc, line 377: @mflilian i think you can remove this if check

internal/adapters/operation/operation_storage_redis.go
```diff
@@ -365,3 +365,28 @@ func buildOperationFromMap(opMap map[string]string) (*operation.Operation, error
...
+		for _, operationID := range operationsIDs {
```
guilhermocc, line 378: @mflilian Before removing the operation from sorted set, you must verify if this key exists first

guilhermocc (changes requested):
Before removing the operation from sorted set, you must verify if this key exists first

internal/adapters/operation/operation_storage_redis.go
arthur29, line 377: Hey @mflilian are you executing a for loop range on the `operationsIDs` but this `if` ensures that it will always be empty. I think that you are trying to ensure that it is not empty, right?

internal/adapters/operation/operation_storage_redis_test.go
```diff
@@ -876,3 +876,136 @@ func TestCleanOperationsHistory(t *testing.T) {
 	})
 
 }
+func TestCleanExpiredOperations(t *testing.T) {
```
arthur29, line 879:
```suggestion

func TestCleanExpiredOperations(t *testing.T) {
```

internal/adapters/operation/operation_storage_redis_test.go
```diff
@@ -876,3 +876,136 @@ func TestCleanOperationsHistory(t *testing.T) {
...
+			operationIds, err := client.ZRangeByScore(context.Background(), fmt.Sprintf("operations:%s:lists:history", schedulerName), &redis.ZRangeBy{
+				Min: "-inf",
+				Max: "+inf",
+			}).Result()
+			require.NoError(t, err)
+			assert.NotEmpty(t, operationIds)
+
+			for _, op := range expectedOperations {
+				result, _ := client.HGetAll(context.Background(), fmt.Sprintf("operations:%s:%s", schedulerName, op.ID)).Result()
+				assert.NotEmpty(t, result)
```
arthur29, line 991: May what do you think here to test the exact return values? If some operation hash fields are equal to that previously created.

arthur29 (changes requested):

internal/adapters/operation/operation_storage_redis.go
mflilian: just did it sir, thanks

internal/adapters/operation/operation_storage_redis.go
mflilian: hey, arthur! you're right, just updated this PR fixing it.

internal/adapters/operation/operation_storage_redis_test.go
mflilian: good idea! I'll update the test within their exact return values. thanks

internal/adapters/operation/operation_storage_redis.go
```diff
@@ -365,3 +365,31 @@ func buildOperationFromMap(opMap map[string]string) (*operation.Operation, error
...
 	}, nil
 }
+
+func (r *redisOperationStorage) CleanExpiredOperations(ctx context.Context, schedulerName string) error {
+	operationsIDs, err := r.client.ZRangeByScore(ctx, r.buildSchedulerHistoryOperationsKey(schedulerName), &redis.ZRangeBy{
+		Min: "-inf",
+		Max: "+inf",
+	}).Result()
+
+	if len(operationsIDs) > 0 {
```
guilhermocc, line 375: @mflilian Before proceeding to the if below, we should check and treat if this method returns error. We have defined the `err` var but we aren't checking it.

internal/adapters/operation/operation_storage_redis.go
```diff
@@ -365,3 +365,31 @@ func buildOperationFromMap(opMap map[string]string) (*operation.Operation, error
 		ExecutionHistory: executionHistory,
 	}, nil
 }
+
+func (r *redisOperationStorage) CleanExpiredOperations(ctx context.Context, schedulerName string) error {
+	operationsIDs, err := r.client.ZRangeByScore(ctx, r.buildSchedulerHistoryOperationsKey(schedulerName), &redis.ZRangeBy{
```
guilhermocc, line 370: @mflilian Before proceeding to the if below, we should check and treat if this method returns error. We have defined the `err` var but we aren't checking it.

internal/adapters/operation/operation_storage_redis_test.go
```diff
@@ -876,3 +876,140 @@ func TestCleanOperationsHistory(t *testing.T) {
...
+					definitionContentsRedisKey: op.Input,
+					executionHistoryRedisKey:   executionHistoryJson,
+				}).Err()
+				require.NoError(t, err)
+			}
+
+			operationsReturned, err := storage.ListSchedulerFinishedOperations(context.Background(), schedulerName)
+			assert.NoError(t, err)
+			assert.NotEmptyf(t, operationsReturned, "expected at least one operation")
+			assert.Equal(t, expectedOperations, operationsReturned)
```
guilhermocc, line 978:
```suggestion
```
@mflilian we don't need to list operations here, we are already testing this feature in another test

internal/adapters/operation/operation_storage_redis_test.go
```diff
@@ -876,3 +876,140 @@ func TestCleanOperationsHistory(t *testing.T) {
...
+				err = client.HSet(context.Background(), storage.buildSchedulerOperationKey(op.SchedulerName, op.ID), map[string]interface{}{
+					idRedisKey:                 op.ID,
+					schedulerNameRedisKey:      op.SchedulerName,
+					statusRedisKey:             strconv.Itoa(int(op.Status)),
+					definitionNameRedisKey:     op.DefinitionName,
+					createdAtRedisKey:          op.CreatedAt.Format(time.RFC3339Nano),
+					definitionContentsRedisKey: op.Input,
+					executionHistoryRedisKey:   executionHistoryJson,
+				}).Err()
+				require.NoError(t, err)
```
guilhermocc, line 972: @mflilian, we are adding all operations to the history, and creating all operations hashMaps here, this means that none of these operations will be considered expired since all of them exists, this test is not doing what it says.

In this scenario, we should add all operations IDs to the history, but only create it's hashMaps for, let's say two operations (e.g operations[0] and operations[1]), that way operations[2] and operations[3] would be considered as expired since they don't exist as hashMaps in the redis, but are contained in the history.

internal/adapters/operation/operation_storage_redis_test.go
```diff
@@ -876,3 +876,140 @@ func TestCleanOperationsHistory(t *testing.T) {
...
+			require.NoError(t, err)
+			assert.NotEmpty(t, operationIds)
+
+			for _, op := range expectedOperations {
+				result, _ := client.HGetAll(context.Background(), fmt.Sprintf("operations:%s:%s", schedulerName, op.ID)).Result()
+				require.Equal(t, op.ID, result[idRedisKey])
+				require.Equal(t, op.SchedulerName, result[schedulerNameRedisKey])
+				require.Equal(t, op.DefinitionName, result[definitionNameRedisKey])
+				require.EqualValues(t, op.Input, result[definitionContentsRedisKey])
+			}
```
guilhermocc, line 996: Here we are asserting that all expected operations (operations[0] and operations[1]) exists in the DB as hashMaps, this test will always pass.

The assertion we should do here is that the operations history sorted set does not contain the expired operations IDs anymore, something like this:
```
			operationsIDs, err := client.ZRangeByScore(ctx, r.buildSchedulerHistoryOperationsKey(schedulerName), &redis.ZRangeBy{
				Min: "-inf",
				Max: "+inf",
			}).Result()
			assert.NoError(t, err)
			assert.NotContains(t, operationsIDs, operations[2].ID, operations[3].ID)
			assert.Contains(t, operationsIDs, operations[0].ID, operations[1].ID)
```
Here, we are considering that operations[0] and operations[1] are not expired, so they still exists in the history, but operations[2] and operations[3] were expired, so our clean up method removed them from the history

internal/adapters/operation/operation_storage_redis_test.go
```diff
@@ -876,3 +876,140 @@ func TestCleanOperationsHistory(t *testing.T) {
...
+			operationsReturned, err := storage.ListSchedulerFinishedOperations(context.Background(), schedulerName)
+			assert.NoError(t, err)
+			assert.Empty(t, operationsReturned)
+
+			err = storage.CleanExpiredOperations(context.Background(), schedulerName)
+
+			assert.NoError(t, err)
+		})
+
+	})
```
guilhermocc, line 1014: @mflilian What do you think about adding some error scenarios test here?

internal/adapters/operation/operation_storage_redis.go
```diff
@@ -365,3 +365,31 @@ func buildOperationFromMap(opMap map[string]string) (*operation.Operation, error
...
+		Min: "-inf",
+		Max: "+inf",
+	}).Result()
+
+	if len(operationsIDs) > 0 {
+
+		pipe := r.client.Pipeline()
+		for _, operationID := range operationsIDs {
+			operationExists := r.client.Exists(ctx, r.buildSchedulerHistoryOperationsKey(schedulerName)).Val()
```
guilhermocc, line 379:
```suggestion
			operationExists := r.client.Exists(ctx, r.buildSchedulerOperationKey(schedulerName)).Val()
```
We are using the history key, but we should use `buildSchedulerOperationKey` that is the operation key

internal/adapters/operation/operation_storage_redis.go
arthur29: Hey @mflilian do you already add a test to this case?

mflilian: i updated this case

guilhermocc (approved):
Good Job!

guilhermocc (approved):

internal/adapters/operation/operation_storage_redis_test.go
```diff
@@ -876,3 +876,146 @@ func TestCleanOperationsHistory(t *testing.T) {
...
+			operationsTTlMap := map[Definition]time.Duration{}
+			storage := NewRedisOperationStorage(client, clock, operationsTTlMap)
+
+			operationsReturned, err := storage.ListSchedulerFinishedOperations(context.Background(), schedulerName)
+			assert.NoError(t, err)
+			assert.Empty(t, operationsReturned)
+
+			err = storage.CleanExpiredOperations(context.Background(), schedulerName)
+
+			assert.NoError(t, err)
```
arthur29, line 1001: Hey @mflilian I am missing here an assertion that ensures that there is no operation on the sorted set. I think that doing something like the last test test before:
```golang
                        operationIds, err := client.ZRangeByScore(context.Background(), fmt.Sprintf("operations:%s:lists:history", schedulerName), &redis.ZRangeBy{
					Min: "-inf",
					Max: "+inf",
				}).Result()
				require.NoError(t, err)
				assert.NotEmpty(t, operationIds)
				assert.Contains(t, operationIds, operations[0].ID, operations[1].ID)
				assert.NotContains(t, operationIds, operations[2].ID, operations[3].ID)
			}
```

internal/adapters/operation/operation_storage_redis_test.go
```diff
@@ -876,3 +876,146 @@ func TestCleanOperationsHistory(t *testing.T) {
...
+			t.Run("if client is closed it returns error", func(t *testing.T) {
+				client := test.GetRedisConnection(t, redisAddress)
+				clock := clockmock.NewFakeClock(time.Now())
+				operationsTTlMap := map[Definition]time.Duration{}
+				storage := NewRedisOperationStorage(client, clock, operationsTTlMap)
+				client.Close()
+
+				err := storage.CleanExpiredOperations(context.Background(), schedulerName)
+
+				assert.EqualError(t, err, "failed to list operations for \"test-scheduler\": redis: client is closed")
```
arthur29, line 1015:
```suggestion
assert.ErrorContains(t, err, "failed to list operations for \"test-scheduler\":")
```

Another comment here is that `failed to list operations for` when I see this error on the logs I will not correlate it to a cleanup function, it is more related to list operations for example `GetOperation`, `ListSCHedulerActiveOperations`, etc. What do you think about improving this log message by adding something that connects it with the `CleanExpiredOperations` function?

arthur29 (changes requested):

internal/adapters/operation/operation_storage_redis.go
```diff
@@ -365,3 +365,35 @@ func buildOperationFromMap(opMap map[string]string) (*operation.Operation, error
...
+
+	if err != nil {
+		return errors.NewErrUnexpected("failed to list operations for \"%s\" to perform clean up", schedulerName).WithError(err)
+	}
+
+	if len(operationsIDs) > 0 {
+
+		pipe := r.client.Pipeline()
+		for _, operationID := range operationsIDs {
+			operationExists := r.client.Exists(ctx, r.buildSchedulerOperationKey(schedulerName, operationID)).Val()
```
arthur29, line 383: What do you think to change `Val()` to `Result()`?

arthur29 (approved):
LGTM

PR #517, opened 2022-08-16
Operations storage - update list scheduler finished op method 

mflilian:
### What ❓

On this PR you can see those implementations listed bellow: 
1. It was added to ListSchedulerFinishedOperations method two arguments: page and pageSize
2. New return parameter: int, represents the total number of finishedOperations
3. New logic:
-Before making any query, call CleanExpiredOperations method to ensure the operations history sorted set has no expired operations.
-Get total number of operations;
-Get operations ID with pagination;

### Why 🤔
Given the need to paginate the operations, this task proposes to implement an existent method on the operation storage that remove expired operations.

#### Postscript 📮 
1. This PR should be merged after [this one](https://github.com/topfreegames/maestro/pull/514)
2. The pipeline is breaking because of a missing call on Operations Manager, this will be fixed on the next PR.


internal/adapters/operation/operation_storage_redis.go
```diff
@@ -217,22 +215,29 @@ func (r *redisOperationStorage) ListSchedulerActiveOperations(ctx context.Contex
...
 
-func (r *redisOperationStorage) ListSchedulerFinishedOperations(ctx context.Context, schedulerName string) (operations []*operation.Operation, err error) {
-	currentTime := r.clock.Now()
-	lastDay := r.clock.Now().Add(-24 * time.Hour)
-	operationsIDs, err := r.getFinishedOperationsFromHistory(ctx, schedulerName, lastDay, currentTime)
-	var nonExistentOperations []string
-	var executions = make(map[string]redis.Cmder)
+func (r *redisOperationStorage) ListSchedulerFinishedOperations(ctx context.Context, schedulerName string, page, pageSize int64) (operations []*operation.Operation, total int64, err error) {
+	if err := r.CleanExpiredOperations(ctx, schedulerName); err != nil {
+		return nil, 0, fmt.Errorf("failed to clean scheduler expired operations: %w", err)
```
arthur29, line 220: Hey @mflilian when errors occur normally integer fields are filled with invalid numbers semantically saying. So what do you thing on the errors to put numbers less than zero to symbolize an invalid `page`?

internal/adapters/operation/operation_storage_redis.go
```diff
@@ -335,12 +317,13 @@ func (r *redisOperationStorage) buildSchedulerHistoryOperationsKey(schedulerName
...
-		operationsIDs, err = r.client.ZRangeByScore(ctx, r.buildSchedulerHistoryOperationsKey(schedulerName), &redis.ZRangeBy{
-			Min: fmt.Sprint(from.Unix()),
-			Max: fmt.Sprint(to.Unix()),
+		operationsIDs, err = r.client.ZRevRangeByScore(ctx, r.buildSchedulerHistoryOperationsKey(schedulerName), &redis.ZRangeBy{
+			Min:    "-inf",
+			Max:    "+inf",
+			Offset: page * pageSize,
+			Count:  pageSize,
 		}).Result()
 		return err
```
arthur29, line 328: Since are you naming the error return as err, you don't need to return it here. In this return statement, I think that the naked return is more appropriate.

internal/adapters/operation/operation_storage_redis_test.go
```diff
@@ -858,9 +897,9 @@ func TestCleanOperationsHistory(t *testing.T) {
 		t.Run("if operations history is not empty it deletes all operations and history", func(t *testing.T) {
 			client := test.GetRedisConnection(t, redisAddress)
 			clock := clockmock.NewFakeClock(time.Now())
-			operationsTTLMap := map[Definition]time.Duration{}
-			storage := NewRedisOperationStorage(client, clock, operationsTTLMap)
-			expectedOperations := []*operation.Operation{operations[1], operations[0]}
+			operationsTTlMap := map[Definition]time.Duration{}
+			storage := NewRedisOperationStorage(client, clock, operationsTTlMap)
```
arthur29, line 901:
```suggestion
			operationsTTLMap := map[Definition]time.Duration{}
			storage := NewRedisOperationStorage(client, clock, operationsTTLMap)
```

internal/adapters/operation/operation_storage_redis_test.go
```diff
@@ -912,7 +946,7 @@ func TestCleanOperationsHistory(t *testing.T) {
 			operationsTTLMap := map[Definition]time.Duration{}
 			storage := NewRedisOperationStorage(client, clock, operationsTTLMap)
 
-			operationsReturned, err := storage.ListSchedulerFinishedOperations(context.Background(), schedulerName)
+			operationsReturned, _, err := storage.ListSchedulerFinishedOperations(context.Background(), schedulerName, 0, 10)
```
arthur29, line 949: As the test ensures that there is no `operation` the total return should return `zero`. What do you think to `assert.Equal(t, total, 0)`.

internal/adapters/operation/operation_storage_redis.go
mflilian: Hey @arthur29 , unfortunately this return is inside of the function that runs RunWithMetrics and it does not accept the naked return 😢

internal/adapters/operation/operation_storage_redis.go
mflilian: It makes sense, @arthur29 ! Just applied your suggestions

internal/adapters/operation/operation_storage_redis.go
arthur29: Ok, I understand now, why this error is required.

arthur29 (approved):

PR #518, opened 2022-08-17
Feat/store noop operations in new sorted set

guilhermocc:
### What ❓ 
Update operations storage **_UpdateOperationStatus_** method to store no-action operations in another sorted set instead the _history_.

### Why 🤔 
We are currently including health controller operations that have no action in the operations history. With the high number of no-action health controller operations that are executed periodically, our operations listing ends up with a lot of "noise" with operations that don't really bring the information that the final user needs.

### DIFF
--- a/internal/adapters/operation/operation_storage_redis.go
+++ b/internal/adapters/operation/operation_storage_redis.go
@@ -394,7 +394,7 @@ func (r *redisOperationStorage) getFinishedOperationsFromHistory(ctx context.Con
 
 func (r *redisOperationStorage) updateOperationInSortedSet(ctx context.Context, pipe redis.Pipeliner, schedulerName string, op *operation.Operation, newStatus operation.Status) error {
 	var listKey string
-	operationTookAction, err := operationHasNoAction(op)
+	operationTookAction, err := operationHasAction(op)
 	if err != nil {
 		return errors.NewErrUnexpected("failed to check if operation took action when updating status").WithError(err)
 	}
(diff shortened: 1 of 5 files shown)


internal/adapters/operation/operation_storage_redis.go
```diff
@@ -307,6 +313,38 @@ func (r *redisOperationStorage) UpdateOperationDefinition(ctx context.Context, s
...
+			return err
+		})
+
+		if err != nil {
+			return fmt.Errorf("failed to clean expired operations: %w", err)
+		}
+	}
+	return nil
+}
+
```
guilhermocc, line 347: No changes here, just moved this function

internal/adapters/operation/operation_storage_redis.go
```diff
@@ -350,6 +392,41 @@ func (r *redisOperationStorage) getFinishedOperationsFromHistory(ctx context.Con
...
+	if op.DefinitionName == healthcontroller.OperationName {
+		def := healthcontroller.SchedulerHealthControllerDefinition{}
+		err := def.Unmarshal(op.Input)
+		if err != nil {
+			return false, err
+		}
+		return !*def.TookAction, nil
+	}
+	return false, nil
+}
```
arthur29, line 428: Hey @guilhermocc this is confusing me. Here the function search for `noop`, but the return is to a variable called `operationTookAction`.

arthur29: And during the switch, it is tested as if `operationTookAction`

arthur29 (changes requested):

arthur29 (approved):
LGTM

PR #520, opened 2022-08-18
Feat/clean expired noaction operations

guilhermocc:
### What ❓ 
Update **CleanExpiredOperations** method to also clean expired no-action operations

### Why 🤔 
Since we are now storing no-action operations in another sorted set, we need to ensure that expired operations that reside in this sorted set must also be deleted


mflilian (approved):
good job!

arthur29 (approved):
LGTM

PR #525, opened 2022-08-25
Normalize operations struct names

guilhermocc:
### What ❓ 
Normalize operations struct names

### Why 🤔 
After updating our package names, we must update some struct names to match its package name and remove redundancy


arthur29 (approved):
LGTM

PR #527, opened 2022-08-26
Ensure nonexistent game rooms are also deleted from instance storage

guilhermocc:
### What ❓ 
Update health controller executor to delete nonexistent game room instances during nonexistent game rooms check

### Why 🤔 
Maestro runtime watcher can happen to lose some runtime events, if it miss delete event, the game room instance will remain in our redis storage.


mflilian (approved):
good job sir

arthur29 (approved):
LGTM

PR #531, opened 2022-09-02
Refactor took action

arthur29:
This PR proposes to create inside the `operation.Definition` the function `NoAction` that symbolizes when an operation does not result in an action that should be logged. To accomplish that `operation_storage_redis` was changed to look at the operation method instead of the health controller field and the `make generate` was run, but with a new version since it is running from the `go1.18`.


internal/core/operations/definition.go
```diff
@@ -48,4 +48,6 @@ type Definition interface {
 	// Name returns the definition name. This is used to compare and identify it
 	// among other definitions.
 	Name() string
+	// NoAction return a boolean showing when an operation does not change the runtime.
+	NoAction() bool
```
caiordjesus, line 52: WDYT about calling it HasAction? I think it would be more semantic

internal/core/operations/healthcontroller/definition.go
```diff
@@ -55,11 +55,19 @@ func (def *Definition) Marshal() []byte {
...
 	if err != nil {
 		return fmt.Errorf("error marshalling update scheduler operation definition: %w", err)
 	}
 
 	return nil
 }
+
+func (d *Definition) NoAction() bool {
+	if d.TookAction != nil {
+		return !*d.TookAction
```
caiordjesus, line 69: Considering we renamed this method to HasAction, it would be more simple to read to return directly the TookAction

caiordjesus (commented):
In my opinion refactoring from "TookAction" to "NoAction" was not exactly a good choice, it made me having not a so good time reading since I was always thinking about the positive, but the method brings the negative

caiordjesus (approved):

PR #533, opened 2022-09-05
Implement storagecleanup operation executor

arthur29:
This PR proposes to create the executor to the operation `storagecleanup`. It only calls the `operationstorage.CleanExpiredOperations` that executes all the logic.


caiordjesus (commented):
Misses providers injecting dependencies

caiordjesus (approved):

PR #535, opened 2022-09-06
Smooth the operation storage clean-up along the time

arthur29:
This PR proposes to smooth the execution of the storage clean-up operation over time.

### Why
When putting the version `v10.2.21` in production after some execution was realized that the memory was not going down (see the graph)
image
Investigating deeply what is happening, there were two inputs:

1. The worker pod was restarting.
image
2. Memory usage is creating peaks. Exactly in the `storagecleanup` operation time.
image

To fix this problem this PR proposes to use an [exponential distribution](https://en.wikipedia.org/wiki/Exponential_distribution) to avoid all the operations being executed at the same time and guarantee that in meantime it will be executed almost the same number of times. So to ease the integration with this kind of statistical process into the `storagecleanup` scheduler the module [memoryless](https://pkg.go.dev/github.com/m-lab/go/memoryless) was used here.

In this way is expected that neither all of `storagecleanup` execute at the same time, and the memory load along the time won't create peaks like those in the second image, and finally that the Redis memory usage goes down.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -168,6 +168,7 @@ require (
 	github.com/ldez/tagliatelle v0.2.0 // indirect
 	github.com/leodido/go-urn v1.2.1 // indirect
 	github.com/lib/pq v1.10.1 // indirect
+	github.com/m-lab/go v0.1.53 // indirect
 	github.com/magiconair/properties v1.8.5 // indirect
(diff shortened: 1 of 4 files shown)


caiordjesus (approved):
Excited to see it working

PR #539, opened 2022-09-15
Improve termination grace period

arthur29:
### Why
Currently, `TerminationGracePeriod` is being handled as nanoseconds which is not the best approach since game teams use seconds to handle this variable historically.

### What was done
- Change proto to use `google.protobuf.Duration`
- Change `requestadapters` to cast from `durationpb` instead of `int64`


tobiasbu (approved):
LGTM! 🚀

PR #542, opened 2022-10-06
Fix error on room manager during create room method

arthur29:
### Why
During the new room's creation in the `RoomsManager` there are a lot of errors like this one:
```
"msg":"failed to process event","service_name":"runtime_watcher","error":"failed to update room instance 
<<INSTANCE_NAME>>: failed to update game room status: failed to get game room: room <<INSTANCE_NAME>> 
not found in scheduler <<SCHEDULER_NAME>>"}
```

This error happens when the room starts before the creation process finishes and the proper game room was created in the room storage.
### What
This PR proposes to change the sequence during the creation process to the room being created before the instance creation, so if the room starts to ping the game room will already be created in the storage and it could update the room status by the ping.

### DIFF
--- a/internal/adapters/runtime/kubernetes/game_room.go
+++ b/internal/adapters/runtime/kubernetes/game_room.go
@@ -24,15 +24,18 @@ package kubernetes
 
 import (
 	"context"
+	"fmt"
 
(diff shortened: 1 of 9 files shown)


tobiasbu (approved):
🚀

PR #545, opened 2022-12-07
Create DogStatsD client restart routine (v9)

tobiasbu:
# Context

Recent reports show that the `DogStatsD` client stops to send metrics to Datadog. This occurs especially on `ready_or_occupied` status. This behavior creates a bad experience for monitoring room status.

# Solution

Since Datadog uses a UDP connection, we can't detect when a packet is not sent. So the proposal in this PR is to create a routine that recreates the Datadog client for every defined time duration.

Another thing that was identified is that the logger was not set on `DogStatsD`.

### DIFF
--- a/config/local.yaml
+++ b/config/local.yaml
@@ -43,6 +43,7 @@ reporters:
     host: "0.0.0.0:8125"
     prefix: "maestro."
     region: "region"
+    restartTimeout: 1m
   http:
     timeout: "5s"
     putURL: "http://localhost:8080"
--- a/reporters/dogstatsd.go
+++ b/reporters/dogstatsd.go
@@ -9,19 +9,27 @@ package reporters
 
 import (
 	"fmt"
+	"sync"
+	"time"
 
+	"github.com/DataDog/datadog-go/statsd"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/viper"
 	"github.com/topfreegames/extensions/dogstatsd"
-	constants "github.com/topfreegames/maestro/reporters/constants"
+	"github.com/topfreegames/maestro/reporters/constants"
 	handlers "github.com/topfreegames/maestro/reporters/dogstatsd"
 )
 
 // DogStatsD reports metrics to a dogstatsd.Client
 type DogStatsD struct {
-	client dogstatsd.Client
-	logger *logrus.Logger
-	region string
+	client       dogstatsd.Client
+	logger       *logrus.Logger
+	statsdClient *statsd.Client
+	mutex        sync.Mutex
+	ticker       *time.Ticker
+	region       string
+	host         string
+	prefix       string
 }
 
 func toMapStringString(o map[string]interface{}) map[string]string {
@@ -43,7 +51,9 @@ func (d *DogStatsD) Report(event string, opts map[string]interface{}) error {
 	}
 	opts[constants.TagRegion] = d.region
 	handler := handlerI.(func(dogstatsd.Client, string, map[string]string) error)
+	d.mutex.Lock()
 	err := handler(d.client, event, toMapStringString(opts))
+	d.mutex.Unlock()
 	if err != nil {
 		d.logger.Error(err)
 	}
@@ -63,6 +73,7 @@ func loadDefaultDogStatsDConfigs(c *viper.Viper) {
 	c.SetDefault("reporters.dogstatsd.host", "localhost:8125")
 	c.SetDefault("reporters.dogstatsd.prefix", "test.")
 	c.SetDefault("reporters.dogstatsd.region", "test")
+	c.SetDefault("reporters.dogstatsd.restartTimeout", time.Duration(0))
 }
 
 // NewDogStatsD creates a DogStatsD struct using host and prefix from config
@@ -74,8 +85,24 @@ func NewDogStatsD(config *viper.Viper, logger *logrus.Logger) (*DogStatsD, error
 	if err != nil {
 		return nil, err
 	}
+
+	restartTimeout := config.GetDuration("reporters.dogstatsd.restartTimeout")
 	r := config.GetString("reporters.dogstatsd.region")
-	dogstatsdR := &DogStatsD{client: c, region: r}
+
+	dogstatsdR := &DogStatsD{
+		client:       c,
+		statsdClient: c.Client.(*statsd.Client),
+		logger:       logger,
+		region:       r,
+		host:         host,
+		prefix:       prefix,
+	}
+
+	if restartTimeout.Nanoseconds() > 0 {
+		dogstatsdR.ticker = time.NewTicker(restartTimeout)
+		go dogstatsdR.restartTicker()
+	}
+
 	return dogstatsdR, nil
 }
 
@@ -83,4 +110,22 @@ func NewDogStatsD(config *viper.Viper, logger *logrus.Logger) (*DogStatsD, error
 // dogstatsd.Client -- or a mock client
 func NewDogStatsDFromClient(c dogstatsd.Client, r string) *DogStatsD {
 	return &DogStatsD{client: c, region: r}
+}
+
+func (d *DogStatsD) restartTicker() {
+	for range d.ticker.C {
+		d.mutex.Lock()
+		if err := d.statsdClient.Close(); err != nil {
+			d.logger.Errorf("DogStatsD: failed to close statsd connection during restart: %s", err.Error())
+		}
+		d.mutex.Unlock()
+
+		c, err := dogstatsd.New(d.host, d.prefix)
+		if err == nil {
+			d.statsdClient = c.Client.(*statsd.Client)
+			d.client = c
+		} else {
+			d.logger.Errorf("DogStatsD: failed to recreate dogstatsd client during restart. %s", err.Error())
+		}
+	}
 }
--- a/reporters/dogstatsd.go
+++ b/reporters/dogstatsd.go
@@ -25,7 +25,7 @@ type DogStatsD struct {
 	client       dogstatsd.Client
 	logger       *logrus.Logger
 	statsdClient *statsd.Client
-	mutex        sync.Mutex
+	mutex        sync.RWMutex
 	ticker       *time.Ticker
 	region       string
 	host         string
@@ -45,15 +45,15 @@ func toMapStringString(o map[string]interface{}) map[string]string {
 // Report finds a matching handler to some 'event' metric and delegates
 // further actions to it
 func (d *DogStatsD) Report(event string, opts map[string]interface{}) error {
+	d.mutex.RLock()
+	defer d.mutex.RUnlock()
 	handlerI, prs := handlers.Find(event)
 	if prs == false {
 		return fmt.Errorf("reportHandler for %s doesn't exist", event)
 	}
 	opts[constants.TagRegion] = d.region
 	handler := handlerI.(func(dogstatsd.Client, string, map[string]string) error)
-	d.mutex.Lock()
 	err := handler(d.client, event, toMapStringString(opts))
-	d.mutex.Unlock()
 	if err != nil {
 		d.logger.Error(err)
 	}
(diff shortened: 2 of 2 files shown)


reporters/dogstatsd.go
```diff
@@ -84,3 +102,25 @@ func NewDogStatsD(config *viper.Viper, logger *logrus.Logger) (*DogStatsD, error
...
+				}
+			}
+
+			c, err := dogstatsd.New(d.host, d.prefix)
+			if err == nil {
+				d.statsdClient = c.Client.(*statsd.Client)
+				d.client = c
+			} else {
+				d.logger.Errorf("DogStatsD: failed to recreate dogstatsd client during restart. %s", err.Error())
+				d.client = nil
```
wl-yankabuki, line 121:
```suggestion
```
this can't be null, what about just using the old dogstatsd instance?

tobiasbu: Oh, yeah! I was in doubt about how to deal with the old client instance. I think it's better to maintain old references indeed.

tobiasbu: Thanks for the comment!

tobiasbu: Fixed!

reporters/dogstatsd.go
```diff
@@ -84,3 +102,23 @@ func NewDogStatsD(config *viper.Viper, logger *logrus.Logger) (*DogStatsD, error
...
+			
+			c, err := dogstatsd.New(d.host, d.prefix)
+			if err == nil {
+				d.statsdClient = c.Client.(*statsd.Client)
+				d.client = c
+			} else {
+				d.logger.Errorf("DogStatsD: failed to recreate dogstatsd client during restart. %s", err.Error())
+			}
+
+			d.scheduleRestart(timeout)
```
Henrod, line 121: I'm pretty sure this will build up in the heap and result in out-of-memory. A better solution is to have loop with a time.Ticker executing this. WDYT?

reporters/dogstatsd.go
```diff
@@ -84,3 +102,23 @@ func NewDogStatsD(config *viper.Viper, logger *logrus.Logger) (*DogStatsD, error
...
+func (d *DogStatsD) scheduleRestart(timeout time.Duration) {
+	if timeout.Nanoseconds() > 0 {
+		time.AfterFunc(timeout, func() {
+			if err := d.statsdClient.Close(); err != nil {
+				d.logger.Errorf("DogStatsD: failed to close statsd connection during restart: %s", err.Error())
+			}
+			
+			c, err := dogstatsd.New(d.host, d.prefix)
+			if err == nil {
+				d.statsdClient = c.Client.(*statsd.Client)
```
Henrod, line 115: This isn't thread-safe, right? The goroutine calling this would start failing. Would we let it fail (since losing a few metrics is fine) or control it through a lock? If we'll let it fail, we need to at least document it through comments and turn any error log into a warning log. WDYT?

reporters/dogstatsd.go
tobiasbu: Thanks for pointing out this! I'm not much familiar with the Golang time package. I'll change to `time.Ticker`

reporters/dogstatsd.go
```diff
@@ -84,3 +102,23 @@ func NewDogStatsD(config *viper.Viper, logger *logrus.Logger) (*DogStatsD, error
 func NewDogStatsDFromClient(c dogstatsd.Client, r string) *DogStatsD {
 	return &DogStatsD{client: c, region: r}
 }
+
+func (d *DogStatsD) scheduleRestart(timeout time.Duration) {
+	if timeout.Nanoseconds() > 0 {
+		time.AfterFunc(timeout, func() {
+			if err := d.statsdClient.Close(); err != nil {
```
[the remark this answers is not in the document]
tobiasbu: I actually was in this doubt too. Looking at the statsd client has a mutex that locks when the data is being written in the UDP connection stream. I could reuse the same mutex. Do you think this is a reliable way to make it safer?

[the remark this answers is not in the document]
Henrod: But would you be able to reuse that mutex once the client is closed? Otherwise, you could create a new one.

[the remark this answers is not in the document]
tobiasbu: You're right! Looking at the internal implementation again, I cannot reuse the same mutex. Even if the mutex is not attached to the connection, it is still being used to lock while the data is streamed.

We still need to create another mutex since the connection can be closed while the data is being flushed.

reporters/dogstatsd.go
tobiasbu: Changed to use time.Ticker!

reporters/dogstatsd.go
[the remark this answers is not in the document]
tobiasbu: Created sync.Mutex for DogStatsD instance

reporters/dogstatsd.go
```diff
@@ -84,3 +109,21 @@ func NewDogStatsD(config *viper.Viper, logger *logrus.Logger) (*DogStatsD, error
 func NewDogStatsDFromClient(c dogstatsd.Client, r string) *DogStatsD {
 	return &DogStatsD{client: c, region: r}
 }
+
+func (d *DogStatsD) restartTicker() {
+	for range d.ticker.C {
+		d.mutex.Lock()
+		if err := d.statsdClient.Close(); err != nil {
```
Henrod, line 116: 1. Why does this statsdClient exist if it isn't being used besides this close here?
2. Why is the lock/unlock here? We need to lock/unlock the client being used to report, this one exists just to be closed?

reporters/dogstatsd.go
```diff
@@ -74,8 +83,24 @@ func NewDogStatsD(config *viper.Viper, logger *logrus.Logger) (*DogStatsD, error
...
 		return nil, err
 	}
+
+	restartTimeout := config.GetDuration("reporters.dogstatsd.restartTimeout")
 	r := config.GetString("reporters.dogstatsd.region")
-	dogstatsdR := &DogStatsD{client: c, region: r}
+
+	dogstatsdR := &DogStatsD{
+		client:       c,
+		statsdClient: c.Client.(*statsd.Client),
```
Henrod, line 92: What's the difference between client and statsdClient?

config/local.yaml
```diff
@@ -43,6 +43,7 @@ reporters:
     host: "0.0.0.0:8125"
     prefix: "maestro."
     region: "region"
+    restartTimeout: 30m
```
Henrod, line 46: In my opinion, this should be 1 minute, maybe 5 minutes at most. Otherwise, we'll be 30 minutes without metrics and the problem will be perceptive by the user anyway.

Henrod (changes requested):

reporters/dogstatsd.go
```diff
@@ -9,7 +9,10 @@ package reporters
 
 import (
 	"fmt"
+	"sync"
+	"time"
 
+	"github.com/DataDog/datadog-go/statsd"
```
Henrod, line 15: We already have the dogstatsd down below. Is it necessary to have both?

reporters/dogstatsd.go
tobiasbu: The `c.Client` is an interface from [this library](https://github.com/topfreegames/extensions). It abstracts the original implementation of Datadog's client.

While `statsdClient` is the casted version of `c.Client` to `statsd.Client`. I needed cast to be able to access the `Close` method.

reporters/dogstatsd.go
tobiasbu: 1. Because the current implementation of maestro wraps the interface from Datadog's client implemented in the TFGCo Go Extensions lib, which in turn also hides the original implementation of `statsd.Client`. So, to be able to access the `Close` function on the restartTicker, I need to cast in the constructor of maestro-Datadog's client ([here](https://github.com/topfreegames/maestro/blob/v9/reporters/dogstatsd.go#L73)). 
    You can see the wrapper of Datadog's client from TFGCo Go Extensions [here](https://github.com/topfreegames/extensions/blob/93d8d3815c98908f0fbb0e3050fdadd135bf9ede/dogstatsd/dogstatsd.go#L17).
2. Ops! Actually, we should lock the report functions too! I'll fix it soon!

reporters/dogstatsd.go
tobiasbu: I need to import this package to be able to access the `statsd.Client` which has the `Close` function, as I mentioned in [this comment](https://github.com/topfreegames/maestro/pull/545#discussion_r1044629580).

config/local.yaml
tobiasbu: Changed to 1 minute!

reporters/dogstatsd.go
tobiasbu: Added locking in report function.

reporters/dogstatsd.go
```diff
@@ -84,3 +111,21 @@ func NewDogStatsD(config *viper.Viper, logger *logrus.Logger) (*DogStatsD, error
 func NewDogStatsDFromClient(c dogstatsd.Client, r string) *DogStatsD {
 	return &DogStatsD{client: c, region: r}
 }
+
+func (d *DogStatsD) restartTicker() {
+	for range d.ticker.C {
+		d.mutex.Lock()
+		if err := d.statsdClient.Close(); err != nil {
+			d.logger.Errorf("DogStatsD: failed to close statsd connection during restart: %s", err.Error())
```
Henrod, line 119: If you failed to close, shouldn't you return here? Otherwise, you'll keep unclosed resources, potentially leaking memory.

reporters/dogstatsd.go
Henrod: I think this mutex should include this whole function, right? Even the part that assigns the new client. Otherwise, the report function could use a closed reporter, leading to an error. 

What I suggest is:
```suggestion
                defer d.mutex.Unlock()
		if err := d.statsdClient.Close(); err != nil {
```

And remove the Unlock below. So, it'll Unlock at every return of the function.

reporters/dogstatsd.go
```diff
@@ -9,19 +9,27 @@ package reporters
...
 
 // DogStatsD reports metrics to a dogstatsd.Client
 type DogStatsD struct {
-	client dogstatsd.Client
-	logger *logrus.Logger
-	region string
+	client       dogstatsd.Client
+	logger       *logrus.Logger
+	statsdClient *statsd.Client
+	mutex        sync.Mutex
```
Henrod, line 28: Can you use a RWMutex? That way we allow multiple reads but only one write.

Henrod (changes requested):

reporters/dogstatsd.go
tobiasbu: You're totally right!

Question: since this is a go routine, we're using a ticker to loop over an interval, should I use `continue` instead?

reporters/dogstatsd.go
tobiasbu: Deferring, in this case, will unlock the Mutex only when the function call is finished. Since we're using a for loop over a ticker, does the `defer` close after the iteration over a tick?

reporters/dogstatsd.go
tobiasbu: Changed to RWMutex

reporters/dogstatsd.go
Henrod: Yes!

reporters/dogstatsd.go
Henrod: That works. Don't worry about changing it, but another idea was to separate in another method. So you could do:
```
for {
  mutex.Lock()
  submethod()
  mutex.Unlock()  
}
```
But that is more useful for methods with many returns since the developer doesn't need to add the unlock in all of them, with the risk of forgetting one.

reporters/dogstatsd.go
```diff
@@ -37,6 +45,8 @@ func toMapStringString(o map[string]interface{}) map[string]string {
 // Report finds a matching handler to some 'event' metric and delegates
 // further actions to it
 func (d *DogStatsD) Report(event string, opts map[string]interface{}) error {
+	d.mutex.Lock()
+	defer d.mutex.Unlock()
```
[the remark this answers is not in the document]
tobiasbu: I see... Interesting! I'll change it!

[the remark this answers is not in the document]
tobiasbu: Changed it to RLock and RUnlock

reporters/dogstatsd.go
```diff
@@ -84,3 +111,23 @@ func NewDogStatsD(config *viper.Viper, logger *logrus.Logger) (*DogStatsD, error
 func NewDogStatsDFromClient(c dogstatsd.Client, r string) *DogStatsD {
 	return &DogStatsD{client: c, region: r}
 }
+
+func (d *DogStatsD) restartTicker() {
+	for range d.ticker.C {
+		d.mutex.Lock()
+		err := d.statsdClient.Close()
+		d.mutex.Unlock()
```
henrod, line 119: Oh, did you change this back? It was correct before. You should have an Unlock before every end of the loop. Be a `continue`, be the end of the bracket. 

Maybe my comment was misleading? What I meant was this:
1. Have a submethod, such as (with a few changes too):
```
func (d *DogStatsD) restartTicker() error {
		err := d.statsdClient.Close()
		if err != nil {
			return fmt.Errorf("failed to close statsd connection: %w", err)
		}

		c, err := dogstatsd.New(d.host, d.prefix)
		if err != nil {
			return fmt.Errorf("failed to recreate dogstatsd client:%w", err)
		}

		d.statsdClient = c.Client.(*statsd.Client)
		d.client = c
}
```

2. Call the submethod in the loop method:

```
func (d *DogStatsD) restartTickerInLoop() {
		for range d.ticker.C {
				d.mutex.Lock()
				if err := d.restartTicker(); err != nil {
						d.logger.Errorf("DogStatsD: failed to close statsd connection during restart: %s", err.Error())
				}
				d.mutex.Unlock()
		}
}
```

But, please, I was ok with the previous solution. You can also just revert it back. As long as the Lock/Unlock encompass the whole close+assignment.

tobiasbu: Oh! 🤦🤦🤦 Sorry, I misunderstood.

I applied your suggestion. I agree with you. It's more clear and can avoid misinterpretation in the future.

Sorry again for taking your time on this review 🙇‍♂️

henrod: Hey, Tobias, don't worry about it 🙂

henrod:
LGTM!

henrod (approved):

PR #546, opened 2022-12-19
Add logs to reporter and watcher

tobiasbu:
## Context

Add logging functions to reporters and to the watcher to allow us to identify if the `Report` functions are dispatching errors correctly.

Related to #545.

### DIFF
--- a/reporters/dogstatsd/handlers.go
+++ b/reporters/dogstatsd/handlers.go
@@ -71,7 +71,10 @@ func GruStatusHandler(c dogstatsd.Client, event string,
 	if err != nil {
 		return err
 	}
-	return c.Gauge(fmt.Sprintf("gru.%s", opts["status"]), gauge, tags, 1)
+	if err := c.Gauge(fmt.Sprintf("gru.%s", opts["status"]), gauge, tags, 1); err != nil {
+		return fmt.Errorf("dogstatsd failed to GRU Status gauge %s: %w", opts["status"], err)
+	}
+	return nil
 }
 
 // GruTimingHandler calls dogstatsd.Client.Timing with tags formatted as key:value
@@ -134,5 +137,8 @@ func GaugeHandler(
 	if err != nil {
 		return err
 	}
-	return c.Gauge(event, gauge, tags, 1)
+	if err := c.Gauge(event, gauge, tags, 1); err != nil {
+		return fmt.Errorf("dogstatsd failed to gauge %s: %w", opts["status"], err)
+	}
+	return nil
 }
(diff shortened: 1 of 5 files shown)


manuellysuzik (approved):

reporters/dogstatsd/handlers.go
```diff
@@ -135,6 +134,5 @@ func GaugeHandler(
 	if err != nil {
 		return err
 	}
-	c.Gauge(event, gauge, tags, 1)
-	return nil
+	return c.Gauge(event, gauge, tags, 1)
```
henrod, line 137: Can you wrap this and other errors with `fmt.Errorf`?

watcher/watcher.go
```diff
@@ -286,7 +286,9 @@ func (w *Watcher) reportRoomsStatusesRoutine() {
 		case <-podStateCountTicker.C:
 			w.PodStatesCount()
 		case <-roomStatusTicker.C:
-			_ = w.ReportRoomsStatuses()
+			if err := w.ReportRoomsStatuses(); err != nil {
```
henrod, line 289: Do you think adding a "started reporting" log will help? Also, one after this for loop showing that it won't report anymore.

watcher/watcher.go
```diff
@@ -461,12 +463,19 @@ func (w *Watcher) ReportRoomsStatuses() error {
...
 	for _, r := range roomDataSlice {
-		_ = reporters.Report(reportersConstants.EventGruStatus, map[string]interface{}{
+		err := reporters.Report(reportersConstants.EventGruStatus, map[string]interface{}{
 			reportersConstants.TagGame:      w.GameName,
 			reportersConstants.TagScheduler: w.SchedulerName,
 			"status":                        r.Status,
 			"gauge":                         r.Gauge,
 		})
+		if err != nil {
+			w.Logger.Error(err)
```
henrod, line 473: This is the most important error to show. We need to find it by a tag, so can you change it to this?
```suggestion
			w.Logger.WithError(err).Error("failed to report room status")
```

henrod (changes requested):

watcher/watcher.go
tobiasbu: Yeah, it could help. I added only in the forrange of roomDataSlice

watcher/watcher.go
tobiasbu: Oh yeah! I just have fixed it some minutes ago 😅

reporters/dogstatsd/handlers.go
tobiasbu: Wrapping gauge errors with fmt.Errorf

watcher/watcher.go
tobiasbu: Add suggested logs!

henrod (approved):
Let's 🚢 it!

PR #548, opened 2022-12-22
Ignore reporters errors (v9)

tobiasbu:
## Context

The latest version (v9.15.1) has issues related to the ping mechanism.

## Solution

We need to ignore the reporter errors explicitly. The reason is if we return the reporter's errors, it could bring issues to the ping mechanism, and we would not be able to find any room.

### DIFF
--- a/reporters/reporter.go
+++ b/reporters/reporter.go
@@ -8,7 +8,6 @@
 package reporters
 
 import (
-	"fmt"
 	"strings"
 	"sync"
 
@@ -52,15 +51,11 @@ func copyOpts(src map[string]interface{}) map[string]interface{} {
 
 // Report is Reporters' implementation of the Reporter interface
 func (r *Reporters) Report(event string, opts map[string]interface{}) error {
-	var aggregatedErrors []error
 	for _, reporter := range r.reporters {
-		if err := reporter.Report(event, copyOpts(opts)); err != nil {
-			aggregatedErrors = append(aggregatedErrors, err)
-		}
-	}
-
-	if len(aggregatedErrors) > 0 {
-		return fmt.Errorf("failed to report '%s' event: %v", event, aggregatedErrors)
+		// We ignore the reporter errors explicitly here for the following reason:
+		// if we return these errors, it could bring issues in the ping mechanism,
+		// and we would not be able to find any room.
+		_ = reporter.Report(event, copyOpts(opts))
 	}
 	return nil
 }
--- a/reporters/reporter.go
+++ b/reporters/reporter.go
@@ -8,8 +8,11 @@
 package reporters
 
 import (
+	"errors"
 	"strings"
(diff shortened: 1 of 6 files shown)


paulakfleck:
LGTM!

reporters/reporter.go
```diff
@@ -52,15 +51,11 @@ func copyOpts(src map[string]interface{}) map[string]interface{} {
 
 // Report is Reporters' implementation of the Reporter interface
 func (r *Reporters) Report(event string, opts map[string]interface{}) error {
```
tobiasbu, line 53: To clarify, this function is used in many places in the project. Especially [here](https://github.com/topfreegames/maestro/blob/v9/models/room.go#L333).

reporters/reporter.go
```diff
@@ -52,15 +51,11 @@ func copyOpts(src map[string]interface{}) map[string]interface{} {
...
-			aggregatedErrors = append(aggregatedErrors, err)
-		}
-	}
-
-	if len(aggregatedErrors) > 0 {
-		return fmt.Errorf("failed to report '%s' event: %v", event, aggregatedErrors)
+		// We ignore the reporter errors explicitly here for the following reason:
+		// if we return these errors, it could bring issues in the ping mechanism,
+		// and we would not be able to find any room.
+		_ = reporter.Report(event, copyOpts(opts))
```
henrod, line 58: In order to help us debug, can you log the error (still, without returning it) instead of ignoring it?

watcher/watcher.go
```diff
@@ -286,11 +286,9 @@ func (w *Watcher) reportRoomsStatusesRoutine() {
 		case <-podStateCountTicker.C:
 			w.PodStatesCount()
 		case <-roomStatusTicker.C:
-			w.Logger.Info("Start to report rooms status")
```
henrod, line 289: Why remove these logs?

henrod (dismissed):

reporters/reporter.go
tobiasbu: Sure!

watcher/watcher.go
tobiasbu: These ones weren't very helpful. But I can leave the logs there too

reporters/reporter.go
```diff
@@ -77,6 +84,8 @@ func Report(event string, opts map[string]interface{}) error {
 
 // MakeReporters creates Reporters' singleton from config/{}.yaml
 func MakeReporters(config *viper.Viper, logger logrus.FieldLogger) {
+	GetInstance().setLogger(logger)
```
tobiasbu, line 87: Since the `Reporters` is a Singleton, I needed to add this function to be able to log the `Report` errors.

lyracampos (approved):
LGTM!

PR #549, opened 2022-12-30
Update events forwarder cache to run cleanup manually when cache expires

wl-yankabuki:
Relates to the following issue: https://github.com/patrickmn/go-cache/issues/141

**Description**
The caching dependency is not running the cleanup properly due to how the cache is implemented in the events forwarder client. 

The cache implementation only runs calls `OnEvict` each `cleanupInterval`, for each cache item that is expired. Whenever we try to use a cached connection that is expired, we create a new connection, which in turn uses a new expiration date.

If we try to access a cached connection that is expired but not properly evicted, we will end up replacing it before calling the `close()` procedure and we will end up with opened grpc connections. 

***

The pipeline change is necessary due to Github updating its [ubuntu-latest image to Ubuntu 22.04 ](https://github.com/actions/runner-images) and [Ubuntu 22.04 not being supported by K3s](https://pkg.go.dev/github.com/orlangure/gnomock/preset/k3s)

### DIFF
--- a/internal/adapters/events/forwarder_client.go
+++ b/internal/adapters/events/forwarder_client.go
@@ -54,7 +54,7 @@ type ForwarderClient struct {
 
 // NewForwarderClient instantiate a new grpc forwarder client.
 func NewForwarderClient() *ForwarderClient {
-	cache := cache.New(24*time.Hour, 25*time.Hour)
+	cache := cache.New(24*time.Hour, 0)
 	cache.OnEvicted(func(_key string, clientFromCache interface{}) {
 		ForwarderClient := clientFromCache.(*grpc.ClientConn)
 		ForwarderClient.Close()
@@ -150,6 +150,7 @@ func (f *ForwarderClient) getGrpcClient(address Address) (pb.GRPCForwarderClient
 		if err != nil {
 			return nil, err
 		}
+		f.c.DeleteExpired()
 		f.c.Set(string(address), client, cache.DefaultExpiration)
 		return pb.NewGRPCForwarderClient(client), nil
 	}
--- a/internal/adapters/events/forwarder_client_test.go
+++ b/internal/adapters/events/forwarder_client_test.go
@@ -33,7 +33,6 @@ import (
 
 	"testing"
 
-	"github.com/patrickmn/go-cache"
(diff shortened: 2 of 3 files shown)


internal/adapters/events/forwarder_client.go
```diff
@@ -54,7 +54,7 @@ type ForwarderClient struct {
 
 // NewForwarderClient instantiate a new grpc forwarder client.
 func NewForwarderClient() *ForwarderClient {
-	cache := cache.New(24*time.Hour, 25*time.Hour)
+	cache := cache.New(24*time.Hour, 0)
```
tobiasbu, line 57: Why `0`?

internal/adapters/events/forwarder_client_test.go
```diff
@@ -204,9 +190,7 @@ func TestForwarderClient_SendRoomStatus(t *testing.T) {
 				)
 				require.NoError(t, err)
 			}
-			f := &ForwarderClient{
-				c: tt.fields.c,
-			}
+			f := NewForwarderClient()
```
tobiasbu, line 193: Nice!

tobiasbu (commented):
LGTM!

internal/adapters/events/forwarder_client.go
wl-yankabuki: 0 disables the janitor go routine that runs the cache evictions.
Since we are doing this manually, we don't need to use this

manuellysuzik:
LGTM!

manuellysuzik (approved):

henrod:
Not sure I got it. Aren't the clients being closed in the OnEvict method? How are the connections being kept open?

wl-yankabuki:
> Not sure I got it. Aren't the clients being closed in the OnEvict method? How are the connections being kept open?

The OnEvict is not guaranteed to run for all cache items. The execution of this function only happens in the cleanup ticket that runs in a goroutine. If you replace the cache item after its expiration, but before the goroutine execution, the onEvict is never called, since the map item is replaced with the new one with a different expiration time.

image

henrod (approved):
LGTM

PR #550, opened 2023-02-06
Forward events when event type is status

henrod:
# Objectives

1. Forward event even when the event type is status, that is when the room changes its status.
2. Support status with capitalized letters


reinaldooli (approved):
LGTM!

PR #553, opened 2023-02-09
Add game to forward player event

wl-yankabuki:
### Fix

During the update from v9 to next, we stopped filling the game field to be sent by the events forwarder implementation.
In v9: https://github.com/topfreegames/maestro/blob/b20b9b51722b5333b6a53ae6033d6bd05ae9397c/eventforwarder/forward.go#L193

This change will instead of passing the game in the metadata, pass it using the room.Game field that is already available in the proto for player events.


tobiasbu (approved):
LGTM!

PR #557, opened 2023-02-14
Add error messages to rooms operation history

wl-yankabuki:
### Why
Nowadays the operation history does not persist why an operation has failed for most use cases. This change intends to add the error to the operation history for room operations.

### DIFF
--- a/internal/core/operations/rooms/add/executor.go
+++ b/internal/core/operations/rooms/add/executor.go
@@ -40,32 +40,35 @@ import (
 )
 
 type Executor struct {
-	roomManager ports.RoomManager
-	storage     ports.SchedulerStorage
-	logger      *zap.Logger
+	roomManager      ports.RoomManager
+	storage          ports.SchedulerStorage
+	operationManager ports.OperationManager
 }
 
 var _ operations.Executor = (*Executor)(nil)
 
-func NewExecutor(roomManager ports.RoomManager, storage ports.SchedulerStorage) *Executor {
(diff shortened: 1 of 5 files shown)


internal/core/operations/rooms/add/executor.go
```diff
@@ -40,32 +40,35 @@ import (
...
 		zap.String(logs.LogFieldOperationID, op.ID),
 	)
 	amount := definition.(*Definition).Amount
 	scheduler, err := ex.storage.GetScheduler(ctx, op.SchedulerName)
 	if err != nil {
-		executionLogger.Error(fmt.Sprintf("Could not find scheduler with name %s, can not create rooms", op.SchedulerName), zap.Error(err))
-		return err
+		executionLogger.Error("error fetching scheduler from storage, can not create rooms", zap.Error(err))
+		getSchedulerStorageErr := fmt.Errorf("error fetching scheduler from storage: %w", err)
+		ex.operationManager.AppendOperationEventToExecutionHistory(ctx, op, getSchedulerStorageErr.Error())
```
tobiasbu, line 70: I was thinking about this again and you can avoid repeating it by doing:

```suggestion
		getSchedulerStorageErr := fmt.Errorf("error fetching scheduler from storage: %w", err)
		executionLogger.Error(getSchedulerStorageErr.Error())
		
		ex.operationManager.AppendOperationEventToExecutionHistory(ctx, op, getSchedulerStorageErr.Error())
```

wl-yankabuki: that will change the logging message fields, I think the easier approach is to create a constant for this message, since We already merged the other changes I'll update them in a new PR

PR #563, opened 2023-02-16
Update scheduler state to in-sync after creation

wl-yankabuki:
### Why

The scheduler state is not being updated after its creation, this change updates it to in-sync when the create scheduler operation finished. Also, when a new version operation is executed, it will set the scheduler to creating until the switch version operation finishes.


lyracampos (approved):
LGTM!

PR #573, opened 2023-03-01
Add safe to evict annotation to kubernetes pods

reinaldooli:
### What does this PR do?

Add `safe-to-evict` annotation to every pod created by maestro.

### Context

It'll allow kubernetes to scale down cluster nodes, resulting in infra cost savings.


wl-yankabuki (approved):
LGTM

tobiasbu (approved):
LGTM!

PR #576, opened 2023-03-13
Use ready rooms to calculate the desired number of rooms

reinaldooli:
### What does this MR do?

Uses `Ready` game rooms to calculate the desired number of rooms.


tobiasbu (approved):
LGTM!

PR #577, opened 2023-03-17
Add instrumentation to postgres and redis

reinaldooli:
### What does this PR do?

Add instrumentation, with OpenTelemetry and Jaeger to the storage services (Postgres and Redis).<br />
I need to upgrade the `go-pg` library in order to add instrumentation to it.


cmd/commom/commom.go
```diff
@@ -79,3 +84,66 @@ func MatchPath(path, pattern string) bool {
...
+	provider := trace.NewTracerProvider(trace.WithResource(res))
+	otel.SetTracerProvider(provider)
+
+	//exp, err := stdouttrace.New(stdouttrace.WithPrettyPrint())
+	//if err != nil {
+	//	return nil, err
+	//}
+	//
+	//bsp := trace.NewBatchSpanProcessor(exp)
+	//provider.RegisterSpanProcessor(bsp)
```
tobiasbu, line 132: 👀

internal/service/adapters.go
```diff
@@ -222,7 +225,16 @@ func createRedisClient(c config.Config, url string) (*redis.Client, error) {
...
 	opts.PoolSize = c.GetInt(redisPoolSizePath)
-	return redis.NewClient(opts), nil
+
+	hostPort := strings.Split(opts.Addr, ":")
+
+	client := redis.NewClient(opts)
+	client.AddHook(redisotel.NewTracingHook(redisotel.WithAttributes(
+		semconv.NetPeerNameKey.String(hostPort[0]),
+		semconv.NetPeerPortKey.String(hostPort[1])),
+	))
```
tobiasbu, line 235: Should it not check if `tracing` is enabled to add this hook?

cmd/commom/commom.go
```diff
@@ -79,3 +84,66 @@ func MatchPath(path, pattern string) bool {
...
 }
+
+func ConfigureTracer(ctx context.Context, serviceName string) (func() error, error) {
+	tracerUrl := os.Getenv("OTEL_EXPORTER_JAEGER_ENDPOINT")
+	switch {
+	case tracerUrl != "":
+		return configureJaeger(ctx, serviceName)
+	default:
+		return configureStdout(ctx, serviceName)
+	}
```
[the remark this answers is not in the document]
tobiasbu: Should it not check if `tracing` is enabled here before configuring both Jaeger and stdout?

cmd/commom/commom.go
```diff
@@ -79,3 +85,66 @@ func MatchPath(path, pattern string) bool {
 	}
 	return match
 }
+
+func ConfigureTracer(ctx context.Context, serviceName string) (func() error, error) {
+	tracerUrl := os.Getenv("OTEL_EXPORTER_JAEGER_ENDPOINT")
```
wl-yankabuki, line 90:
```suggestion
	tracerUrl := configs.GetBool("api.tracing.url")
```

cmd/commom/commom.go
```diff
@@ -79,3 +85,66 @@ func MatchPath(path, pattern string) bool {
...
+	switch {
+	case tracerUrl != "":
+		return configureJaeger(ctx, serviceName)
+	default:
+		return configureStdout(ctx, serviceName)
+	}
```
wl-yankabuki, line 96:
```suggestion
	tracerType := config.GetString("api.tracing.type")
	switch {
	case tracerType == "jaeger":
		return configureJaeger(ctx, serviceName)
	default:
		return configureStdout(ctx, serviceName)
	}
```

I think this leaves things clearer for the user than an environment var that is specific to the lib. 
Let the jaeger dependency handle its configuration instead

internal/service/adapters.go
reinaldooli: Ops, I'll add the check!

cmd/commom/commom.go
reinaldooli: I was thinking of it, and maybe logging traces to STDOUT isn't a good idea.
Thus, we may panic if the tracing is enabled, and we don't have the tracing url set, wdyt?

reinaldooli: Another thing, is that we should make explicit that we only accept Jaeger, so the configuration will be something like this:
```yaml
api:
  ...
  tracing:
    jaeger:
      enabled: true
      url: "http://jaeger.url/api/traces"
```
Does this make sense?

wl-yankabuki: I like this, this is a better idea, the point is just to make it clearer to the user how to enable/disable the jaeger with the configuration

tobiasbu (approved):
LGTM!

tobiasbu (approved):

PR #583, opened 2023-05-16
Add annotations to the scheduler

manuellysuzik:


test/data/handlers/fixtures/response/schedulers_handler/create_scheduler.json
```diff
@@ -76,6 +76,10 @@
           "metadata": {}
         }
       }
-    ]
+    ],
+    "annotations": [{
+      "name": "annotation",
```
wl-yankabuki, line 81: this looks cumbersome, what about using the annotations as a map like kubernetes, it would look better IMO

```
"annotations": {
     "http://enable-something": "true",
    "http://enable-something-else": "true"
}
```

manuellysuzik: I also do prefer that could be this way but we need to have mapped annotations, right?

internal/adapters/storage/postgres/scheduler/db_scheduler.go
```diff
@@ -113,3 +116,35 @@ func (s *Scheduler) ToScheduler() (*entities.Scheduler, error) {
...
+	if len(annotationMap) == 0 {
+		return nil
+	}
+
+	annotations := make([]*entities.Annotations, len(annotationMap))
+
+	for i, annotation := range annotationMap {
+		for name, opt := range *annotation {
+			annotation := entities.NewAnnotations(name, opt)
+			annotations[i] = annotation
```
reinaldooli, line 145: If there are many annotations, `annotations[i]` will store only the last one. Is this expected?

internal/adapters/storage/postgres/scheduler/scheduler_storage_pg.go
```diff
@@ -398,7 +398,7 @@ func (s schedulerStorage) RunWithTransaction(ctx context.Context, transactionFun
 	s.transactionsMap[transactionID] = transaction
 
 	defer func() {
-		if err := recover(); err != nil {
+		if err := recover(); err != any(nil) {
```
reinaldooli, line 401: Why this?

internal/core/entities/scheduler.go
```diff
@@ -63,6 +63,19 @@ type Scheduler struct {
 	CreatedAt       time.Time
 	MaxSurge        string                 `validate:"required,max_surge"`
 	Forwarders      []*forwarder.Forwarder `validate:"dive"`
+	Annotations     []*Annotations
+}
+
+type Annotations struct {
```
reinaldooli, line 69: It represents one `Annotation`, why are you calling it `Annotations`?

reinaldooli (changes requested):

internal/adapters/storage/postgres/scheduler/db_scheduler.go
manuellysuzik: Could you explain why? I don't understand, because I've tested and all annotations came as expected
Captura de Tela 2023-05-18 às 10 25 14

internal/adapters/storage/postgres/scheduler/scheduler_storage_pg.go
manuellysuzik: The linter in GOLAND was complaining about using nil with an `any` type. So I search for how to fix it.

internal/api/handlers/requestadapters/schedulers_test.go
```diff
@@ -549,6 +549,7 @@ func TestFromApiCreateSchedulerRequestToEntity(t *testing.T) {
 							},
 						},
 					},
+					Annotations: []*api.Annotation{},
```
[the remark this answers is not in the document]
manuellysuzik: Yes. Most of the schedulers won't have annotations. But this will allow for those who want to add annotations. I've added the annotations in some tests, but to be honest I don't know how to test it since the main change happens in the runtime.  When we apply the annotation to the game rooms I think will be possible to add tests. WDYT?

test/data/handlers/fixtures/response/schedulers_handler/create_scheduler.json
wl-yankabuki: yes, you have to change your mapping to use this format

manuellysuzik: I tried to follow the same interpretation from the environment. The YAML is stored as a map but the API contract I think is best to be this way.

reinaldooli:
Pushed a commit that fixes the connection refused error we were seeing.
There are some linter errors and other failing tests, please check them.

manuellysuzik:
> Pushed a commit that fixes the connection refused error we were seeing. There are some linter errors and other failing tests, please check them.

Done! Thank you very much for the help with the broken pipeline <3

reinaldooli (approved):

internal/adapters/storage/postgres/scheduler/db_scheduler.go
```diff
@@ -113,3 +116,31 @@ func (s *Scheduler) ToScheduler() (*entities.Scheduler, error) {
 		Autoscaling:     info.Autoscaling,
 	}, nil
 }
+
+func parseFromAnnotationEntityToMap(entities []*entities.Annotation) *map[string]string {
```
wl-yankabuki, line 120: now that annotation is a map, you don't really need this entity, you can just use the map instead

manuellysuzik: it's ok to not declare an entity to annotations? If it's ok I going to change

wl-yankabuki (approved):

PR #585, opened 2023-05-17
add new metric for v9 to auto scale working

diegopereirawildlife:


wl-yankabuki (changes requested):
you have to implement the reporter for this event as well, check the reporters/dogstatsd.go and reporters/dogstatsd/handlers.go

wl-yankabuki (approved):

PR #586, opened 2023-05-18
Feature/add metric v9 auto scale working

diegopereirawildlife:
Include the correct float parse for auto-scale metric


watcher/watcher.go
```diff
@@ -858,7 +858,7 @@ func (w *Watcher) AutoScale() error {
 		 reporters.Report(reportersConstants.EventWatcherAutoScale, map[string]interface{}{
 			 reportersConstants.TagGame:      w.GameName,
 			 reportersConstants.TagScheduler: w.SchedulerName,
-			 reportersConstants.ValueGauge:   1,
+			 reportersConstants.ValueGauge:   fmt.Sprintf("%.2f", float64(1)),
```
wl-yankabuki, line 861: why not just use 
```suggestion
			 reportersConstants.ValueGauge:  "1.00",
```

wl-yankabuki (approved):

PR #593, opened 2023-05-25
fix pathscheduler message

manuellysuzik:
Refactor PatchScheduler to receive annotations


wl-yankabuki (approved):

internal/core/services/schedulers/patch/patch_scheduler_test.go
```diff
@@ -84,6 +84,24 @@ func TestPatchScheduler(t *testing.T) {
...
+			Output: Output{
+				ChangeSchedulerFunc: func() *entities.Scheduler {
+					scheduler := basicSchedulerToPatchSchedulerTests()
+					scheduler.Annotations = map[string]string{"newannotation": "valueforannotation"}
+
+					return scheduler
+				},
+				Error: nil,
+			},
+		},
```
reinaldooli, line 104: I think we should add a test case with more than one annotation too.

reinaldooli (approved):

PR #601, opened 2023-10-10
Add reason for GR deletion and create kubernetes event

reinaldooli:
### What does this MR do?

- Add reason for the GR deletion;
- Try to create a K8's event before the GR deletion, the event will contain the deletion reason;


internal/adapters/runtime/kubernetes/game_room.go
```diff
@@ -59,7 +63,9 @@ func (k *kubernetes) CreateGameRoomInstance(ctx context.Context, scheduler *enti
 	return instance, nil
 }
 
-func (k *kubernetes) DeleteGameRoomInstance(ctx context.Context, gameRoomInstance *game_room.Instance) error {
+func (k *kubernetes) DeleteGameRoomInstance(ctx context.Context, gameRoomInstance *game_room.Instance, reason string) error {
+	_ = k.createKubernetesEvent(ctx, gameRoomInstance.SchedulerID, gameRoomInstance.ID, "GameRoomDeleted", reason)
```
hspedro, line 67: It seems you're passing `reason` as the last arg which is `message string` in `createKubernetesEvent`. Shouldn't it be inverted with `GameRoomDeleted`?

internal/adapters/runtime/kubernetes/game_room.go
```diff
@@ -84,3 +90,25 @@ func (k *kubernetes) CreateGameRoomName(ctx context.Context, scheduler entities.
...
+		return errors.NewErrUnexpected("error getting game room instance: %s", err)
+	}
+
+	k.getEventRecorder(pod).Event(pod, corev1.EventTypeNormal, reason, message)
+
+	return nil
+}
+
+func (k *kubernetes) getEventRecorder(pod *corev1.Pod) record.EventRecorder {
+	eventBroadcaster := record.NewBroadcaster()
```
hspedro, line 110: Do we need to create a new broadcaster + start recording each time we send an event? Couldn't we instantiate this when creating the `kubernetes` struct and just access via a getter (create if `nil` or return the instance)?

internal/core/operations/rooms/remove/executor.go
```diff
@@ -77,7 +77,7 @@ func (e *Executor) Execute(ctx context.Context, op *operation.Operation, definit
 
 	if removeDefinition.Amount > 0 {
 		logger.Info("start removing rooms", zap.Int("amount", removeDefinition.Amount))
-		err := e.removeRoomsByAmount(ctx, op.SchedulerName, removeDefinition.Amount, op)
+		err := e.removeRoomsByAmount(ctx, op.SchedulerName, removeDefinition.Amount, op, removeDefinition.Reason)
```
hspedro, line 80: nit: at some point, it will be easier to just pass `removeDefinition` ref to those functions and access whatever field they want 😬

internal/core/services/rooms/room_manager.go
```diff
@@ -300,43 +305,53 @@ func (m *RoomManager) SchedulerMaxSurge(ctx context.Context, scheduler *entities
 }
 
 func (m *RoomManager) UpdateGameRoomStatus(ctx context.Context, schedulerId, gameRoomId string) error {
+	if _, err := m.updateGameRoomStatus(ctx, schedulerId, gameRoomId); err != nil {
```
hspedro, line 308: Isn't this the same as `return m.updateGameRoomStatus`? And in that case we could just dedup the functions and have just one `UpdateGameRooomStatus`

hspedro (commented):
Left some minor comments. Overall I think it looks great and it will be super helpful operational-wise, great stuff mate 🚀

hspedro (approved):
LGTM

PR #605, opened 2024-02-20
fix: downscale calculations for a low ready target

reinaldooli:
When having a low `readyTarget` (les than 0.5) the calculation that checks if maestro can downscale was returning wrong values. This was preventing maestro to downscale rooms properly.

### DIFF
--- a/internal/core/services/autoscaler/autoscaler_test.go
+++ b/internal/core/services/autoscaler/autoscaler_test.go
@@ -227,6 +227,25 @@ func TestCanDownscale(t *testing.T) {
 
 			assert.True(t, allow)
 		})
+
+		t.Run("When the ready target is low", func(t *testing.T) {
+			mockRoomStorage := mock.NewMockRoomStorage(ctrl)
+
+			mockRoomStorage.EXPECT().GetRoomCountByStatus(gomock.Any(), scheduler.Name, game_room.GameStatusReady).Return(500, nil)
+			mockRoomStorage.EXPECT().GetRoomCountByStatus(gomock.Any(), scheduler.Name, game_room.GameStatusOccupied).Return(330, nil)
+
+			policy := roomoccupancy.NewPolicy(mockRoomStorage)
+			autoscaler := autoscaler.NewAutoscaler(autoscaler.PolicyMap{policyType: policy})
+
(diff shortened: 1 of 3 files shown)


internal/core/services/autoscaler/policies/roomoccupancy/policy.go
```diff
@@ -113,18 +113,18 @@ func (p *Policy) CanDownscale(policyParameters autoscaling.PolicyParameters, cur
...
+	desiredNumberOfRooms, err := p.CalculateDesiredNumberOfRooms(policyParameters, currentState)
+	if err != nil {
+		return false, fmt.Errorf("Error calculating the desired number of rooms: %w", err)
 	}
 
-	readyRooms := currentState[ReadyRoomsKey].(int)
-	occupiedRooms := currentState[OccupiedRoomsKey].(int)
-
-	occupationRate := float64(occupiedRooms) / float64(occupiedRooms+readyRooms)
+	readyRooms = currentState[ReadyRoomsKey].(int)
```
hspedro, line 126: I think this is causing the ineffassign error by the linter, you don't need to redefine here

hspedro (approved):

internal/core/services/autoscaler/policies/roomoccupancy/policy.go
```diff
@@ -113,18 +113,18 @@ func (p *Policy) CanDownscale(policyParameters autoscaling.PolicyParameters, cur
...
+	desiredNumberOfReadyRooms := int(math.Ceil(float64(desiredNumberOfRooms) * readyTarget))
```
hspedro, line 127: do we still need to convert to int after `math.Ceil` since we need to use as a float on the next line?

internal/core/services/autoscaler/autoscaler_test.go
```diff
@@ -227,6 +227,25 @@ func TestCanDownscale(t *testing.T) {
 
 			assert.True(t, allow)
 		})
+
+		t.Run("When the ready target is low", func(t *testing.T) {
```
hspedro, line 231: I think the name of the test could describe what it should happen when the ready target is low

hspedro (approved):

PR #610, opened 2024-03-17
fix(runtime): check maxUnavailable in existing PDB

hspedro:
### What does this MR do?

- If there's an existing PDB in the namespace that has set maxUnavailable, we need to reset and set minAvailable

### Progress

- [x] Feature
- [x] Integration Tests

### DIFF
--- a/internal/adapters/runtime/kubernetes/scheduler.go
+++ b/internal/adapters/runtime/kubernetes/scheduler.go
@@ -209,14 +209,26 @@ func (k *kubernetes) MitigateDisruption(
 		}
 	}
 
-	currentPdbMinAvailable := pdb.Spec.MinAvailable.IntVal
+	var currentPdbMinAvailable int32
+	// PDB might exist and is based on MaxUnavailable
+	if pdb.Spec.MinAvailable != nil {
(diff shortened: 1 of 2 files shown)


internal/adapters/runtime/kubernetes/scheduler.go
```diff
@@ -209,14 +209,27 @@ func (k *kubernetes) MitigateDisruption(
...
+
 	if currentPdbMinAvailable == int32(float64(roomAmount)*incSafetyPercentage) {
 		return nil
 	}
 
-	pdb.Spec.MinAvailable = &intstr.IntOrString{
-		Type:   intstr.Int,
-		IntVal: int32(float64(roomAmount) * incSafetyPercentage),
+	pdb.Spec = v1Policy.PodDisruptionBudgetSpec{
+		MaxUnavailable: nil,
```
joaobologna, line 223: is this required, even if max unavailable by default is nil?

joaobologna (approved):
LGTM

internal/adapters/runtime/kubernetes/scheduler.go
hspedro: No, I can delete since I'm creting a new struct. Will remove

PR #614, opened 2024-04-25
feat: handle grpc errors & add code metric to forwarder

gussf:


internal/adapters/events/events_forwarder.go
```diff
@@ -46,20 +49,20 @@ func NewEventsForwarder(forwarderClient ports.ForwarderClient) *eventsForwarder
 }
 
 // ForwardRoomEvent forwards room events. It receives the room event attributes and forwarder configuration.
-func (f *eventsForwarder) ForwardRoomEvent(ctx context.Context, eventAttributes events.RoomEventAttributes, forwarder entities.Forwarder) error {
+func (f *eventsForwarder) ForwardRoomEvent(ctx context.Context, eventAttributes events.RoomEventAttributes, forwarder entities.Forwarder) (int32, error) {
```
hspedro, line 52: Why do we return an `int32` as the code? Since we are only adding for metrics and we send the metric as a string, makes sense for me to return it as a string.

internal/adapters/events/events_forwarder.go
```diff
@@ -46,20 +49,20 @@ func NewEventsForwarder(forwarderClient ports.ForwarderClient) *eventsForwarder
...
 	switch eventAttributes.EventType {
 	case events.Arbitrary:
 		if roomEvent, ok := eventAttributes.Other["roomEvent"].(string); ok {
 			event := f.buildRoomEventMessage(eventAttributes, forwarder, roomEvent)
 			eventResponse, err := f.forwarderClient.SendRoomEvent(ctx, forwarder, &event)
 			return handlerGrpcClientResponse(forwarder, eventResponse, err)
 		}
-		return errors.NewErrInvalidArgument("invalid or missing eventAttributes.Other['roomEvent'] field")
+		return UnknownCode, errors.NewErrInvalidArgument("invalid or missing eventAttributes.Other['roomEvent'] field")
```
hspedro, line 60: This should be `codes.InvalidArgument` from gprc, no?

internal/adapters/events/events_forwarder.go
```diff
@@ -46,20 +49,20 @@ func NewEventsForwarder(forwarderClient ports.ForwarderClient) *eventsForwarder
...
 
 	case events.Ping:
 		event, err := f.buildRoomStatusMessage(eventAttributes, forwarder)
 		if err != nil {
-			return errors.NewErrInvalidArgument("failed to build room status message: %s", err)
+			return UnknownCode, errors.NewErrInvalidArgument("failed to build room status message: %s", err)
```
hspedro, line 65: ditto

internal/adapters/events/events_forwarder.go
```diff
@@ -68,19 +71,19 @@ func (f *eventsForwarder) ForwardRoomEvent(ctx context.Context, eventAttributes
 	case events.Status:
 		event, err := f.buildRoomStatusMessage(eventAttributes, forwarder)
 		if err != nil {
-			return errors.NewErrInvalidArgument("failed to build room status message: %s", err)
+			return UnknownCode, errors.NewErrInvalidArgument("failed to build room status message: %s", err)
```
hspedro, line 74: ditto

internal/adapters/events/events_forwarder.go
```diff
@@ -68,19 +71,19 @@ func (f *eventsForwarder) ForwardRoomEvent(ctx context.Context, eventAttributes
...
 		}
 
 		eventResponse, err := f.forwarderClient.SendRoomStatus(ctx, forwarder, &event)
 
 		return handlerGrpcClientResponse(forwarder, eventResponse, err)
 	}
 
-	return errors.NewErrUnexpected("failed to forward event room. event type doesn't exists \"%s\"", eventAttributes.EventType)
+	return UnknownCode, errors.NewErrUnexpected("failed to forward event room. event type doesn't exists \"%s\"", eventAttributes.EventType)
```
hspedro, line 82: This can be a `codes.NotFound`

internal/adapters/events/events_forwarder.go
```diff
@@ -217,12 +220,17 @@ func fromPlayerEventTypeToGrpcPlayerEventType(eventType events.PlayerEventType)
...
+// handlerGrpcClientResponse checks if the gRPC response had unexpected communication or network errors.
+func handlerGrpcClientResponse(forwarder entities.Forwarder, eventResponse *pb.Response, err error) (int32, error) {
 	if err != nil {
-		return err
-	}
-	if eventResponse.Code != 200 {
-		return errors.NewErrUnexpected("failed to forward event room at \"%s\"", forwarder.Name)
+		grpcStatus, ok := status.FromError(err)
+		if !ok {
+			return eventResponse.Code, errors.NewErrUnexpected("failed to forward event room at \"%s\" with unknown grpc code: %w", forwarder.Name, err)
```
hspedro, line 228: why don't we return `grpcStatus` and let go of this custom `.Code`? The function is already specific to grpc

internal/adapters/events/events_forwarder.go
```diff
@@ -217,12 +220,17 @@ func fromPlayerEventTypeToGrpcPlayerEventType(eventType events.PlayerEventType)
 	}
 }
 
-func handlerGrpcClientResponse(forwarder entities.Forwarder, eventResponse *pb.Response, err error) error {
+// handlerGrpcClientResponse checks if the gRPC response had unexpected communication or network errors.
+func handlerGrpcClientResponse(forwarder entities.Forwarder, eventResponse *pb.Response, err error) (int32, error) {
```
hspedro, line 224: can't we always return a `codes.Code` instead of `int32`?

internal/core/services/events/events_forwarder_service.go
```diff
@@ -180,14 +180,17 @@ func (es *EventsForwarderService) forwardRoomEvent(
...
+		reportRoomEventForwardingFailed(scheduler.Game, event.SchedulerID, string(code))
+		es.logger.Error(fmt.Sprintf("Failed to forward room events for room %s and scheduler %s", event.RoomID, event.SchedulerID),
+			zap.Error(err),
+			zap.Int32("code", code))
+
 		return err
 	}
 
-	reportRoomEventForwardingSuccess(scheduler.Game, event.SchedulerID)
+	reportRoomEventForwardingSuccess(scheduler.Game, event.SchedulerID, string(code))
```
hspedro, line 193: What does `string(code)` returns here? From my understanding this is converting an `int32` to string, thus it will report something like `1`, isn't it?

internal/adapters/events/events_forwarder.go
gussf: In this specific case, I opted for the `.Code` due to the parse to status failing

I'm using the .Code whenever the status.Code isn't available

internal/adapters/events/events_forwarder.go
```diff
@@ -68,19 +70,19 @@ func (f *eventsForwarder) ForwardRoomEvent(ctx context.Context, eventAttributes
...
+			return codes.InvalidArgument, errors.NewErrInvalidArgument("failed to build room status message: %s", err)
 		}
 
 		eventResponse, err := f.forwarderClient.SendRoomStatus(ctx, forwarder, &event)
 
 		return handlerGrpcClientResponse(forwarder, eventResponse, err)
 	}
 
-	return errors.NewErrUnexpected("failed to forward event room. event type doesn't exists \"%s\"", eventAttributes.EventType)
+	return codes.NotFound, errors.NewErrUnexpected("failed to forward event room. event type doesn't exists \"%s\"", eventAttributes.EventType)
```
hspedro, line 81: I think we should return `codes.Unknown` and leave `NotFound` strictly to 404 of the client

internal/adapters/events/events_forwarder.go
```diff
@@ -217,12 +219,30 @@ func fromPlayerEventTypeToGrpcPlayerEventType(eventType events.PlayerEventType)
...
+			return codes.Unknown, errors.NewErrUnexpected("failed to forward event room at \"%s\" with unknown grpc code: %s", forwarder.Name, err.Error())
+		}
+
+		return grpcStatus.Code(), errors.NewErrUnexpected("failed to forward event room at \"%s\" with code %s", forwarder.Name, grpcStatus.Code().String())
 	}
-	if eventResponse.Code != 200 {
-		return errors.NewErrUnexpected("failed to forward event room at \"%s\"", forwarder.Name)
+
+	var ret codes.Code
+	// Matchmaker sends the status code as a http one, we normalize it to grpc codes for logs and metrics.
```
hspedro, line 234: nit: would replace with `Forwarder event's  "code" field is being used as an HTTP status code. we normalize it to grpc codes for logs and metrics`

This event definition is actually in this repo

hspedro (approved):
Approved with minor comments

internal/adapters/events/events_forwarder.go
```diff
@@ -217,12 +219,30 @@ func fromPlayerEventTypeToGrpcPlayerEventType(eventType events.PlayerEventType)
...
+	switch eventResponse.Code {
+	case 200:
+		ret = codes.OK
+	case 404:
+		ret = codes.NotFound
+	case 500:
+		ret = codes.Internal
+	default:
+		ret = codes.Unknown
```
hspedro, line 243: Add a case for `406` which MM returns when room does not exist in its Redis.  We can map to `codes.FailedPrecondition`

hspedro (changes requested):
Requesting changes to add the `406` case to the mapping

internal/adapters/events/events_forwarder.go
reinaldooli: This is something we expect to receive when forwarding events, we should add some documentation about this, thus whoever develops a forwarder will know what to return for specific use cases.

gussf: Added both suggestions

hspedro (approved):

PR #618, opened 2024-05-08
[HealthController] Perform rolling update of scheduler versions

hspedro:
* refactor(healthcontroller): perform rolling update

It is now responsibility of the healthcontroller to perform a rolling update. When there was a switch on the current scheduler active version we need to rollout pods with the new version. Health controller either does autoscaling or rolling update.

On rolling update, the following will happen:
    1. Are there rooms with a previous scheduler version? If so, start
       update
    2. Compute how many rooms we can surge (above desired from autoscaling)
    3. Enqueue priority operation to create this rooms
    4. Check how many Ready rooms we have above desired from autoscaling
    5. Enqueue operation to delete those rooms - they are a buffer, should
       never offend readyTarget

* refactor(switchversion): do not replace pods

    The switch version operation will just update the scheduler version.
    Let healthcontroller perform the rolling update

### DIFF
--- a/internal/core/operations/healthcontroller/executor.go
+++ b/internal/core/operations/healthcontroller/executor.go
@@ -51,6 +51,7 @@ type Config struct {
 type Executor struct {
 	autoscaler       ports.Autoscaler
 	roomStorage      ports.RoomStorage
+	roomManager      ports.RoomManager
 	instanceStorage  ports.GameRoomInstanceStorage
 	schedulerStorage ports.SchedulerStorage
 	operationManager ports.OperationManager
@@ -60,10 +61,20 @@ type Executor struct {
 var _ operations.Executor = (*Executor)(nil)
 
 // NewExecutor creates a new instance of Executor.
-func NewExecutor(roomStorage ports.RoomStorage, instanceStorage ports.GameRoomInstanceStorage, schedulerStorage ports.SchedulerStorage, operationManager ports.OperationManager, autoscaler ports.Autoscaler, config Config) *Executor {
+func NewExecutor(
+	roomStorage ports.RoomStorage,
+	roomManager ports.RoomManager,
+	instanceStorage ports.GameRoomInstanceStorage,
+	schedulerStorage ports.SchedulerStorage,
+	operationManager ports.OperationManager,
+	autoscaler ports.Autoscaler,
+	config Config,
+) *Executor {
 	return &Executor{
-		autoscaler:       autoscaler,
+		autoscaler: autoscaler,
+		// TODO: replace roomStorage operations with roomManager
 		roomStorage:      roomStorage,
+		roomManager:      roomManager,
 		instanceStorage:  instanceStorage,
 		schedulerStorage: schedulerStorage,
 		operationManager: operationManager,
@@ -99,7 +110,7 @@ func (ex *Executor) Execute(ctx context.Context, op *operation.Operation, defini
 
 	if len(expiredRooms) > 0 {
 		logger.Sugar().Infof("found %v expired rooms to be deleted", len(expiredRooms))
-		err = ex.enqueueRemoveExpiredRooms(ctx, op, logger, expiredRooms)
+		err = ex.enqueueRemoveRooms(ctx, op, logger, expiredRooms)
 		if err != nil {
 			logger.Error("could not enqueue operation to delete expired rooms", zap.Error(err))
 		}
@@ -113,6 +124,11 @@ func (ex *Executor) Execute(ctx context.Context, op *operation.Operation, defini
 	}
 	reportDesiredNumberOfRooms(scheduler.Game, scheduler.Name, desiredNumberOfRooms)
 
+	// Check if the system is in a rollingUpdate by listing rooms that are not the current scheduler version
+	roomsPreviousSchedulerVersion, isRollingUpdate := ex.checkRollingUpdate(ctx, logger, scheduler, availableRooms)
+	if isRollingUpdate {
+		return ex.performRollingUpdate(ctx, op, def, logger, scheduler, desiredNumberOfRooms, availableRooms, roomsPreviousSchedulerVersion)
+	}
 	err = ex.ensureDesiredAmountOfInstances(ctx, op, def, scheduler, logger, len(availableRooms), desiredNumberOfRooms)
 	if err != nil {
 		logger.Error("cannot ensure desired amount of instances", zap.Error(err))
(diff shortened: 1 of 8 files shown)


gmurayama (approved):

internal/adapters/runtime/kubernetes/scheduler.go
```diff
@@ -219,6 +219,10 @@ func (k *kubernetes) MitigateDisruption(
 		return nil
 	}
 
+	pdb, err = k.clientSet.PolicyV1().PodDisruptionBudgets(scheduler.Name).Get(ctx, scheduler.Name, metav1.GetOptions{})
+	if err != nil || pdb == nil {
+		return errors.NewErrUnexpected("non recoverable error when getting PDB for scheduler '%s': %s", scheduler.Name, err)
+	}
```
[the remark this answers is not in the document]
hspedro: In theory, the PDB object can be changed in the runtime in the meantime after we getting the initial state/ask for creation (beginning of the function) and before updating the value.  This should never happen in production because there is only one agent setting this PDB in the namespace and it's Maestro worker. However, on tests we were seeing intermittent failures running parallel cases, hence why I'm adding this to be safer before updating the object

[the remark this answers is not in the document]
hspedro: This is the kind of error thrown in that scenario: https://github.com/topfreegames/maestro/actions/runs/9422115862/job/25957670592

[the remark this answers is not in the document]
gmurayama: Nice, thanks for the clarification. It is kinda counterintuitive just by reading the code, perhaps add a small context as a comment?

internal/core/operations/schedulers/switchversion/executor.go
```diff
@@ -132,179 +81,14 @@ func (ex *Executor) Execute(ctx context.Context, op *operation.Operation, defini
...
-	err := ex.deleteNewCreatedRooms(ctx, logger, op.SchedulerName)
-	ex.clearNewCreatedRooms(op.SchedulerName)
-	if err != nil {
-		logger.Error("error deleting newly created rooms", zap.Error(err))
-		deleteCreatedRoomsErr := fmt.Errorf("error rolling back created rooms: %w", err)
-		ex.operationManager.AppendOperationEventToExecutionHistory(ctx, op, deleteCreatedRoomsErr.Error())
-		return err
-	}
-
-	logger.Info("finished Rollback routine")
```
joaobologna, line 158: switch version will no longer have rollbacks?

internal/core/operations/healthcontroller/executor.go
```diff
@@ -369,3 +385,122 @@ func (ex *Executor) canPerformDownscale(ctx context.Context, scheduler *entities
...
+		logger.Error("failed to perform rolling update while getting max surge amount of rooms", zap.Error(err))
+		return err
+	}
+	maxRoomsToSurge := desiredNumberOfRooms + maxSurgeAmount - len(availableRoomsIDs)
+	if len(roomsWithPreviousSchedulerVersion) < maxRoomsToSurge {
+		maxRoomsToSurge = len(roomsWithPreviousSchedulerVersion)
+	}
+	if maxRoomsToSurge <= 0 {
+		maxRoomsToSurge = 1
+	}
```
joaobologna, line 433: can you expand what scenario each equation covers? for example, when can `maxRoomsToSurge` be negative? if desired is in considering downscale?

internal/core/operations/healthcontroller/executor.go
```diff
@@ -369,3 +385,122 @@ func (ex *Executor) canPerformDownscale(ctx context.Context, scheduler *entities
...
+	bufferRoomsToBeRemoved := len(readyRoomIDs) - desiredNumberOfRooms
+	if bufferRoomsToBeRemoved < 0 {
+		logger.Sugar().Debugf("can not delete old rooms without offending autoscale: %d", bufferRoomsToBeRemoved)
+		return []string{}, nil
+	}
+	if bufferRoomsToBeRemoved > len(roomsWithPreviousSchedulerVersion) {
+		return roomsWithPreviousSchedulerVersion, nil
+	}
+	return roomsWithPreviousSchedulerVersion[:bufferRoomsToBeRemoved], nil
```
joaobologna, line 505: should we consider the down threshold here?

internal/core/operations/schedulers/switchversion/executor.go
hspedro: No. Previously, the rollback synchronously looped through all newly created rooms and deleted. This generated some issues:

1. Sync operation can take a while to ask for deletion of all game rooms and no operations would run for that scheduler
2. The deletion would not respect any scaling mechanism, thus it would lead to a drastic reduction in the number of game rooms

Now, the switch version can only fail if the update in SQL fails. Since it is ACID, if the update fails then the transaction fails in the DB as well, thus we will not end up in a situation where we need to roll back a transaction manually in the DB.

The following will happen:

1. Operation to switch scheduler version from v1 to v2
2. Change in the DB the active version to v2
3. Next _health_controller_ operation will check that we have a rolling update - still exists pods in the non-active versions
4. _health_controller_ performs rolling update adding v2 and deleting v1 pods
5. At some point, one wants to roll back to v1 and will send an API command to switch the scheduler version. This will be translated into the switch version operation to the worker for v1
6. _switch_version_ will change in the DB the version back to v1
7. Next, the _health_controller_ operation detects an ongoing rolling update to v1
8. _health_controller_ performs rolling update adding v1 and deleting v2 pods

This is in line with how Kubernetes does. Rolling back a version triggers a new deployment and the system in each iteration can read its current state and determine what actions to take, rather than trying to control the full lifecycle of the operation in one batch as we were doing on the previous _switch_version_

internal/core/operations/healthcontroller/executor.go
hspedro: It's an overkill. With this computation, I'm only deleting rooms considering the `ready` amount, it's similar to Kubernetes when you define `maxUnavailable: 0`. We will never go below the desired number of rooms and we always downscale to the desired.

I think that during rolling update, the games don't want to fine-grain control how aggressive we are downscaling the pods if we do it safely. The only downside of this approach is that the process will take longer because we never go below the desired, that's when people generally use `maxUnavailble: 25%` or something like it

joaobologna: makes sense :) 

unrelated: sice we do `len(readyRoomIDs) - desiredNumberOfRooms`, can we ever get stuck or take a long time to renew rooms?

joaobologna: before scaling and rolling updates were two district actions, right? now, they can be done together (as i understand), if so, it's nice to respect the same behavior if possible

one pro for not doing is the possibility where we never/take too long to renew pods because of down threshold

PR #619, opened 2024-05-09
fix(scheduler): optional downThreshold

hspedro:
Many clients forget to update downThreshold or simply do not care how Maestro uses it for downscale. To reduce their cognitive load, we can make the parameter optional and if not set always downscale (0.99)

## Changelog

- The `scheduler.autoscaling.policy.parameters.roomOccupancy.downThreshold` parameter is now optional with a default value of 0.99 to always allow downscale
- Enhanced error messages to be more descriptive of values out of bounds


internal/core/services/autoscaler/policies/roomoccupancy/policy.go
```diff
@@ -85,7 +85,7 @@ func (p *Policy) CalculateDesiredNumberOfRooms(policyParameters autoscaling.Poli
 
 	readyTarget := policyParameters.RoomOccupancy.ReadyTarget
 	if readyTarget >= float64(1) || readyTarget <= 0 {
-		return -1, errors.New("Ready target must be between 0 and 1")
+		return -1, errors.New("Ready target must be between 0 and 0.99")
```
joaobologna, line 88: why not set the message as `Ready target must be greater than 0 and less than 1`?

internal/core/entities/autoscaling/autoscaling.go
```diff
@@ -87,5 +88,5 @@ type RoomOccupancyParams struct {
 	// ReadyTarget indicates the target percentage of ready rooms a scheduler should maintain.
 	ReadyTarget float64 `validate:"gt=0,lte=0.9"`
 	// DownThreshold indicates the percentage of occupied rooms a scheduler should have to trigger a downscale event.
-	DownThreshold float64 `validate:"gte=0,lte=0.99"`
+	DownThreshold float64 `validate:"gt=0,lt=1"`
```
joaobologna, line 91: nit: 

```suggestion
	DownThreshold float64 `validate:"gt=0,lte=0.99"`
```

to be the same as `ReadyTarget`

internal/api/handlers/requestadapters/schedulers.go
```diff
@@ -372,6 +372,9 @@ func fromApiRoomOccupancyPolicyToEntity(roomOccupancy *api.RoomOccupancy) *autos
 	roomOccupancyParam.ReadyTarget = float64(readyTarget)
 
 	downThreshold := roomOccupancy.GetDownThreshold()
+	if downThreshold == 0 {
```
joaobologna, line 375: do we have a test to assert if default threshold is assigned if == 0?

hspedro: We have tests that validate that inputs without `DownThreshold` actually produce outputs with `DownThreshold: float64(autoscaling.DefaultDownThreshold)`. The `.GetDownThreshold` will return 0 if the value is not set. If set, but is 0, it will fail on the API validation thus it will never reach this code

joaobologna: great! ty

joaobologna (approved):

PR #621, opened 2024-05-27
Cancel operations upon worker shutdown

hspedro:
When the worker receives a termination signal, it will close the contexts and in-progress operations stay forever in this state. Adding a wrap in the execute function that observes context cancellations and propagates this error to be handled by the parent function, which will set the state to either `Canceled` or `Error`.


internal/core/worker/operationexecution/operation_execution_worker.go
```diff
@@ -161,65 +167,77 @@ func (w *OperationExecutionWorker) executeOperationFlow(operationID string) erro
...
 
-	w.finishOperationAndLease(op, def, loopLogger)
-
 	return nil
 }
 
-func (w *OperationExecutionWorker) rollbackOperation(op *operation.Operation, def operations.Definition, executionErr error, loopLogger *zap.Logger, executor operations.Executor) {
-	w.operationManager.AppendOperationEventToExecutionHistory(w.workerContext, op, "Starting operation rollback")
+func (w *OperationExecutionWorker) rollbackOperation(ctx context.Context, op *operation.Operation, def operations.Definition, executionErr error, loopLogger *zap.Logger, executor operations.Executor) {
+	w.operationManager.AppendOperationEventToExecutionHistory(ctx, op, "Starting operation rollback")
```
joaobologna, line 180: that is a missing history log that we know have, is that right?

internal/core/worker/operationexecution/operation_execution_worker.go
```diff
@@ -161,65 +167,77 @@ func (w *OperationExecutionWorker) executeOperationFlow(operationID string) erro
...
+		msg = "operation canceled by the user"
+		var cancelFunc func()
+		executionErrCtx, cancelFunc = context.WithTimeout(context.Background(), TerminationGracePeriodTimeout)
+		defer cancelFunc()
 	} else {
 		op.Status = operation.StatusError
-
-		loopLogger.Error("operation execution failed", zap.Error(executionErr))
-		w.operationManager.AppendOperationEventToExecutionHistory(w.workerContext, op, "Operation execution failed")
+		msg = "operation execution failed"
```
joaobologna, line 204: when will this branch execute? an execution error that is not `executionErr.Error()` or `context.Canceled.Error()`

so what is it?

internal/core/worker/operationexecution/operation_execution_worker.go
```diff
@@ -161,65 +167,77 @@ func (w *OperationExecutionWorker) executeOperationFlow(operationID string) erro
...
-	w.operationManager.AppendOperationEventToExecutionHistory(w.workerContext, op, "Operation finished")
+	w.operationManager.AppendOperationEventToExecutionHistory(ctx, op, "Operation finished")
 }
 
-func (w OperationExecutionWorker) executeOperationWithLease(operationContext context.Context, op *operation.Operation, def operations.Definition, executor operations.Executor) error {
-	return w.executeCollectingLatencyMetrics(op.DefinitionName, func() error {
-		return executor.Execute(operationContext, op, def)
-	})
+func (w OperationExecutionWorker) executeOperationWithLease(operationContext context.Context, op *operation.Operation, def operations.Definition, executor operations.Executor) (err error) {
+	done := make(chan error, 1)
```
joaobologna, line 229: cant we catch context cancelled in the function that calls `executeOperationWithLease`? i think the err return will be context canceled

internal/core/worker/operationexecution/operation_execution_worker.go
```diff
@@ -161,65 +167,77 @@ func (w *OperationExecutionWorker) executeOperationFlow(operationID string) erro
...
+	go func() {
+		done <- w.executeCollectingLatencyMetrics(op.DefinitionName, func() error {
+			return executor.Execute(operationContext, op, def)
+		})
+	}()
+	select {
+	case <-operationContext.Done():
```
joaobologna, line 236: i would vote for handling context cancelation in the function that actually calls this method

why? because its the caller responsibility to deal with that i would say

internal/core/worker/operationexecution/operation_execution_worker_test.go
```diff
@@ -458,6 +456,7 @@ func TestSchedulerOperationsExecutionLoop(t *testing.T) {
 		go func() {
 			pendingOpsChan <- expectedOperation.ID
 
+			time.Sleep(10 * time.Millisecond)
```
joaobologna, line 459: why have we added time sleep for these tests? and follow up: isnt 10s too high?

joaobologna (approved):
LGTM

PR #622, opened 2024-06-05
tests: fix cancel msg on operation worker

hspedro:
Fixing failed run by changing the expected cancel string: https://github.com/topfreegames/maestro/actions/runs/9384998563/job/25842068782


joaobologna (approved):
LGTM

PR #623, opened 2024-06-05
[v9.17.0] Maestro v9 to coexist with other Game Room providers

hspedro:
* refactor(controller): do not fail scheduler if namespace exists

When creating a scheduler, we fail if the namespace already exists. This is problematic if there are other GR providers running at the same namespace, like maestro v10 for example. Thus, don't fail scheduler creation if namespace exists.

* refactor(controller): do not delete namespace on scheduler delete

There can be other GR providers running into the same namespace, thus we don't want to cascade delete the k8s namespace on scheduler deletion

* refactor(watcher): filter pod events by heritage label

There can be other resources/pods running in the namespace that are not managed by Maestro. In that case, we want to filter the watcher to only receive pod events from the ones created by v9

### DIFF
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -87,26 +87,23 @@ func CreateScheduler(
 		return err
 	}
 
-	if exists {
-		logger.Error("namespace already exists, aborting scheduler creation")
-		return fmt.Errorf(`namespace "%s" already exists`, namespace.Name)
-	}
-
-	err = mr.WithSegment(models.SegmentNamespace, func() error {
-		return namespace.Create(clientset)
-	})
-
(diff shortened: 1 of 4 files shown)


controller/controller.go
```diff
@@ -840,7 +837,8 @@ func UpdateSchedulerConfig(
 }
 
 // MustUpdatePods returns true if it's necessary to delete old pod and create a new one
-//  so this have the new configuration.
+//
+//	so this have the new configuration.
```
joaobologna, line 841: is this intended?

joaobologna (approved):
LGTM

PR #625, opened 2024-06-17
fix(forwarder): port v9 keepalive configs

hspedro:
The forwarder will use KeepAlive params from GRPC, that uses HTTP/2 ping frames, to check for broken connections and refresh. If not used, we'll default to whatever the underlying OS is using to manage the port - which is around 15-20 minutes.

Porting the work of #600


miguelreiswildlife (approved):

joaobologna (approved):
LGTM

PR #628, opened 2024-06-27
Do not return error on not found for delete operations

miguelreiswildlife:


hspedro (commented):
LGTM after fixing tests and linter. Also, on tests, you could add a test case for the remove executor in which we don't return error on `NotFound`

hspedro (approved):
LGTM

PR #639, opened 2024-09-23
refactor(disruption): use fixed maxUnavailable

hspedro:
Some schedulers might decide on a high `readyTarget`, which means a low amount of occupied rooms in regards to total rooms, this causes the PDB to have a relatively low number compared to the total. Since PDB does not filter by room status, we are still subject to many rooms being evicted and the system is highly impacted. To mitigate this, Maestro will now set a fixed amount for maxUnavailable with the default value of 5%. It can be configured via:

  MAESTRO_SERVICES_SCHEDULERMANAGER_DEFAULTPDBMAXUNAVAILABLE="10%"

The variable accepts integers greater than 0 and percentage strings.

## Summary of changes

- Added a new variable on Scheduler proto for `pdbMaxUnavailable` that is a string
- Added this variable to the payloads of creating and updating a scheduler
- Added validation for this field checking for integers below 1 and percentage values below 0 and above 100
- Added a config to SchedulerManager with the DefaultPdbMaxUnavailable. This config is in the YAML as well with the default value of 5%
- Refactored RuntimeWatcher to remove the MitigateDisruption call based on the number of occupied rooms
- Refactored runtime Scheduler with Create, Delete, and UpdateScheduler functions that update the namespace PDBs
- Added a check to Scheduler's NewVersion operation to check if pdb max unavailable changed and call the runtime update to reflect the PDB change

## Scheduler Usage

```yaml
pdbMaxUnavailable: "15%" # or 50 (any integer >1)
```


reinaldooli:
I believe it's better to make this configuration scheduler/game basis. We might end up with different games in the same cluster asking for different configuration values.

internal/core/entities/pdb/pdb.go
```diff
@@ -0,0 +1,81 @@
...
+const (
+	Int    Type = iota // The IntOrString holds an int.
+	String             // The IntOrString holds a string.
+)
+
+func ValidateDefault(defaultVal string) (string, error) {
+	if strings.HasSuffix(defaultVal, "%") {
+		percentageStr := strings.TrimSuffix(defaultVal, "%")
+		percentage, err := strconv.Atoi(percentageStr)
+		if err != nil || percentage < 0 || percentage > 100 {
```
joaobologna, line 49: if we dont allow `< 1` if not percentage, why allow here?

internal/core/validations/validations.go
```diff
@@ -74,6 +74,31 @@ func IsMaxSurgeValid(maxSurge string) bool {
 	return true
 }
 
+// IsPdbMaxUnavailableValid check if PdbMaxUnavailable is valid. A PdbMaxUnavailable valid is either
+// a string that can be converted to a number greater than 0, or a string with percentage that its
+// value is greater than 0 and less than 100. Empty strigs are valid, we'll use the defualt value on
+// SchedulerManager
+func IsPdbMaxUnavailableValid(pdbMaxUnavailable string) bool {
```
joaobologna, line 81: why duplicate this validation here?

config/config.yaml
```diff
@@ -74,6 +76,8 @@ operations:
       limit: 1000
 
 services:
+  schedulerManager:
+    defaultPdbMaxUnavailable: "5%"
```
joaobologna, line 80: what is this for? from the doc, it says default will be 10%

> *pdbMaxUnavailable**: Defines the disruption budget for game rooms. Optional and defaults to 10%.

also, whats the difference between this and 

```
scheduler:
        maxUnavailable: 10%
```

internal/core/entities/pdb/pdb.go
hspedro: Good point. 1% might be a valid percentage, but probably allow setting 1 as non-percentage number as well. WDYT?

internal/core/validations/validations.go
hspedro: At first, it was because at one point we're validating the arguments that come from the HTTP API and at the other, we want to validate what we're reading from the Viper config. But I guess I can just call the same function when reading Viper config

config/config.yaml
hspedro: scheduler.maxUnavailable is garbage that I forgot to remove. I'll update the docs with the 5% default

internal/core/validations/validations.go
hspedro: I've refactored to use the same validation function and removed this one 🙏🏽

internal/core/entities/pdb/pdb.go
hspedro: I've removed this function in favor of the API one that specifies >0 and <100 for percentage string and >0 for raw numbers

joaobologna: great!

docs/reference/Scheduler.md
```diff
@@ -403,4 +406,13 @@ It is represented as:
...
+- **hostPortRange**: The [port range](#portrange) for the port to be allocated in the host. Mutually exclusive with the port range configured in the root structure.
+
+#### PDB Max Unavailable
+
+A string value that defines the disruption budget of Game Rooms from a specific scheduler.
+Maestro will create a [PDB Resource](https://kubernetes.io/docs/tasks/run-application/configure-pdb/)
+to prevent evictions drastically impacting availability of the Game Rooms.
+
+By default this value is set to 10%, so at worst runtime can evit 10% of the pods. There is no way to control
+what pods will be evicted - if it prefers ready, pending, etc.
```
joaobologna, line 418:
```suggestion
By default, this value is set to 5%, so at worst runtime can evict 5% of the pods. There is no way to control
what pods will be evicted - if it prefers ready, pending, etc.
```

it's 5% right?

internal/core/entities/scheduler_test.go
```diff
@@ -93,6 +93,7 @@ func TestNewScheduler(t *testing.T) {
 	forwarders := []*forwarder.Forwarder{fwd}
 	annotations := map[string]string{"imageregistry": "https://hub.docker.com/"}
 	labels := map[string]string{"scheduler": "scheduler-name"}
+	pdbMaxUnavailable := "10%"
```
joaobologna, line 96:
```suggestion
	pdbMaxUnavailable := "5%"
```

nit

internal/core/services/schedulers/patch/patch_scheduler.go
```diff
@@ -49,7 +49,8 @@ const (
 	LabelAutoscaling = "autoscaling"
 	// LabelSchedulerForwarders is the forwarders key in the patch map.
 	LabelSchedulerForwarders = "forwarders"
-
+	// LabelTopologyPDBMaxUnavailable is the PDB's maxUnavailable spec key in the patch map.
```
joaobologna, line 52:
```suggestion
	// LabelPDBMaxUnavailable is the PDB's maxUnavailable spec key in the patch map.
```

joaobologna (approved):
LGTM

docs/reference/Scheduler.md
```diff
@@ -403,4 +406,13 @@ It is represented as:
...
\ No newline at end of file
+- **hostPortRange**: The [port range](#portrange) for the port to be allocated in the host. Mutually exclusive with the port range configured in the root structure.
+
+#### PDB Max Unavailable
+
+A string value that defines the disruption budget of Game Rooms from a specific scheduler.
+Maestro will create a [PDB Resource](https://kubernetes.io/docs/tasks/run-application/configure-pdb/)
+to prevent evictions drastically impacting availability of the Game Rooms.
+
+By default this value is set to 10%, so at worst runtime can evit 10% of the pods. There is no way to control
```
reinaldooli, line 417: wdyt about adding a ref link to the configuration?

docs/reference/Scheduler.md
hspedro: Right, thanks!

docs/reference/Scheduler.md
hspedro: Do you mean the kubernetes configuration doc or maestro yaml/env configuration?

internal/api/handlers/schedulers_handler.go
```diff
@@ -169,6 +169,7 @@ func (h *SchedulersHandler) NewSchedulerVersion(ctx context.Context, request *ap
 		return nil, status.Error(codes.InvalidArgument, apiValidationError.Error())
 	}
 
+	fmt.Println("New Scheduler Version?")
```
reinaldooli, line 172:
```suggestion
```

docs/reference/Scheduler.md
hspedro: I've added a document specifying the order of precedence in Maestro configuration, also there's a link to kubernetes

PR #640, opened 2024-10-08
fix(room_manager): only sort rooms removal if major version diff

hspedro:
We only want to apply sorting by version if the room has a major version of difference compared to the active scheduler version. This prevents minor updates forcing the sort, which would cause occupied rooms to be deleted first (from a prior minor version) when we don't want that behavior. Thus, check if it has a major version diff between room and active scheduler

### DIFF
--- a/internal/core/entities/scheduler.go
+++ b/internal/core/entities/scheduler.go
@@ -26,6 +26,7 @@ import (
 	"errors"
 	"time"
 
+	"github.com/Masterminds/semver"
 	"github.com/topfreegames/maestro/internal/core/entities/autoscaling"
 	"github.com/topfreegames/maestro/internal/core/entities/port"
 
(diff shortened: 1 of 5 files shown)


internal/core/entities/scheduler.go
```diff
@@ -202,3 +204,25 @@ func (s *Scheduler) HasValidPortRangeConfiguration() error {
 
 	return nil
 }
+
+func (s *Scheduler) HasMajorVersionDifference(otherVersion string) bool {
+	return extractMajorVersion(otherVersion) != extractMajorVersion(s.Spec.Version)
```
reinaldooli, line 209: You can use the `semver` lib and just check the major difference, like [here](https://github.com/topfreegames/maestro/blob/4fe0422078ef7a1144dca887397fae06a8561250/internal/core/operations/schedulers/newversion/executor.go#L310)

hspedro: fixed to use this lib

PR #641, opened 2024-10-08
feat: add occupied slots to game rooms api

reinaldooli:
This is part of the changes to allow Maestro calculations to be based on the number of matches running in a Game Server.
These changes will allow maestro to receive the number of running matches in a Game Server.

### Changes
- Add `running_matches` param to the service definition files (proto)
- Create/Update Room Storage with the number of running matches. Later we'll be able to aggregate the values in this storage to know the total amount of running matches by scheduler.


internal/core/entities/game_room/game_room.go
```diff
@@ -130,6 +130,7 @@ type GameRoom struct {
 	IsValidationRoom bool
 	LastPingAt       time.Time
 	CreatedAt        time.Time
+	OccupiedSlots    int
```
hspedro, line 133: nit: why don't we use the same naming as the proto? IMO it would simplify maintenance and they seem to map 1:1, right?

hspedro, line 133: nit: why don't we use the same naming as the proto? IMO it would simplify maintenance and they seem to map 1:1, right?

hspedro (approved):

PR #643, opened 2024-10-15
refactor(runtime): make topologySpread configurable

hspedro:
Read the TopologySpreadConstraints maxSkew, topologyKey, and whenUnsatisfiable from the YAML config so we can tune based on cluster/game. By default, it will be set to 5. Using a low value leads to underutilized nodes which impacts in cost


joaobologna (approved):
LGTM

PR #646, opened 2024-10-29
feat: allow maestro to deal with multiple matches per game server

reinaldooli:
This pull request introduces changes to enable Maestro to manage multiple matches per Game Server effectively.

- **MatchAllocation Configuration**: Adds a **MatchAllocation** block to the **Scheduler** configuration with two parameters: **MaxMatches** (maximum number of matches per game room) and **MinFreeSlots** (minimum number of free slots required for a room to be considered available again).

- **Backward Compatibility**: If **MaxMatches** is set to 1 (default value), Maestro will continue its previous behavior without slot-based scaling. The new match-based calculations apply only when **MaxMatches** is greater than 1.

- **New Game Server States**: Introduces two new statuses for Game Servers with MaxMatches greater than 1:
  - **Active:** Game Servers that are hosting matches but are not yet at full capacity.
  - **Cooling:** Game Servers that have reached full capacity and will only transition back to **Active** after the number of running matches falls below **MinFreeSlots**.

These updates provide more granular control over room availability and autoscaling, helping to optimize server utilization when using a multiple-matches-per-room strategy.

### DIFF
--- a/internal/core/entities/allocation/allocation.go
+++ b/internal/core/entities/allocation/allocation.go
@@ -29,7 +29,7 @@ type MatchAllocation struct {
 	MaxMatches int `validate:"required,min=1,max=30"`
 	// MinFreeSlots defines the minimum number of free slots that a room must
 	// have to become available for allocation.
-	MinFreeSlots int `validate:"required,min=1"`
+	MinFreeSlots int `validate:"required,min=1,max=30"`
 }
 
 func NewDefaultMatchAllocation() *MatchAllocation {


internal/core/entities/allocation/allocation.go
```diff
@@ -0,0 +1,40 @@
...
+package allocation
+
+// MatchAllocation is the struct that defines the configuration for the match
+// allocation in the Game Rooms.
+type MatchAllocation struct {
+	// MaxMatches defined the maximum number of matches that a room can host.
+	MaxMatches int `validate:"required,min=1,max=30"`
+	// MinFreeSlots defines the minimum number of free slots that a room must
+	// have to become available for allocation.
+	MinFreeSlots int `validate:"required,min=1"`
```
hspedro, line 32: we should add `max=30` here because the upper limit is MaxMatches, right? And in addition we'll validate one value against the other as well but `max=30` we "gain" a validation prior to us checking one against the other

PR #648, opened 2024-11-19
docs: add best practices

tsuzukayama:

### DIFF
--- a/docs/tutorials/BestPractices.md
+++ b/docs/tutorials/BestPractices.md
@@ -32,15 +32,18 @@ This section outlines a few recommended defaults games that uses Maestro.These a
 
 ###  Minimum number of rooms
 
-They were defined by getting the minimum number of rooms that each region had during a 1 week avaliation window.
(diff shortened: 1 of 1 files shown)


docs/tutorials/BestPractices.md
```diff
@@ -0,0 +1,46 @@
...
+**Recommendation:** Set the minimum number of rooms to match or exceed the baseline of the previous scheduler. This provides a buffer to maintain stability during the transition.
+
+### 2. Post-Rollout Adjustments:
+
+#### Rollback Ready Target / Min Room Count:
+
+Once the rollout has finished, you can rollback the ready target and/or minimum number of rooms to their standard levels.
+
+
+## Recommended defaults for games
```
hspedro, line 29: I see this section as more about explaining the benefits of setting the min and the rationale we used to come up with those numbers. I wouldn't track TFG games usage in the OSS repo, these numbers are dynamic and will change, new games will be added (internal vs external names), etc.

We have an internal repo for these kind of documentation as well

PR #652, opened 2024-11-26
fix(healthcontroller): cap room removal in update finish

hspedro:
When rolling update is finishing, we should cap rooms to be deleted to the amount of rooms leftover in non-active scheduler versions

### DIFF
--- a/internal/core/operations/healthcontroller/executor.go
+++ b/internal/core/operations/healthcontroller/executor.go
@@ -134,7 +134,7 @@ func (ex *Executor) Execute(ctx context.Context, op *operation.Operation, defini
 	}
 	reportDesiredNumberOfRooms(scheduler.Game, scheduler.Name, desiredNumberOfRooms)
 
-	err = ex.ensureDesiredAmountOfInstances(ctx, op, def, scheduler, logger, len(availableRooms), desiredNumberOfRooms, isRollingUpdating)
+	err = ex.ensureDesiredAmountOfInstances(ctx, op, def, scheduler, logger, availableRooms, desiredNumberOfRooms, isRollingUpdating)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


reinaldooli (approved):

internal/core/operations/healthcontroller/executor.go
```diff
@@ -451,6 +452,32 @@ func GetDesiredNumberOfRooms(
 	return
 }
 
+func AdjustDownscaleAmount(ctx context.Context, availableRooms []string, amountToBeDeleted int, roomStorage ports.RoomStorage, scheduler *entities.Scheduler, logger *zap.Logger) (downscaleAmount int) {
```
miguelreiswildlife, line 455: wdyt of naming this something like `CapRollingUpdateDownscaleAmount`

internal/core/operations/healthcontroller/executor.go
```diff
@@ -451,6 +452,32 @@ func GetDesiredNumberOfRooms(
...
+		room, err := roomStorage.GetRoom(ctx, scheduler.Name, roomID)
+		if err != nil {
+			logger.Warn("error getting room while downscaling to check delete cap", zap.Error(err))
+			continue
+		}
+		if room.Version != scheduler.Spec.Version {
+			roomAmountWithInactiveVersion += 1
+		}
+		if roomAmountWithInactiveVersion >= amountToBeDeleted {
+			break
```
miguelreiswildlife, line 468:
```suggestion
			return roomAmountWithInactiveVersion
```

miguelreiswildlife (approved):

internal/core/operations/healthcontroller/executor.go
hspedro: Nice, changed

PR #656, opened 2025-01-30
fix: downscale calculation considering matches

reinaldooli:
The `CanDowscale` method from the `RoomsOccupancy` policy was using the ready rooms to calculate when allowing/denying the down scale. Now it's considering the amount of desired slots over the amount of current free slots.

### DIFF
--- a/internal/core/services/autoscaler/policies/roomoccupancy/policy.go
+++ b/internal/core/services/autoscaler/policies/roomoccupancy/policy.go
@@ -36,12 +36,12 @@ import (
 )
 
 const (
-	// ReadyRoomsKey is the key to total rooms in the CurrentState map.
-	ReadyRoomsKey = "RoomsOccupancyTotalRooms"
 	// RunningMatchesKey is the key to running matches in the CurrentState map.
 	RunningMatchesKey = "RoomsOccupancyRunningMatches"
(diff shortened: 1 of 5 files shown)


internal/core/services/autoscaler/policies/roomoccupancy/policy.go
```diff
@@ -61,75 +61,98 @@ func NewPolicy(roomStorage ports.RoomStorage) *Policy {
...
+	if err != nil {
+		return nil, fmt.Errorf("error fetching active game rooms amount: %w", err)
+	}
+
+	occupiedCount, err := p.roomStorage.GetRoomCountByStatus(ctx, scheduler.Name, game_room.GameStatusOccupied)
+	if err != nil {
+		return nil, fmt.Errorf("error fetching occupied game rooms amount: %w", err)
+	}
+
+	totalFreeSlots := (readyCount + activeCount + occupiedCount) * scheduler.MatchAllocation.MaxMatches
```
hspedro, line 79: Why are we counting occupied as a free slot? Shouldn't we count how many matches are actively running against the number of total possible matches?

hspedro: Oh, nvm, just saw the calculation below

hspedro (approved):
LGTM

PR #660, opened 2025-04-11
fix: sort operations before pagination

tsuzukayama:
The sorting for operations is currently broken. It applies the sort function after the pagination happened. That way, it won´t sort by the oldest operations, it'll only sort the current page. This MR fixes that by applying the Sort function on the Redis operation, sorting by the created date.


hspedro (approved):
LGTM

gmurayama (approved):

PR #662, opened 2025-05-15
chore: replace k3s, reduce log level, update deps, multi-arch builds

hspedro:
## Summary

This PR introduces several improvements and updates to Maestro’s local development, CI setup, and core runtime, focusing on modernization, reliability, and developer experience. The main highlights include upgrading core dependencies, refactoring Kubernetes integration to use k3d, enabling multi-architecture image builds, and addressing several runtime and test stability issues.

---

### 1. **Upgrade Core Dependencies**
- **Redis 8 & Postgres 17**: Updated the local development and test environment to use the latest stable versions of Redis (8.0-alpine) and Postgres (17-alpine) in `docker-compose.yml` and related scripts.
- **Go 1.23**: Bumped the Go version to 1.23 across the codebase and CI workflows.

### 2. **Kubernetes Integration Refactor**
- **Switch from k3s to k3d**: Replaced the deprecated k3s-based local Kubernetes cluster with k3d, which is better maintained and compatible with modern CI runners.
  - Refactored Makefile targets and documentation to use `k3d` for cluster management.
  - Updated integration test setup and developer instructions accordingly.
  - Adjusted test and local environment scripts to ensure seamless cluster startup and teardown.

### 3. **Multi-Architecture Image Builds**
- **Enable Multi-Arch Support**: Docker image builds now support both `linux/amd64` and `linux/arm64` platforms by default, with local builds using the host architecture.
- **CI/CD**: Updated build and publish steps to ensure images are available for both major architectures.

### 4. **Forwarder Improvements**
- **gRPC Dial Options**: Added support for passing custom gRPC dial options (e.g., `WithBlock()`) to the forwarder client. This is particularly useful for tests that require deterministic connection establishment.
- **Test Stability**: Integration tests now use blocking gRPC dials to avoid race conditions and flaky test failures.

### 5. **Runtime and Metrics Logging**
- **Reduce Metrics Log Level**: Lowered the log level for metrics reporting to avoid log spam, especially in production environments.
- **Race Condition Fixes**: Addressed a race condition in the game room watcher logic, improving runtime stability.
- **PDB Update Retry**: Implemented retry-on-conflict logic for PodDisruptionBudget (PDB) updates, following Kubernetes best practices.

### 6. **CI/CD and Documentation**
- **CI Image Update**: Updated GitHub Actions workflows to use a non-deprecated Ubuntu image.
- **Documentation**: Revised developer and CI documentation to reflect the new k3d setup, dependency versions, and updated Makefile targets.

---

## How it was tested

- **Local Testing**:
  - Verified local development environment with the new Redis, Postgres, and k3d setup.
  - Ran all integration and unit tests locally on both amd64 and arm64 hosts.
  - Manually validated Docker image builds for both architectures.

- **Continuous Integration**:
  - All changes were validated in CI pipelines, including:
    - Cluster bring-up/tear-down with k3d.
    - Full test suite execution (unit, integration, and e2e).
    - Multi-arch Docker image build and push.
    - Linting and static analysis.

- **Test Improvements**:
  - Integration tests for the forwarder and runtime components were stabilized by ensuring gRPC connections are established before test execution.
  - Addressed previously flaky tests related to cluster readiness and resource conflicts.

---

## Documentation Updates

- Updated all relevant documentation and how-to guides to:
  - Reference k3d instead of k3s.
  - Reflect new dependency versions (Go, Redis, Postgres).
  - Document new Makefile targets and local setup instructions.
  - Note multi-arch image support and how to use it.

---

## Additional Notes

- These changes are backward compatible for most developer workflows, but **all contributors should update their local environments** to use k3d and the new dependency versions.
- Please refer to the updated documentation for migration steps and troubleshooting tips.


joaobologna (approved):
LGTM

.gitignore
```diff
@@ -52,6 +52,7 @@ env
 # K3S files
 kubeconfig.yaml
 e2e/framework/maestro/kubeconfig.yaml
```
joaobologna, line 54: is this skill required?

Dockerfile
```diff
@@ -30,13 +34,12 @@ COPY . /build
...
 
-RUN apk add --update make
-
-RUN cd /build && \
-    make build-linux-x86_64 && \
-    mv ./bin/maestro-linux-x86_64 /app/maestro && \
-    mv internal/service/migrations /app/migrations && \
-    mv ./config /app/config
+# Instead of 'make build', compile directly for the target platform
+# Output the binary directly to /app/maestro
```
joaobologna, line 38: nice

Makefile
```diff
@@ -48,35 +48,65 @@ run/unit-tests: ## Execute unit tests.
...
-	@go test -tags=integration -count=1 -timeout 20m -coverprofile=coverage.out -covermode=atomic $(OPTS) ./...
+run/integration-tests: k3d/up k3d/kubeconfig ## Execute integration tests.
+	@KUBECONFIG_PATH="$(shell pwd)/e2e/framework/maestro/.k3d-kubeconfig.yaml"; \
+	echo "INFO: Checking for Kubeconfig at $$KUBECONFIG_PATH for integration tests..."; \
+	if [ -f "$$KUBECONFIG_PATH" ] && [ -s "$$KUBECONFIG_PATH" ]; then \
+		echo "INFO: Using kubeconfig from $$KUBECONFIG_PATH for integration tests..."; \
+		KUBECONFIG="$$KUBECONFIG_PATH" go test -tags=integration -count=1 -timeout 20m -coverprofile=coverage.out -covermode=atomic $(OPTS) ./... ; \
+	else \
+		echo "WARNING: Kubeconfig $$KUBECONFIG_PATH not found or is empty. Please ensure 'make k3d/up' and 'make k3d/kubeconfig' completed successfully."; \
+		echo "INFO: Attempting to run integration tests with default kubeconfig (this might fail if not configured for a K8s cluster)..."; \
```
joaobologna, line 59: i would fail and not run the test because it might be hard for those with this failure to check the initial lines that contain this INFO

or echo after running the command
wdyt?

Makefile
```diff
@@ -48,35 +48,65 @@ run/unit-tests: ## Execute unit tests.
...
+run/runtime-integration-tests: k3d/up k3d/kubeconfig ## Execute runtime integration tests.
+	@KUBECONFIG_PATH="$(shell pwd)/e2e/framework/maestro/.k3d-kubeconfig.yaml"; \
+	echo "INFO: Checking for Kubeconfig at $$KUBECONFIG_PATH for runtime integration tests..."; \
+	if [ -f "$$KUBECONFIG_PATH" ] && [ -s "$$KUBECONFIG_PATH" ]; then \
+		echo "INFO: Using kubeconfig from $$KUBECONFIG_PATH for runtime integration tests..."; \
+		KUBECONFIG="$$KUBECONFIG_PATH" go test -tags=integration -count=1 -timeout 20m ./internal/adapters/runtime/kubernetes/... ; \
+	else \
+		echo "WARNING: Kubeconfig $$KUBECONFIG_PATH not found or is empty. Please ensure 'make k3d/up' and 'make k3d/kubeconfig' completed successfully."; \
+		echo "INFO: Attempting to run runtime integration tests with default kubeconfig (this might fail if not configured for a K8s cluster)..."; \
+		go test -tags=integration -count=1 -timeout 20m ./internal/adapters/runtime/kubernetes/... ; \
```
joaobologna, line 73: ditto

Makefile
```diff
@@ -150,15 +180,76 @@ deps/down: ## Delete containers dependencies.
...
+
+.PHONY: k3d/up
+k3d/up: ## Create/Start the k3d cluster via docker-compose.
+	@echo "INFO: Ensuring k3d service is up and cluster '$(K3D_CLUSTER_NAME)' is created via docker-compose..."
+	@docker-compose up -d k3d
+	@echo "INFO: Waiting for k3d cluster and kubeconfig file at ./e2e/framework/maestro/.k3d-kubeconfig.yaml..."
+	@timeout=120; \
+	interval=5; \
+	elapsed=0; \
+	until [ -f ./e2e/framework/maestro/.k3d-kubeconfig.yaml ] && [ -s ./e2e/framework/maestro/.k3d-kubeconfig.yaml ]; do \
```
joaobologna, line 211: wait, dont you need to run `k3d/kubeconfig` for this file to exist? from what I see tests will do k3d/up and then k3d/kubeconfig

so how will this file exist? I'm 100% missing something here haha

docs/tutorials/Development.md
```diff
@@ -84,7 +84,7 @@ Use the command below to create a new scheduler:
 > Be aware to change the {{ROOMS_API_ADDRESS}} for the one found above.
 ```shell
 curl --request POST \
-  --url http://localhost:8080/schedulers \
+  --url http://localhost:38080/schedulers \
```
joaobologna, line 87: is this expected?

e2e/framework/maestro/docker-compose.yml
```diff
@@ -87,55 +83,16 @@ services:
...
-        hard: 65535
-    privileged: true
-    restart: always
-    environment:
-    - K3S_URL=https://k3s_server:6443
-    - K3S_TOKEN=maestro
-
   postgres:
-    image: postgres:9.6.9
+    image: postgres:17-alpine
```
joaobologna, line 95: 🎉

e2e/framework/maestro/docker-compose.yml
```diff
@@ -149,7 +106,7 @@ services:
       retries: 5
 
   redis:
-    image: redis:6.2.0-alpine
+    image: redis:8.0-alpine
```
joaobologna, line 109: 🎉 🎉

e2e/suites/storagecleanup_operation_test.go
```diff
@@ -59,7 +59,7 @@ func TestStorageCleanUpOperation(t *testing.T) {
 					"--data-raw '{\"status\": \"ready\",\"timestamp\": \"12312312313\"}' && sleep 1; done"})
 			require.NoError(t, err)
 
-			operations, _, err := operationStorage.ListSchedulerFinishedOperations(context.Background(), scheduler.Name, 0, -1)
+			operations, _, err := operationStorage.ListSchedulerFinishedOperations(context.Background(), scheduler.Name, 0, -1, "desc")
```
joaobologna, line 62: flaky test?

internal/core/services/rooms/room_manager.go
```diff
@@ -185,12 +185,12 @@ func (m *RoomManager) UpdateRoomInstance(ctx context.Context, gameRoomInstance *
...
 	m.Logger.Sugar().Infof("Cleaning room \"%v\", scheduler \"%v\"", roomId, schedulerName)
 	err := m.RoomStorage.DeleteRoom(ctx, schedulerName, roomId)
-	if err != nil && !errors.Is(porterrors.ErrNotFound, err) {
+	if err != nil && !errors.Is(err, porterrors.ErrNotFound) {
 		return fmt.Errorf("failed to delete room state: %w", err)
 	}
 
 	err = m.InstanceStorage.DeleteInstance(ctx, schedulerName, roomId)
-	if err != nil && !errors.Is(porterrors.ErrNotFound, err) {
+	if err != nil && !errors.Is(err, porterrors.ErrNotFound) {
```
joaobologna, line 193: why switch?

docs/tutorials/Development.md
```diff
@@ -138,3 +138,88 @@ Feel free to explore the available endpoints in the [API](../reference/OpenAPI.m
...
+    To start these dependencies, including the k3d cluster:
+    ```bash
+    make deps/up
+    ```
+    Alternatively, to only start/ensure the k3d cluster is up:
+    ```bash
+    make k3d/up
+    ```
+    This command will also wait until the kubeconfig file is ready.
+    The `maestro-dev` cluster will have port `38080` on your host mapped to the cluster's ingress (port 80 on the load balancer).
```
joaobologna, line 167: what is this port used for? to issue requests to k8s? for example, add pods, etc.

what does it mean "on your host mapped to the cluster's ingress (port 80 on the load balancer)"?

internal/adapters/events/forwarder_client.go
```diff
@@ -164,43 +170,61 @@ func (f *ForwarderClient) CacheDelete(forwarderAddress string) error {
 }
 
 func (f *ForwarderClient) getGrpcClient(address Address) (pb.GRPCForwarderClient, error) {
-	if address == "" {
```
joaobologna, line 167: is this intended?

internal/adapters/events/forwarder_client.go
```diff
@@ -62,34 +65,37 @@ type ForwarderClient struct {
...
+	// Apply our defaults if not provided by the user
+	if keepAliveCfg.Time == 0 {
+		keepAliveCfg.Time = DefaultKeepAliveTime
 	}
-	if keepAliveCfg.Timeout > 0 {
-		config.KeepAlive.Timeout = keepAliveCfg.Timeout
+	if keepAliveCfg.Timeout == 0 {
+		keepAliveCfg.Timeout = DefaultKeepAliveTimeout
 	}
+	// Note: PermitWithoutStream defaults to false in grpc/keepalive.ClientParameters.
```
joaobologna, line 85: we dont support this anymore? i dont get it

internal/adapters/events/forwarder_client_test.go
```diff
@@ -225,20 +223,18 @@ func TestSendPlayerEvent(t *testing.T) {
 			"../../../test/data/events_mock/events-forwarder-grpc-send-player-event-failure.json",
 		)
 		require.NoError(t, err)
-
 		basicArrangeForwarderClient(t)
 		event := newPlayerEvent("446bb3d0-0334-4468-a4e7-8068a97caa53")
-
-		response, err := forwarderClientAdapter.SendPlayerEvent(context.Background(), newForwarder(grpcMockAddress), &event)
+		response, err := forwarderClientAdapter.SendPlayerEvent(context.Background(), newForwarder(grpcMockAddress), &event, grpc.WaitForReady(true))
```
joaobologna, line 228: these `grpc.WaitForReady(true)` serve a new purpose on the tests?

internal/adapters/runtime/kubernetes/scheduler_test.go
```diff
@@ -27,37 +27,74 @@ package kubernetes
...
+	"math/rand"
 	"testing"
 	"time"
 
 	"github.com/stretchr/testify/require"
 	"github.com/topfreegames/maestro/internal/core/entities"
 	"github.com/topfreegames/maestro/internal/core/entities/autoscaling"
-	"github.com/topfreegames/maestro/internal/core/ports/errors"
-	"github.com/topfreegames/maestro/test"
+	"github.com/topfreegames/maestro/internal/core/ports/errors" // Import for random string
```
joaobologna, line 38: comment should be in math/rand?

internal/adapters/runtime/kubernetes/scheduler_test.go
```diff
@@ -128,16 +188,25 @@ func TestPDBCreationAndDeletion(t *testing.T) {
 		require.Contains(t, pdb.Spec.Selector.MatchLabels["maestro-scheduler"], scheduler.Name)
 		require.Contains(t, pdb.Labels, "app.kubernetes.io/managed-by")
 		require.Contains(t, pdb.Labels["app.kubernetes.io/managed-by"], "maestro")
+
+		// We are testing PDB creation with scheduler. Deletion of PDB is tied to scheduler deletion.
+		// The t.Cleanup will handle deleting the scheduler, which should delete the PDB.
+		// We then check if the PDB is gone.
+		kubernetesRuntime.DeleteScheduler(ctx, scheduler) //nolint:errcheck
```
joaobologna, line 195: here, in particular, i think it's worth to add `require.NoError` so we don't omit a potential failure cases that would otherwise be hard to check via test output

internal/adapters/runtime/kubernetes/scheduler_test.go
```diff
@@ -152,138 +221,167 @@ func TestPDBCreationAndDeletion(t *testing.T) {
...
+		require.NoError(t, err, "failed to create scheduler for PDB test (with autoscaling)")
 
-		pdb, err := client.PolicyV1().PodDisruptionBudgets(scheduler.Name).Get(ctx, scheduler.Name, metav1.GetOptions{})
+		pdb, err := clientset.PolicyV1().PodDisruptionBudgets(scheduler.Name).Get(ctx, scheduler.Name, metav1.GetOptions{})
 		require.NoError(t, err)
 		require.NotNil(t, pdb)
 		require.Equal(t, pdb.Name, scheduler.Name)
 		require.Equal(t, pdb.Spec.MinAvailable.IntVal, int32(0))
+
+		kubernetesRuntime.DeleteScheduler(ctx, scheduler) //nolint:errcheck
```
joaobologna, line 241: ditto

internal/adapters/runtime/kubernetes/scheduler_test.go
```diff
@@ -152,138 +221,167 @@ func TestPDBCreationAndDeletion(t *testing.T) {
...
-		pdb, err := client.PolicyV1().PodDisruptionBudgets(scheduler.Name).Get(ctx, scheduler.Name, metav1.GetOptions{})
+		pdb, err := clientset.PolicyV1().PodDisruptionBudgets(scheduler.Name).Get(ctx, scheduler.Name, metav1.GetOptions{})
 		require.NoError(t, err)
 		require.NotNil(t, pdb)
 		require.Equal(t, pdb.Name, scheduler.Name)
 		require.Equal(t, pdb.Spec.MinAvailable.IntVal, int32(0))
 	})
 
 	t.Run("should update PDB on mitigation if not equal to current value", func(t *testing.T) {
+		t.Parallel() // Ensure parallel execution
```
joaobologna, line 333: were you able to get consistent green test executions? i fear, even by adding distinct scheduler names, we'll introduce flaky tests because of parallel execution

test/grpc_mock.go
joaobologna, line 1: did you made this setup or its insiped by gripmock doc?

Makefile
hspedro: I agree better to fail open

Makefile
hspedro: previously, file was created/exported in `k3d/kubeconfig`, however, I've changed so it exports in the docker-compose command directly, so `k3d/kubeconfig` is not necessary anymore. `k3d/up` is enough

internal/adapters/events/forwarder_client.go
hspedro: Yes. I've moved validation of grpc connection settings to when creating the connection: 

```go
func (f *ForwarderClient) createGRPCConnection(address string, opts ...grpc.DialOption) (*grpc.ClientConn, error) {
	if address == "" {
		return nil, errors.NewErrInvalidArgument("no rpc server address informed")
	}
```

Thus, this function should only get from cache or create a new one. The error in `createGRPCConnection` will be propagated to this fn either way

internal/adapters/events/forwarder_client.go
hspedro: Yep. Forgot to remove

internal/adapters/events/forwarder_client_test.go
hspedro: Yes. So, on tests we are spawning a container called `grpcmock` implementing the protos simulating the receiving end of matchmaker. However, this container does not have reliable readiness and tests were becoming flaky.

One way to solve this is to pass `WaitForReady(true)` when you are doing the RPC call, which causes the dial/call to be blocking on waiting for a connection from the pool which passes keepalive checks **before** actually performing the call. It's safer with the drawback of being slower. For tests I just want to guarantee they are not flaky, hence why I'm passing

internal/adapters/runtime/kubernetes/scheduler_test.go
hspedro: True 🤣

internal/adapters/runtime/kubernetes/scheduler_test.go
```diff
@@ -68,56 +105,79 @@ func TestSchedulerCreation(t *testing.T) {
...
 
-		ns, err := client.CoreV1().Namespaces().Get(ctx, scheduler.Name, metav1.GetOptions{})
-		require.NoError(t, err)
-		require.Equal(t, v1.NamespaceTerminating, ns.Status.Phase)
+		// Check it's terminating
+		ns, err := clientset.CoreV1().Namespaces().Get(ctx, scheduler.Name, metav1.GetOptions{})
+		if err == nil { // It might be already gone, which is fine
+			require.Equal(t, v1.NamespaceTerminating, ns.Status.Phase)
+		} else {
+			require.True(t, kerrors.IsNotFound(err), "expected namespace to be terminating or not found, but got: %v", err)
```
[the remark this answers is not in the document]
hspedro: That's what I'm doing in `require.True`, if err is not nil, then it must be `kerrors.IsNotFound`, otherwise we'll fail with the message in the third argument ""expected namespace to be terminating or not found, but got: %v""

internal/adapters/runtime/kubernetes/scheduler_test.go
hspedro: Hard to say, I've tested 10 times in a row and they all passed. On CI as well, but we need to monitor for sure. I'd like to maintain in parallel because we can have multiple runtime watches running, and also at some point we had a tool from CPE updating the scheduler PDB, so this conflicts might happen and we should be resilient to that.

Adding the `t.Parallel()` made me found an issue in the update PDB impl, hence why I've added the `RetryOnConflict` from `k8s` API

test/grpc_mock.go
hspedro: I started with the docs, but tbh had a terrible experience with flaky tests and this seems an overkill (in-general) to bring up a container to mock a GRPC API/proto... The comments and setup became more complicated to solve that and it was done via AI 🫠

.gitignore
hspedro: No, removing it

Makefile
joaobologna: should we remove k3d/kubeconfig then? and update docs

docs/tutorials/Development.md
joaobologna: ok, feel free to close the thread once you revert this change

internal/adapters/runtime/kubernetes/scheduler_test.go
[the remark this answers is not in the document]
joaobologna: oh, my bad, true

internal/adapters/runtime/kubernetes/scheduler_test.go
joaobologna: ok, sold

Makefile
hspedro: Yep. I removed k3d/kubeconfig command and simplified the docs!

PR #668, opened 2025-07-07
Added maestro room resilience

ffelipelimao:
## Description

- Changed to goroutine not affected another goroutine to create room, only when all rooms failed the operation will sabe with error
- Remove redis retry opts at create room to avoid retry the entire pipeline when one operation fails

## Evidences

- I made a port foward to this pod running maestro-next at kube-stag cluster

Captura de Tela 2025-07-08 às 17 18 53

- Create a scheduler `"schedulerroom12"` with one `while true` image
Captura de Tela 2025-07-08 às 17 19 05

- Create k8s namespace 
Captura de Tela 2025-07-08 às 17 20 39

- Game room showing at Datadog
Captura de Tela 2025-07-08 às 17 21 05

### DIFF
--- a/internal/core/operations/rooms/add/executor.go
+++ b/internal/core/operations/rooms/add/executor.go
@@ -101,9 +101,6 @@ func (ex *Executor) Execute(ctx context.Context, op *operation.Operation, defini
 			defer cancel()
 
 			errsCh <- ex.createRoom(ctx, scheduler, executionLogger)
-			if err != nil {
-				errsCh <- err
-			}
 		}()
 	}
 


internal/core/operations/rooms/add/executor.go
```diff
@@ -88,24 +88,61 @@ func (ex *Executor) Execute(ctx context.Context, op *operation.Operation, defini
...
+
+		executionLogger.Error(ErrAllRooms.Error(),
+			zap.Error(collectedErrors[0]),
+			zap.Int32("failedRoomsCount", errCount),
+			zap.Any("allErrors", collectedErrors))
+
+		ex.operationManager.AppendOperationEventToExecutionHistory(ctx, op, ErrAllRooms.Error())
+		return ErrAllRooms
+	} else {
+		logMessage := fmt.Sprintf("some rooms failed to be created, errors: %d - amount: %d ", errCount, amount)
```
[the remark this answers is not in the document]
ffelipelimao: agree! done!

internal/core/operations/rooms/add/executor.go
```diff
@@ -88,24 +88,61 @@ func (ex *Executor) Execute(ctx context.Context, op *operation.Operation, defini
...
 	for i := int32(1); i <= amount; i++ {
-		errGroup.Go(func() error {
-			err := ex.createRoom(errContext, scheduler, executionLogger)
-			return err
-		})
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+
+			err := ex.createRoom(ctx, scheduler, executionLogger)
```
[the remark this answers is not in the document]
ffelipelimao: If the error is in the Redis pipeline, it will return an error and the health controller will try to create the room again.

If the error is in the runtime, it already has a routine to delete it

internal/core/services/rooms/room_manager.go#L44

`instance, err := m.Runtime.CreateGameRoomInstance(ctx, scheduler, roomName, *spec)
	if err != nil {
		deleteRoomErr := m.RoomStorage.DeleteRoom(ctx, scheduler.Name, room.ID)
		if deleteRoomErr != nil {
			return nil, nil, fmt.Errorf("error deleting room during create game room instance error: %w", deleteRoomErr)
		}
		return nil, nil, err
	}`

internal/core/operations/rooms/add/executor.go
```diff
@@ -88,24 +88,61 @@ func (ex *Executor) Execute(ctx context.Context, op *operation.Operation, defini
...
 	}
 
-	if executionErr := errGroup.Wait(); executionErr != nil {
-		executionLogger.Error("Error creating rooms", zap.Error(executionErr))
-		ex.operationManager.AppendOperationEventToExecutionHistory(ctx, op, executionErr.Error())
-		return executionErr
+	go func() {
+		wg.Wait()
+		close(errsCh)
+	}()
```
hspedro, line 109: what's the diff between putting this out of a goroutine and block on the `wg.Wait()` rather than blocking in the `range errsCh` below?

internal/core/operations/rooms/add/executor.go
hspedro, line 99: does this have timeout? my rationale is to have an outside timeout (like context cancellation) so we don't block waiting for N goroutines. If each goroutine has a timeout to the operations with Redis maybe that's fine since it's the way it's currently set up

ffelipelimao: done!

internal/core/operations/rooms/add/executor.go
```diff
@@ -88,24 +89,54 @@ func (ex *Executor) Execute(ctx context.Context, op *operation.Operation, defini
...
-			return err
-		})
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+
+			ctx, cancel := context.WithTimeout(ctx, 5*time.Second)
+			defer cancel()
+
+			err := ex.createRoom(ctx, scheduler, executionLogger)
```
hspedro, line 103: nit: haven't you changed to the below?

```suggestion
			errsCh <- ex.createRoom(ctx, scheduler, executionLogger)
```

hspedro (approved):
LGTM

PR #673, opened 2025-07-22
feat(add): batch room creation

hspedro:
Previously, add room operation created one goroutine to each room to be created. Depending on the amount we get rate limited by k8s or context cancellation from Redis. To mitigate that we operate the creation in batches so we can space out this process and avoid errors from exetrnal providers.

Main Changes:
- Implemented default batch size handling in the executor configuration.
- Added tests to verify batch processing behavior, including scenarios for default batch size, error handling, and varying amounts of rooms.
- Ensured that the executor correctly processes rooms in batches, handling both success and failure cases appropriately.

### DIFF
--- a/internal/core/operations/rooms/add/executor.go
+++ b/internal/core/operations/rooms/add/executor.go
@@ -40,10 +40,14 @@ import (
 	"github.com/topfreegames/maestro/internal/core/operations"
 )
 
-const DefaultAmountLimit = 1000
+const (
+	DefaultAmountLimit = 1000
+	DefaultBatchSize   = 40
+)
 
 type Config struct {
 	AmountLimit int32
+	BatchSize   int32
 }
 
 type Executor struct {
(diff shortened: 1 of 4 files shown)


ffelipelimao (approved):

internal/core/operations/rooms/add/executor.go
```diff
@@ -89,29 +97,54 @@ func (ex *Executor) Execute(ctx context.Context, op *operation.Operation, defini
 		amount = ex.config.AmountLimit
 	}
 
-	var wg sync.WaitGroup
-	errsCh := make(chan error, amount)
+	executionLogger.Info("starting batched room creation",
```
joaobologna, line 100: cant this become noisy? maybe change to debug?

internal/core/operations/rooms/add/executor_test.go
```diff
@@ -185,6 +185,148 @@ func TestExecutor_Execute(t *testing.T) {
...
+		operationsManager.EXPECT().AppendOperationEventToExecutionHistory(gomock.Any(), &op, "more rooms failed than succeeded, errors: 5 and successes: 1 of amount: 6")
+
+		executor := NewExecutor(roomsManager, schedulerStorage, operationsManager, batchConfig)
+		err := executor.Execute(context.Background(), &op, &definition)
+
+		require.NotNil(t, err)
+		require.ErrorContains(t, err, "more rooms failed than succeeded, errors: 5 and successes: 1 of amount: 6")
+	})
+
+	t.Run("should succeed - amount smaller than batch size", func(t *testing.T) {
```
joaobologna, line 274: if the default is 30, the other tests already cover this to some extent

you could remove it, but it’s not a big deal either way

joaobologna (approved):
LGTM

PR #679, opened 2025-08-14
feat(k8s): restart policy handling in game room spec

hspedro:
The RestartPolicy of pods is now by default `Never` instead of `Always`, which reduces weird scenarios where the game room container exited and remains in Completed, ContainerStatusUnknown, or Failed. If the container stops, the pod dies, and Maestro spawns another one.

- Added support for specifying a restart policy in the game room specification.
- Updated conversion functions to handle the new restart policy field.
- Modified relevant tests to include scenarios for the restart policy.


internal/adapters/runtime/kubernetes/game_room_convert.go
```diff
@@ -293,18 +296,71 @@ func convertPodStatus(pod *v1.Pod) game_room.InstanceStatus {
 		return game_room.InstanceStatus{Type: game_room.InstanceTerminating}
 	}
 
+	// Handle terminal pod phases early so we don't incorrectly mark them as pending/unknown
+	switch pod.Status.Phase {
```
ffelipelimao, line 300: make sense to put a defult at this switch case to avoid edge case?

ffelipelimao (approved):

PR #683, opened 2025-09-12
refactor(pdb): updated calculation of desired minimum available rooms

fbgava-wildlife:


hspedro (approved):
LGTM

internal/adapters/runtime/kubernetes/scheduler.go
```diff
@@ -233,7 +231,7 @@ func (k *kubernetes) MitigateDisruption(
 			currentPdbMinAvailable = pdb.Spec.MinAvailable.IntVal
 		}
 
-		desiredMinAvailable := int32(float64(roomAmount) * incSafetyPercentage)
+		desiredMinAvailable := int32(float64(roomAmount) * (1.0 - safetyPercentage))
```
hspedro, line 234: Does int32 rounds to floor always? If not we should do math.Floor

joaobologna (approved):
LGTM

PR #720, opened 2026-05-27
feat(redis): add configurable MaxConnAge to bound failover error window

lucasmonje:
## What does this implement/fix? Explain your changes.

Adds a configurable `MaxConnAge` to the go-redis client used by every maestro Redis adapter, to bound the error window during an ElastiCache failover/switchover.

**Why.** During a switchover (failover, Redis→Valkey engine upgrade, scaling), AWS swaps the primary/replica role of the cluster's nodes. Pooled TCP conns stay attached to the node they originally dialed, which is now a replica, and any subsequent write returns:

```
READONLY You can't write against a read only replica.
```

go-redis does not treat `READONLY` as a bad-conn signal, so the offending conn is returned to the pool and reused, perpetuating the failure until it's recycled by some other means. `IdleTimeout` doesn't help on a hot pool (conns never sit idle long enough for the reaper). `MaxConnAge` is checked at every conn checkout, so it guarantees turnover regardless of traffic; after the bound elapses the next dial re-resolves DNS and lands on the new primary.

**What.**
- `createRedisClient` now reads `adapters.redis.maxConnAge` and sets `opt.MaxConnAge` when > 0. Unset/0 keeps the go-redis default (no limit), so behavior is unchanged unless configured.
- Shared knob across all six Redis adapters, mirroring the existing `adapters.redis.poolSize`. Env var: `MAESTRO_ADAPTERS_REDIS_MAXCONNAGE` (Go duration, e.g. `5m`).

**Operational note.** Set `MAESTRO_ADAPTERS_REDIS_MAXCONNAGE=5m` (or shorter) on the deployment before the planned Redis→Valkey maintenance window. Safe to leave on permanently; overhead is negligible on production-sized pools and it future-proofs against unplanned failovers.

**What this does *not* do.** This bounds the post-failover error window; it does not eliminate it. A request landing on a stale conn during the swap still sees one `READONLY` before that conn is recycled. A retry-on-READONLY wrapper around the room-allocation Lua script would be a complementary follow-up, out of scope here.

This mirrors the matchmaker change in MR !137 (matchmaker is on go-redis v6; maestro stays on v8, where the field is still named `MaxConnAge` — it becomes `ConnMaxLifetime` in v9).

## Any other comments?

- New `//go:build unit` test covers both the set case (`5m` reaches `client.Options().MaxConnAge`) and the unset-keeps-default case.
- Existing integration tests updated with the new `GetDuration` expectation.
- `go build`, `go vet -tags=integration`, `goimports`, license check, and the new unit tests all pass locally. Integration suite (k3d) not run locally.

### DIFF
--- a/internal/service/adapters.go
+++ b/internal/service/adapters.go
@@ -93,6 +93,12 @@ const (
 	operationFlowRedisURLPath = "adapters.operationFlow.redis.url"
 	// Redis configs
 	redisPoolSizePath = "adapters.redis.poolSize"
+	// Connection age at which the go-redis client retires (closes) a conn.
+	// Unset/0 keeps the go-redis default (no limit). Set a finite value to
+	// bound the post-failover window in which pooled conns can keep talking
+	// to a node that has been demoted from primary (ElastiCache Redis->Valkey
+	// engine swaps, scaling, unplanned failover).
+	redisMaxConnAgePath = "adapters.redis.maxConnAge"
 	// Random port allocator
 	portAllocatorRandomRangePath = "adapters.portAllocator.random.range"
 	// Postgres scheduler storage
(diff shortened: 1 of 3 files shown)


fbgava-wildlife (approved):
LGTM

PR #721, opened 2026-06-16
fix(redis): occupancy leak — stop GetRoom resurrecting rooms, and make running-matches orphan-immune

lucasmonje:
## What does this implement/fix? Explain your changes.

The `roomOccupancy` autoscaler input `running_matches` is the score-sum of `scheduler:<name>:occupancy`. That set accumulates orphan members that are never removed, so `running_matches` drifts upward and never returns to reality. Once it sits above the downscale threshold, `CanDownscale` is permanently false: the scheduler stops downscaling and rolling updates can't drain old-version rooms. Observed in prod: `:occupancy` had ~2.3k members vs ~5 in `:status`, `running_matches` pinned at 410 while the API reported 0 occupied; killing pods and switching versions didn't move it (the state is in Redis, not the pods).

Two commits, prevention + immunity:

**1. Stop `GetRoom` from resurrecting deleted rooms into `:occupancy`.**
`GetRoom` (a read) did an unconditional `ZAddNX` into `:occupancy` (added in #659 to backfill rooms predating the occupancy feature). But `GetRoom` is also called for rooms being/already deleted (late pings, in-flight operations), so it re-created an occupancy-only entry (score 0) that no path removes — `GetAllRoomIDs` only iterates `:status`, and ping-expiry/pod-gone cleanup only act on status-tracked rooms. Removed the write: a read must be side-effect free. `CreateRoom` already establishes the occupancy entry, and `UpdateRoom` uses `ZAddXXCh` precisely to avoid resurrecting deleted rooms; reads now match. After this, the only occupancy writers are `CreateRoom` (adds all three sets together) and `UpdateRoom` (`XXCh`, update-only) — neither can create an occupancy-only orphan.

**2. Make `GetRunningMatchesCount` immune to any orphan that already exists.**
Prevention alone doesn't help schedulers that already leaked, and operators shouldn't have to `ZREM` orphans by hand. `GetRunningMatchesCount` now sums occupancy scores only for rooms still present in `:status` (a running match only lives in a tracked room), via a read-only Lua script. This is self-correcting: stale occupancy members are ignored automatically, so the autoscaler and the `running_matches` metric stay correct even before/without any cleanup.

Net: #1 stops the leak at the source; #2 makes the autoscaler correct regardless of existing orphans — no manual reconciliation required.

Tests: added a regression test that `GetRoom` on a missing room creates no occupancy entry; updated the running-matches storage tests to register rooms in `:status` and added a case asserting an orphan occupancy entry is not counted. All `internal/adapters/storage/redis/room` (integration) and `internal/core/services/autoscaler/...` (unit) tests pass.

## Does this close any currently open issues?

No tracked issue; full root-cause analysis with `file:line` references is in the description and commits.

## Any relevant logs, error output, etc?

While stuck: `scheduler <x> can't downscale, occupation is above the threshold` (emitted every cycle with 0 occupied rooms).

## Any other comments?

The `GetRoom` `ZAddNX` regression was introduced in #659 (commit `7590c7fa`, first released v10.11.4, 2025-02-28); present through current `main`. With #2 in place, the historical orphan entries become harmless on deploy; cleaning them from Redis is optional hygiene, no longer required for correctness.

### DIFF
--- a/internal/adapters/storage/redis/room/redis.go
+++ b/internal/adapters/storage/redis/room/redis.go
@@ -88,13 +88,16 @@ end
 // Ignoring them keeps GetRunningMatchesCount — and therefore the roomOccupancy autoscaler and the
 // running_matches metric — correct even if the occupancy set has stale members. Read-only.
 var runningMatchesScript = redis.NewScript(`
-local occupancy_key = KEYS[1]
-local status_key = KEYS[2]
+local occupancy = redis.call('ZRANGEBYSCORE', KEYS[1], '-inf', '+inf', 'WITHSCORES')
+local status = redis.call('ZRANGE', KEYS[2], 0, -1)
+local in_status = {}
+for i = 1, #status do
+    in_status[status[i]] = true
+end
 local sum = 0
-local entries = redis.call('ZRANGEBYSCORE', occupancy_key, '-inf', '+inf', 'WITHSCORES')
-for i = 1, #entries, 2 do
-    if redis.call('ZSCORE', status_key, entries[i]) then
-        sum = sum + tonumber(entries[i + 1])
+for i = 1, #occupancy, 2 do
+    if in_status[occupancy[i]] then
+        sum = sum + tonumber(occupancy[i + 1])
     end
 end
 return sum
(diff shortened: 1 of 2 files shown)


fbgava-wildlife (approved):
LGTM

